PR #49, opened 2015-02-10
core: added deobfuscation feature (basic functionality)

NeoSpb:
Added deobfuscation feature (basic functionality).

New command line options:

    --deobf                Activate deobfuscating
    --deobf-min            The minimum length of a name. The name with a length less
                           than this value will be subjected deobfuscating (default is 2)
    --deobf-max            The maximum length of a name. The name with a length great
                           than this value will be subjected deobfuscating (default is 40)
    --deobf-rewrite-cfg    Forced rewrite map of deobfuscation

Now deobfuscation applied to name of classes and name of packages. In the future it will also apply to fields and methods.

If not exists it creates a file with map of deobfuscation (this file located in same directory as the APK file and has the extension **jobf**)

Each line represents an alias in the following format

    {type} {full_name}={deobfuscated_short_name}

where {type} is:

    p package
    c class

Example:

    p a.a.a=p002a
    p a.a=p001a
    p a=p000a

package a.a.a will be shown in Java code as **p000a.p001a.p002a**

Note: If APK includes small but correct names, you should also correct it by hand, for example:

    p android.support.v4=v4

------

_PS. Yes, I don't think about tests, this feature just working as is. =)_

### DIFF
--- a/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
+++ b/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
@@ -50,6 +50,18 @@ public final class JadxCLIArgs implements IJadxArgs {
 	@Parameter(names = {"-v", "--verbose"}, description = "verbose output")
 	protected boolean verbose = false;
 
+	@Parameter(names = {"--deobf"}, description = "activate deobfuscation")
+	protected boolean deobfuscationOn = false;
+
+	@Parameter(names = {"--deobf-min"}, description = "min length of name")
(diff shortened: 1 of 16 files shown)


YASME-Tim:
Cool :smile:

skylot:
Great! Thank you.
But I don't like Deobfuscator.instance() singleton, maybe I will refactor it )

NeoSpb:
I used singleton to simplify the code (also changes in future). Yes of course, you can refactor it if necessary. (You're the Boss =))

PR #76, opened 2015-07-22
Enable file drop operation for loading it.

jpstotz:
This patch enabled Jadx-gui to accept files via drag & drop from external applications (e.g. file explorer). The dropped file will be loaded.


skylot:
Ok. Thank you.
Sadly, your solution not work on Linux with Java 6 but work on Java 7 and later,
workaround can be found here: http://stackoverflow.com/questions/811248/how-can-i-use-drag-and-drop-in-swing-to-get-file-path , but I don't like this tricky code, so I will leave jadx without drag and drop support on Linux with Java 6 :)

PR #171, opened 2017-04-30
Deobfuscation fixes

daramos:
Fixes the following issues with deobfuscation:
- Name generation on variables of deobfuscated objects were sometimes based off of the original class name, and not the alias.
- Renamed classes/methods referenced in a different dex file would not be renamed properly.
- Deobfuscator.postProcess may overwrite Deobfuscated method names for overriden methods.

Also fixes one issue not related to the  deobfuscation:
- Inner classes residing in an empty package name were not being linked to their parent class due to an issue with ClassInfo.splitNames

Thanks for this project!


asLody:
Good job!

PR #196, opened 2018-01-03
fix Chinese translation and improve deobfusction function

Qrilee:


jadx-core/src/main/java/jadx/core/utils/StringUtils.java
```diff
@@ -193,4 +196,18 @@ private static void commonEscapeAndAppend(StringBuilder sb, char c) {
...
 		}
 	}
+
+	private static final Pattern P0 = Pattern.compile("^[0o]+$");
+	private static final Pattern P1 = Pattern.compile("^[iI]+$");
+	private static final Pattern P2 = Pattern.compile("^[uv]+$");
+	private static final Pattern P3 = Pattern.compile("[0-9]");
+
+	public static boolean dirtyName(String name) {
+		return name != null && !"".equals(name) && (P0.matcher(name.toLowerCase()).matches() || P1.matcher(name).matches() || P2.matcher(name.toLowerCase()).matches());
```
liach: `!"".equals(name)` -> `!name.isEmpty()`

jadx-gui/src/main/java/jadx/gui/utils/NLS.java
```diff
@@ -6,9 +6,14 @@
 public class NLS {
 
 	private static ResourceBundle messages;
+	private static final boolean IS_CHINESE = Locale.SIMPLIFIED_CHINESE==Locale.getDefault();
 
 	static {
-		load(new Locale("en", "US"));
+        if(IS_CHINESE){
+            load(Locale.SIMPLIFIED_CHINESE);
```
liach: Use a better check here

jadx-core/src/main/java/jadx/core/utils/NameFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+    // Implementations for NameFactory.
+    private static int index = 0;
+
+    public static void reset() {
+        index = 0;
+    }
+
+
+    public static String nextName() {
```
liach: no! this literally broke all naming behaviros.

skylot: I agree, previously jadx make classes names something like 'C0010' but this commit changed it to 'ac'.
For me numbers is easier to compare than letters and also it is still looks like obfuscated.
So, I prefer not to change deobfuscation naming.

jadx-core/src/main/java/jadx/core/utils/StringUtils.java
```diff
@@ -193,4 +196,18 @@ private static void commonEscapeAndAppend(StringBuilder sb, char c) {
 			sb.append(c);
 		}
 	}
+
+	private static final Pattern P0 = Pattern.compile("^[0o]+$");
+	private static final Pattern P1 = Pattern.compile("^[iI]+$");
+	private static final Pattern P2 = Pattern.compile("^[uv]+$");
+	private static final Pattern P3 = Pattern.compile("[0-9]");
```
skylot: I like the idea of implementing deobfuscation based on patterns.
But it is better to make it more general: store pattern and some options in list and process matching in loop

jadx-cli/build.gradle
```diff
@@ -5,8 +5,8 @@ applicationName = 'jadx'
 
 dependencies {
     compile(project(':jadx-core'))
-    compile 'com.beust:jcommander:1.47'
-    compile 'ch.qos.logback:logback-classic:1.1.2'
+    compile 'com.beust:jcommander:1.72'
```
skylot: Looks like updating jcommander broke tests in jadx-cli.

jadx-gui/src/main/java/jadx/gui/utils/NLS.java
skylot: @liach can you be more specific how to check locale, where are a lot of ways to do it and I don't know which is better

liach: I think we can just load the default locale with a fallback

PR #206, opened 2018-01-16
Decode attributes which may contain either enums or values, e.g. layo…

mtdcr:
…ut_width

android:layout_width="UNKNOWN_DATA_0x6401" becomes android:layout_width="100dp".

This works, because the upper layer already handles 'null' return values.


skylot:
Thank you for contribution!

PR #228, opened 2018-02-12
Search dialog improvements

jpstotz:
This pull request contains two improvements for the search dialog:

1. Option added for case insensitive searching
2. Pagination added in case more than 100 results are found


skylot:
This is good improvement! Thank you!

PR #251, opened 2018-04-07
Update JadxArgsValidator.java

khalil2535:
char data type instead of String object


sergey-wowwow:
@khalil2535 
Why? :)

sergey-wowwow:
It’s okay :) There is a hundred of opened issues, you may fix anything you want :D

PR #288, opened 2018-05-30
Xml deobf

sergey-wowwow:
1. Before
```xml
<activity android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|screenSize|smallestScreenSize" android:label="@string/app_name" android:launchMode="singleTask" android:name="org.jitsi.meet.MainActivity" android:resizeableActivity="true" android:supportsPictureInPicture="true" android:windowSoftInputMode="adjustResize">
```
After
```xml
<activity android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|screenSize|smallestScreenSize" android:label="@string/app_name" android:launchMode="singleTask" android:name="p028org.jitsi.meet.MainActivity" android:resizeableActivity="true" android:supportsPictureInPicture="true" android:windowSoftInputMode="adjustResize">
```

2. Before 
```xml
<android.support.v7.view.menu.ActionMenuItemView xmlns:android="http://schemas.android.com/apk/res/android" android:focusable="true" android:gravity="center" android:layout_gravity="center" android:layout_height="wrap_content" android:layout_width="wrap_content" android:paddingBottom="4dp" android:paddingLeft="8dp" android:paddingRight="8dp" android:paddingTop="4dp" android:textAppearance="?attr/actionMenuTextAppearance" android:textColor="?attr/actionMenuTextColor" style="?attr/actionButtonStyle"/>
```
After
```xml
<android.support.p007v7.view.menu.ActionMenuItemView xmlns:android="http://schemas.android.com/apk/res/android" android:focusable="true" android:gravity="center" android:layout_gravity="center" android:layout_height="wrap_content" android:layout_width="wrap_content" android:paddingBottom="4dp" android:paddingLeft="8dp" android:paddingRight="8dp" android:paddingTop="4dp" android:textAppearance="?attr/actionMenuTextAppearance" android:textColor="?attr/actionMenuTextColor" style="?attr/actionButtonStyle"/>
```

PR #290, opened 2018-06-01
Xml deobf 2.0

sergey-wowwow:
https://github.com/skylot/jadx/pull/288


skylot:
That's better. Thanks!

PR #299, opened 2018-06-26
Generates and saves public.xml in apktool style

sergey-wowwow:


jadx-core/src/main/java/jadx/core/utils/files/ZipSecurity.java
```diff
@@ -14,14 +15,24 @@
...
 		}
-		if (file.equals(base)) {
+		if (canonFile.equals(baseDir)) {
 			return true;
 		}
-		return isInSubDirectory(base, file.getParentFile());
+		return isInSubDirectoryInternal(baseDir, canonFile.getParentFile());
+	}
+	
+	public static boolean isInSubDirectory(File baseDir, File file) {
```
skylot: Please convert baseDir to absolute path too. Otherwise, jadx skips all resources if run on file without full path like: `jadx some.apk`.

sergey-wowwow: I changed method ``` isInSubDirectory ```, but didn't change ``` isInSubDirectoryInternal ```, because it's used only for canonical paths in the both calls

PR #305, opened 2018-07-04
show app certificate

asviridenko:


skylot:
Thanks @asviridenko! This is great addition to jadx :+1:

PR #315, opened 2018-07-16
show app certificate

asviridenko:
show the certificate if the certificate file name is not standard
eg https://play.google.com/store/apps/details?id=com.kms.free


skylot:
@asviridenko is there a change what in one apk will be several certificate files?
Because in that case, you will show only one of them, is that correct?

mingyuan-xia:
One apk can have more than one certificates, confirmed.

asviridenko:
@mingyuan-xia Please send me a link to app with several certificates, I will make fix

jpstotz:
May be it would be a good idea to use the official "apksig" library for reading the signatures.
https://mvnrepository.com/artifact/com.android.tools.build/apksig/3.1.3
```
<groupId>com.android.tools.build</groupId>
    <artifactId>apksig</artifactId>
    <version>3.1.3</version>
</dependency>

```
It supports multi-signatures as well as APK signature v1, v2 and may be even the upcoming v3 standard (not sure if the mentioned version of the lib already includes the v3 support).
Furthermore it also provides the verification result.

The base class is [`com.android.apksig.ApkVerifier`](https://android.googlesource.com/platform/tools/apksig/+/refs/heads/master/src/main/java/com/android/apksig/ApkVerifier.java).

skylot:
@jpstotz thanks! apksig is a good library.
As I can see it uses several methods for searching certificates, so I agree, it is better to integrate it. 
There is TODO list for certificate functionality:
- use apksig lib for search certificates in apk
- allow several entries to be shown under 'Certificate' root
- maybe add some icon for indicating verification status
- for certificate files under resources root show content in text format (now shows in binary)

@asviridenko  I will merge this pull request because, I think, it is better search certificates by extension and not hardcoded file names. And we will fix other issues later.

PR #351, opened 2018-08-28
add a menu to copy class name.

javaeryang:
as for developing a xposed module need frequently copy the class name, but there is no such menu easily to copy the class name, so i added it.


skylot:
@javaeryang  thank you!

Issue #308, opened 2018-07-06
Activity.findViewById(...) reffers to a missing R class


sergey-wowwow:
Hi, in smali code in ``` com.ripio.android.activities.RipioWeb ``` I see the following instructions:
```smali
    .line 44
    const v0, 0x7f0b0034

    invoke-virtual {p0, v0}, Lcom/ripio/android/activities/RipioWeb;->setContentView(I)V
```

It means that the constant is hardcoded, but it's decompiled to
```java
    setContentView((int) com.ripio.android.R.layout.activity_webview);
```
nevertheless class ``` com.ripio.android.R ``` is missing (in smali and java code).

APK [com.ripio.android-1.apk.zip](https://github.com/skylot/jadx/files/2170991/com.ripio.android-1.apk.zip)

skylot:
@sergey-wowwow jadx trying to find matching R class in app package (read from manifest), but in this apk such class is missing, there are warning in log:
```
11:28:29 INFO  - Found several 'R' class candidates: [com.google.android.gms.R, com.google.firebase.R, com.google.firebase.firebase_core.R, com.google.firebase.iid.R, com.google.firebase.remoteconfig.R]
11:28:29 WARN  - Unknown 'R' class, create references to 'com.ripio.android.R'
```
So, to produce compilable code jadx must generate R class.
I don't make it yet because R class is generated by standard build process from resources xml files (which jadx restore) and in this case R class will be overwritten.

sergey-wowwow:
@skylot,
Wouldn’t it better to output constants when R class is not found?

skylot:
@sergey-wowwow 
No. Show resources names instead of numbers - jadx main feature which other decompilers don't implement )

sergey-wowwow:
@skylot,
Main feature is quality of generated codes :D

sergey-wowwow:
Hi @skylot,
is that a rule to generate R class in the package directory? Moreover, usually R class is renamed when deobfuscation is enabled. I can implement such feature like I did for ``` public.xml ```, but I'm not sure that it's a correct solution, because referenced field should be already generated in other R class when 
```java
LOG.info("Found several 'R' class candidates: {}", candidates);
```
is logged

skylot:
@sergey-wowwow  I think it is easier and correct to generate R class if not found, because:
- R class can be absent at all
- It is hard to search for obfuscated version (unknown name and package)
- Obfuscated R class can contain renamed fields so it is useless to replace constants with such names

[...]

PR #353, opened 2018-09-02
Generates code of missing R class

sergey-wowwow:
Hi, there are a few things:

- generated code isn't getting saved by some reason
- I believe it's not an issue that I don't create inner classes?
- I mentioned arrays in original R classes like ``` public static final int[] ZoomOval = new int[0]; ```, I will find out why they exist, and add it in a new commit in future

This PR is for https://github.com/skylot/jadx/issues/308

### DIFF
--- a/jadx-core/src/main/java/jadx/api/JadxDecompiler.java
+++ b/jadx-core/src/main/java/jadx/api/JadxDecompiler.java
@@ -91,7 +91,6 @@ public final class JadxDecompiler {
 
 		root.initClassPath();
 		root.loadResources(getResources());
-		root.initAppResClass();
 
 		initVisitors();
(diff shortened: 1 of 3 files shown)


skylot:
@sergey-wowwow I merged your commit and made some changes to fix tests and generation of R class.
Short description of current R class processing:
- if R class not found in app package (read from manifest) it will be generated
- all resources fields will be added to that class
- in result, all resources ids will be reference to that R class
- if original R class is located in another package it will not be used (actually there are can be int array fields which not redirected for now - need to fix)
- all const fields replacement (including resources) can be disabled using `--no-replace-consts` option

I hope it will fix your issues with resource ids replacements

PR #378, opened 2018-10-25
make the import class name clickable

jpstotz:
Originally I wanted to make the complete class name including package name clickable. But I did not manage to do so as it seems like I can not specify the clickable text section manually. Now at least the class name is clickable

### DIFF
--- a/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
+++ b/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
@@ -91,7 +91,24 @@ public class ClassGen {
 			Collections.sort(sortImports);
 
 			for (String imp : sortImports) {
-				clsCode.startLine("import ").add(imp).add(';');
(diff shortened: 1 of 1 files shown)


jpstotz:
This "commitlint" checker complained that no "type" was set. As the "commitlint" project does not provide a documentation when to use which type I just used "feat" (in the meaning of "new feature") - just to make this silly checker happy.

If you which to have a different type please tell me.

skylot:
@jpstotz thank you! I totally forget about link class names in imports :(
I agree that link for class with the package is better, but for now, it is not clear to me how to do it. But I will try!

About commitlint: recently I started to use semantic release approach for bundle release automation and it requires to make the commit in a format described [here](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commit-message-format) for generate release changelog. But I still not add any guidelines about it in jadx, sorry for that :(

jpstotz:
If I understood commitlint correctly you can't just define your own types. Therefore IMHO it is commitlint' job to provide a documentation and a may be a HowTo.

Issue #363, opened 2018-09-28
？？？


jp1017:
what's this? v0.8.0, only the jadx-dui.exe file.

![image](https://user-images.githubusercontent.com/7868514/46185400-33bf5080-c30c-11e8-8cf1-ca4f4e8c2edd.png)

Code2Life:
same issue

Code2Life:
seems to be i18n problem, changing language could be a temporary solution~
![image](https://user-images.githubusercontent.com/14833440/46721148-ebf1df00-cca4-11e8-80dd-3315faaad913.png)

jp1017:
Thanks for https://github.com/skylot/jadx/pull/386

PR #386, opened 2018-11-07
I18n resource loading corrected

jpstotz:
The i18n resource loading was a bit hacky especially the `convertCharset(String)` method.
This is the main reason for issue #363. With this patch I was even able to load the Chinese language resources on my European Windows system. Before it was only shown like in the screenshots of #363.


skylot:
@jpstotz thanks for your help, much appreciated!

PR #396, opened 2018-11-22
fix: return type lost after type inference

soldierkam:
The problem is described via test case. I'm not sure if it does break smth else.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeInference.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeInference.java
@@ -56,11 +56,13 @@ public class TypeInference extends AbstractVisitor {
 		}
 		ArgType type = assign.getType();
 		for (RegisterArg arg : useList) {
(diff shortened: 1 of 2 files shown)


skylot:
@soldierkam thank you for a nice test, it really helps. Type inference fix not so good, but it works and doesn't break other tests :)
Anyway, current type inference approach is bad because it doesn't have enough information to make correct desicion and right now I am working on the solution. I will collect type constraints for every variable and on type update make a strict check using such constraints, also update will be cross variable and recursive which gives more powerful type propagation and inference. I commit the current state of this work to [type-inference-wip](https://github.com/skylot/jadx/tree/type-inference-wip) branch.

sergey-wowwow:
@skylot 
How far the new approach from being finished? Can it already be used to get sources with bigger quality?

skylot:
@sergey-wowwow almost finished, only 2 tests fail for now, but these tests with try/catch :( 
So you can try, but errors will occur.

sergey-wowwow:
@skylot
Do they also fail for master branch? Or only here?

PR #417, opened 2018-12-23
Forcibly use SourceFileName as name or prefix of classes when deobfuscation is enabled.

unoexperto:
Hi @skylot ! I didn't know how to contact you to ask if this change looks reasonable to you hence PR.

I find this modification very useful when refactoring decopiled apps. It covers two cases:

1) It forces renaming of classes to source file name if deobfuscation is enabled. Current master does that onlt if original name matches [minLen..maxLen] condition;
2) Often ProGuard slices one original class to multiple pieces. Having number of classes named as `source-file-name_obfuscated-class-name` simplifies process of refactoring because you're instantly see that X number of classes really came from one file before.

Another change I did was in `getAliasFromSourceFile` where you check duplicate of alias by iterating over `clsMap.values()`. I find this filter to be too strict because often there are classes with same name in different packages/dependencies (e.g. `Validation` in Apache Commons and in Facebook SDK).

### DIFF
--- a/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java
+++ b/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java
@@ -10,7 +10,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 
+import jadx.core.dex.nodes.*;
+import jadx.core.utils.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


jpstotz:
I would only consider the SourceFileName attribute if it ends in ".java" - otherwise the value has been manipulated and has nothing to do with the class you are in.

Second I would only consider the SourceFileName attribute for regular (not inner) classes as inner classes usually have a different name.

unoexperto:
@jpstotz Hmm I was under impression that `AType.SOURCE_FILE` attribute contains name without extension. Perhaps it's stripped somewhere before reaching Deobfuscator. I'll check.

I agree on second point. This patch doesn't touch inner classes.

skylot:
@unoexperto SourceFileName due to obfuscators and protectors can *literally* contains anything (somewhere I saw a random binary data in it). And jadx has `UseSourceNameAsClassAlias` option to ignore such garbage. So please don't ignore or forcing this option.
Other your ideas are great, but implementation not so good. I often get class names like `Filename_Filename` and inner classes always renamed to `C000**` even if the original name was good.
About alias check: you also need to check inner class names for clashes.

In conclusion: I like what are you doing, but your code needs some adjustments. I can provide more details if you will improving your PR, but it will be hard because this part of jadx don't have any tests and it is easy to break everything :(

P.S. I hide my email from profile to force people opening issues and PR, this allows discussion between people interested in this project. Anyway, you can use email from commit history to contact me :)

its0x08 (approved):

Issue #354, opened 2018-09-03
Some APK never decompiled, jadx consumes GBs of memory


sergey-wowwow:
Hi, I've noticed that some APKs are never decompiled, but consumer 16 GBs of memory (I don't have more, so I finished it), but when I run decompilation again, it finishes successfully. 
Example of such APK: https://drive.google.com/file/d/16wImr7l426R30u4X2UznHmqJrOCL5U8w/view?usp=sharing

Output of ``` jstack ```:
```
Full thread dump Java HotSpot(TM) 64-Bit Server VM (25.74-b02 mixed mode):

"pool-1-thread-4" #13 prio=5 os_prio=0 tid=0x000000003e7f0800 nid=0x880 waiting for monitor entry [0x00000000438fe000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at jadx.core.ProcessClass.process(ProcessClass.java:26)
        - waiting to lock <0x00000004044a7f98> (a jadx.core.dex.info.ClassInfo)
        at jadx.core.ProcessClass.lambda$processDependencies$0(ProcessClass.java:51)
        at jadx.core.ProcessClass$$Lambda$14/156488300.accept(Unknown Source)
        at java.lang.Iterable.forEach(Iterable.java:75)
        at jadx.core.ProcessClass.processDependencies(ProcessClass.java:51)
        at jadx.core.ProcessClass.process(ProcessClass.java:37)
        - locked <0x00000004044aaa38> (a jadx.core.dex.info.ClassInfo)
        at jadx.api.JadxDecompiler.processClass(JadxDecompiler.java:293)
        at jadx.api.JavaClass.decompile(JavaClass.java:62)
        - locked <0x000000040180e5f0> (a jadx.api.JavaClass)
        at jadx.api.JadxDecompiler.lambda$appendSourcesSave$0(JadxDecompiler.java:201)
        at jadx.api.JadxDecompiler$$Lambda$9/1739876329.run(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)

   Locked ownable synchronizers:
        - <0x00000004017fab40> (a java.util.concurrent.ThreadPoolExecutor$Worker)

[...]

PR #418, opened 2018-12-26
Decompilation and indexing are messed-up for large apps

jpstotz:
Currently the DecompileJob generates the background jobs for decompiling the classes and then returns. 
Then the ShutdownTask is triggered which waits up to 5 minutes and then continues even if decompile jobs are still running.
If the decompilation of the remaining jobs requires more than 5 minutes Jadx is running into troubles:
This means that it starts indexing when decompilation is still running and therefore all data structures by Jadx don't cover the whole app. 

Alls this is visible to the user as the progress bar disappears way too early. Even is the CPU utilization shows that decompilation is still running.

For large APKs 5 minutes is way too less. I used the Facebook app linked in #354 and was still busy processing decompile job 5077 of 104946 when the ShutdownTask finally triggered (and that was with already re-enabled multi-threaded decompilation in MainGui.resetCache() - BTW: did you set the decompilation threads to one only because of memory problems or are there real multi-threading issues in Jadx decompilation?).

Therefore I increased the wait time to 5 days which should be enough even for large apps.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/jobs/BackgroundJob.java
+++ b/jadx-gui/src/main/java/jadx/gui/jobs/BackgroundJob.java
@@ -44,7 +44,7 @@ public abstract class BackgroundJob {
 				public Boolean call() throws Exception {
 					runJob();
 					executor.shutdown();
-					return executor.awaitTermination(5, TimeUnit.MINUTES);
+					return executor.awaitTermination(5, TimeUnit.DAYS);
 				}
 			});
 		}


skylot:
@jpstotz thank you!!! I don't know why I changed that timeout from 1 hour to 5 minutes in commit  ea01102f1ddc204487a4b32f9c8ccc0a8565ead2.

About hardcoded single thread: there are was a lot of multi-threading issues like deadlocks and incorrect processing order. But for now, looks like most of them already solved, because I often use CLI version with 8 threads and don't see any issues. Anyway decompilation speed is very slow even in 8 threads, to improve this I already redone type inference algorithm (in type-inference-wip branch) but it is still not complete.

About large memory usage: there are a lot of memory leaks (decompilation data that not needed anymore) due to bad architecture. And it even worse in GUI version because it holds all generated code for search index and code viewer. I plan to refactor code for better unloading of unnecessary data, and for GUI I can disable search if index data is to big for store in memory.

PR #419, opened 2018-12-27
Various UI improvements

jpstotz:
- Enable multi-threaded decompiling
- Added preference for exclude certain packages from decompiling and indexing
- Show message dialog in case classes are not indexed because of low memory (just a warning in the log is IMHO a bit too less. I was personally very surprised when I saw the code for skipping class indexing in Jadx)
- Added preference for disabling indexing skipping because of low memory (for those people that are relying on the fact that all classes has been indexed


jadx-gui/src/main/java/jadx/gui/jobs/IndexJob.java
```diff
@@ -44,7 +45,7 @@ public void run() {
 						List<StringRef> lines = splitLines(cls);
 
 						usageInfo.processClass(cls, linesInfo, lines);
-						if (Utils.isFreeMemoryAvailable()) {
+						if (alwaysIndex || Utils.isFreeMemoryAvailable()) {
```
skylot: Why you add option `SkipClassIndexingIfLowMemory`? If `alwaysIndex ==true` and no free memory left jadx will just hang and GC will eat all CPU. What the purpose of this option?

jpstotz: Effectively the answer is one word: **reliability**

I am a security researcher and I often use Jadx for a quick manual analysis of apps. For doing so I check multiple points (implementations of certain interfaces or extended classes) in an app which I identify using the search function. Therefore a search function that is unreliable because it had not indexed all classes is useless for me. If Jadx is running out of heap while indexing (easy visible by the CPU utilization) than I have to quit Jadx and start over with larger heap (or exclude uninteresting packages).

skylot: Hm, so you want reliability and at the same time add an option to allow hang-up jadx. This is strange, maybe is better to improve notification (which you just did) or completely disable search and index if some classes not indexed? Maybe both?
Also I was planning to add memory usage indicator like in IntelliJ, it will help too.
Anyway I don't think disable index will help for big apps, because jadx mostly stuck at decompilation step, so we need to stop whole decompilation and show alert. Or maybe not trying to put evething in memory, but dump some generated code on disk and made search from disk.

jpstotz: I implemented the `SkipClassIndexingIfLowMemory` option before I implemented the notification window that informs the user about the skipped classes.

If you think that the message is enough I can remove the option.

May be for future work we should think about extending this message to some sort of "decompilation report" showing some statistics on the number of decompiled classes and methods and of course how may problems arised while decompilation and indexing.

BTW: In my tests the decompilation stage with 12 threads and 12GB heap looked quite stable even for large apps, however the indexing stage caused a lot of errors and then stuck at 99% (e.g. using the app linked here: https://github.com/skylot/jadx/issues/354#issuecomment-435513222). Most likely the indexing errors are just side effects from decompilation errors, but I did not had the time to investigate this in deep.

skylot: Yes, please remove this option :)

And I agree about a report, I made something similar for CLI version, see [this method](https://github.com/skylot/jadx/blob/21e11c1d47d955d6316295f67736d4b6a89114b0/jadx-core/src/main/java/jadx/api/JadxDecompiler.java#L273).
Report example:
```
WARN  - Found 12 references to unknown classes
ERROR - 11 errors occurred in following nodes:
ERROR -   Method: android.support.v7.widget.ActionMenuView.onMeasureExactFormat(int, int):void
ERROR -   Method: com.google.android.gms.dynamite.zzf.zza(android.content.Context, java.lang.String, com.google.android.gms.dynamite.zzi):com.google.android.gms.dynamite.zzj
.....
WARN  - 3121 warnings in 690 nodes
ERROR - finished with errors
```

skylot:
@jpstotz thanks! I merged your commits.

PR #423, opened 2019-01-06
Added temporarily node hiding

JannisKirschner:
Right clicking on a node in the jtree gets it removed, making it disapear until a tree reload. 
This makes handling big projects cleaner where you want to exclude certain packages.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/MainWindow.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/MainWindow.java
@@ -311,6 +311,15 @@ public class MainWindow extends JFrame {
 		}
 	}
 
+	private void treeRightClickAction(){
+		try {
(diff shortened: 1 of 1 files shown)


skylot:
@JannisKirschner I understand your idea, but the implementation is not user-friendly. The user can accidentally right-click on the package and it will disappear without a way to undo or revert, also there are no "reload" button, so this is really confusing. I suggest to show menu on right click with a checkbox 'Exclude package', excluded packages will be grayed out and not disappear completely, so a user can restore them. Also, this action can be combined with recently added 'exclude packages' option by @jpstotz in #419, so this will save memory and speed up decompilation.

JannisKirschner:
good idea! I'll look into it :)

skylot:
Implemented in #477

PR #427, opened 2019-01-10
String concatenation visitor: fix for ArrayIndexOutOfBoundsException

jpstotz:
My log file was often very polluted by the string concatenation converter throwing ArrayIndexOutOfBoundsException when the StringBuilder chain contains calls not only to <init> and append but also to `toString()` or any other method of StringBuilder that does not take an argument.

This pull request limit the chain to process only constructor class and calls to append.


skylot:
@jpstotz thank you for fix!
Actually, there are tons of issues with StringBuilder conversion. And I can't force myself to fix them, because this code so cryptic it is hard to understand how it works also it lack tests. I hope someday I will rewrite it completely :)
So can you add JUnit test to reproduce this issue? Or java sample method, so I can add test myself?

jpstotz:
@skylot You are right, JUnit test cases are a good thing. It turned out that I had a major misunderstanding regarding the visitor code which ended up it totally defect code created by it. 

After creating the matching unit test I understand it much better. I think I have covered  that major cases with the unit test. I wasn't sure where to put the unit test (which package). Therefore feel free to move it where you think it belongs to.

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -201,7 +204,15 @@ private static InsnNode convertInvoke(MethodNode mth, InsnNode insn) {
 						}
 
 						for (; argInd < len; argInd++) {  // Add the .append(xxx) arg string to concat
-							concatInsn.addArg(chain.get(argInd).getArg(1));
+							InsnNode node = chain.get(argInd);
+							MethodInfo method = ((CallMthInterface) node).getCallMth();
```
skylot: Add `instanceof` check before the cast, it will make code much safer.
And next time instead of merging just rebase your commits on latest master, also fix for the previous commit is better to squash together, so you will get 2 nice commits on top of master :)

jpstotz: The missing `instanceof` check is on purpose. If we encounter such a case most likely can't simplify the chain. It will then raise an Exception so that anybody may notice it and we can take a closer look onto it. Therefore the effect is the same the only difference is that without the instanceof check nothing is logged.

Rebase: According what I have read you should not do this once you have pushed your commits to a public repo. And as the branch this PR bases on is public... 

Therefore I will keep on using git as I do - at least most of the times it does what I want. 
If they develop such a complex version control system they should make it more usable and intuitive to use.

PR #431, opened 2019-01-16
APK signature check v1/v2 using the apksig library from Google

jpstotz:
Perform APK signature verification vor v1 and v2 signatures and show the result in a new tree node.
Also the errors and warnings generated by apksig library are shown in a simple HTML based view.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/treemodel/ApkSignature.java
+++ b/jadx-gui/src/main/java/jadx/gui/treemodel/ApkSignature.java
@@ -1,0 +1,122 @@
+package jadx.gui.treemodel;
+
+import com.android.apksig.ApkVerifier;
+import com.android.apksig.apk.ApkFormatException;
+import jadx.gui.utils.CertificateManager;
+import jadx.gui.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.swing.*;
+import java.io.File;
+import java.security.cert.Certificate;
+import java.util.List;
+
+public class ApkSignature extends JNode {
+
+	private static final Logger log = LoggerFactory.getLogger(ApkSignature.class);
+	private static final ImageIcon CERTIFICATE_ICON = Utils.openIcon("certificate_obj");
+
+	private final transient File openFile;
+	private String content = null;
+
+	public static ApkSignature getApkSignature(File openFile) {
+		return new ApkSignature(openFile);
+	}
+
+	public ApkSignature(File openFile) {
+		super();
+		this.openFile = openFile;
+	}
+
+	@Override
+	public JClass getJParent() {
+		return null;
+	}
+
+	@Override
+	public Icon getIcon() {
+		return CERTIFICATE_ICON;
+	}
+
+	@Override
+	public String makeString() {
+		return "APK signature";
+	}
+
+	@Override
+	public String getContent() {
+		if (content != null)
+			return this.content;
+		ApkVerifier verifier = new ApkVerifier.Builder(openFile).build();
+		try {
+			ApkVerifier.Result result = verifier.verify();
+			StringBuilder str = new StringBuilder();
+			str.append("<h1>APK signature verification result:</h1>");
+
+			str.append("<p><b>");
+			if (result.isVerified()) {
+				str.append("Signature verification succeeded");
+			} else {
+				str.append("Signature verification failed\n");
+			}
+			str.append("</b></p>");
+
+			writeIssues(str, "Errors", result.getErrors());
+			writeIssues(str, "Warnings", result.getWarnings());
+
+			if (result.isVerifiedUsingV1Scheme()) {
+				str.append("<h2>APK signature v1 found</h2>\n");
+
+				str.append("<blockquote>");
+				for (ApkVerifier.Result.V1SchemeSignerInfo signer : result.getV1SchemeSigners()) {
+					str.append("<h3>Signer " + signer.getName() + " (" + signer.getSignatureFileName() + ")</h3>");
+					writeCertificate(str, signer.getCertificate());
+					writeIssues(str, "Errors", signer.getErrors());
+					writeIssues(str, "Warnings", signer.getWarnings());
+				}
+				str.append("</blockquote>");
+			}
+			if (result.isVerifiedUsingV2Scheme()) {
+				str.append("<h2>APK signature v2 found</h2>\n");
+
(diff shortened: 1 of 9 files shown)


skylot:
@jpstotz looks good.
List of questions/suggestions for future improvements:
- can we delete `Certificate` node, because it duplicates functionality?
- don't show `Apk Signature` for not apk files (dex, jar, etc)
- use common font and color scheme from jadx settings for HtmlPanel
- maybe improve strings localization (not all can be translated now)
- maybe combine several warnings and errors by same type. For example many apps have a lot of such warnings with a tons of same strings:
```
META-INF/CHANGES not protected by signature. Unauthorized modifications to this JAR entry will not be detected. Delete or move the entry outside of META-INF/.
META-INF/README.md not protected by signature. Unauthorized modifications to this JAR entry will not be detected. Delete or move the entry outside of META-INF/.
... repeated for every files
```
better to restructure it to:
```
Entries not protected by signature. Unauthorized modifications to this JAR entry will not be detected. Delete or move the entry outside of META-INF/ :
 - META-INF/CHANGES
 - META-INF/README.md
 - ... list of other files
```

jadx-gui/src/main/java/jadx/gui/treemodel/ApkSignature.java
```diff
@@ -0,0 +1,122 @@
...
+
+	private final transient File openFile;
+	private String content = null;
+
+	public static ApkSignature getApkSignature(File openFile) {
+		return new ApkSignature(openFile);
+	}
+
+	public ApkSignature(File openFile) {
+		super();
```
skylot: `super()` can be removed

jadx-gui/src/main/java/jadx/gui/treemodel/ApkSignature.java
```diff
@@ -0,0 +1,122 @@
...
+
+	@Override
+	public String getContent() {
+		if (content != null)
+			return this.content;
+		ApkVerifier verifier = new ApkVerifier.Builder(openFile).build();
+		try {
+			ApkVerifier.Result result = verifier.verify();
+			StringBuilder str = new StringBuilder();
+			str.append("<h1>APK signature verification result:</h1>");
```
skylot: maybe add to localization files (use NLS) for this string and all strings below

jadx-gui/src/main/java/jadx/gui/treemodel/ApkSignature.java
```diff
@@ -0,0 +1,122 @@
...
+					str.append("<h3>Signer " + (signer.getIndex() + 1) + "</h3>");
+					writeCertificate(str, signer.getCertificate());
+					writeIssues(str, "Errors", signer.getErrors());
+					writeIssues(str, "Warnings", signer.getWarnings());
+				}
+				str.append("</blockquote>");
+			}
+			this.content = str.toString();
+		} catch (Exception e) {
+			log.error(e.getMessage(), e);
```
skylot: The exception can be printed directly into text for easier issue report

jadx-gui/src/main/java/jadx/gui/treemodel/ApkSignature.java
```diff
@@ -0,0 +1,122 @@
...
+		str.append("<pre>" + certMgr.generateHeader() + "</pre>");
+		str.append("<pre>" + certMgr.generatePublicKey() + "</pre>");
+		str.append("<pre>" + certMgr.generateSignature() + "</pre>");
+		str.append("<pre>" + certMgr.generateFingerprint() + "</pre>");
+		str.append("</blockquote>");
+	}
+
+	private void writeIssues(StringBuilder str, String issueType, List<ApkVerifier.IssueWithParams> issues) {
+		if (issues.size() > 0) {
+			str.append("<h3>" + issueType + "</h3><pre>");
```
skylot: Please use `append` instead of `+` for string concatenation using StringBuilder

jadx-gui/src/main/java/jadx/gui/treemodel/JRoot.java
```diff
@@ -37,6 +37,11 @@ public final void update() {
 			add(jRes);
 		}
 
+		ApkSignature signature = ApkSignature.getApkSignature(wrapper.getOpenFile());
+		if (signature!=null) {
```
skylot: Please use code formatting to insert necessary spaces.

jadx-gui/src/main/java/jadx/gui/ui/TabbedPane.java
```diff
@@ -11,16 +11,13 @@
 import java.util.List;
 import java.util.Map;
 
+import jadx.gui.treemodel.*;
```
skylot: I suggest using implicit imports (don't use `.*`)

jpstotz: Right. IntelliJ seem to have a strange understanding of optimized import :(

jpstotz:
The signature page is now localized as long as the text is not coming from the apksig library.
Additionally HTML escaping was required. I found a nice HTML escaper in apache commons text that extends a String builder with escaping.

> can we delete Certificate node, because it duplicates functionality?

For APK files yes, for plain JAR files it makes sense to keep the certificate nodes.

> don't show Apk Signature for not apk files (dex, jar, etc)

Done.

> use common font and color scheme from jadx settings for HtmlPanel

JEditorPane does not have a color scheme system. Applying a font to JEditorPane does not have any effect.

> maybe improve strings localization (not all can be translated now)

Done

> maybe combine several warnings and errors by same type. For example many apps have a lot of such warnings with a tons of same strings:

There is no general way to do so. I implemented it for the issue type `JAR_SIG_UNPROTECTED_ZIP_ENTRY`. The code that does so is not nice but it works.

skylot:
@jpstotz I found a strange thing in your commit. You add google maven repository but it doesn't contain the version of apksig you used (2.3.0), and this version fetched from mavenCentral. The latest stable version from google repository is 3.3.0 ([see here](https://mvnrepository.com/artifact/com.android.tools.build/apksig?repo=google)) and this version also work without any changes in code.
So, this is a typo in version number or you did it intentionally and library can't be upgraded for some reason?

jpstotz:
@skylot That is typo. Originally I used a different library from the Google repo until I noticed that the signature verification functionality wasn't implemented in the library but in the apksig library which was a dependency of that library. When I changed to the apksig library I missed the fact that it is available in multiple repositories and took the wrong one.

PR #455, opened 2019-02-22
fix(gui): Allow APK files without code (no contained dex files)

jpstotz:
APKs without code are getting more and more common.
Originally this was only possible for apps that were implemented purely in native code (never encountered such an app) but nowadays because of the [App Bundle technology](https://developer.android.com/platform/technology/app-bundle/) a lot of apps are constructed out of multiple APK files (e.g. base bundle with the code, the images, the language strings and finally the native libraries).

This PR makes it possible to open such "resource APKs" in Jadx.


skylot:
@jpstotz thanks. 
But I don't think many people use jadx for resource decoding, because there are many unresolved issues and looks like `apktool` is a better choice for this :)

jpstotz:
@skalot Don't forget the AndroidManifest (was also not accessible without this PR). A quick glance into it is IMHO simpler using Jadx than with apktool.

Issue #95, opened 2015-11-02
Support for AAR files


rekire:
I really like this viewer, but why is there no out of the box support for aar files? They also includes a *`classes.jar`*.

rekire:
Cool thank you! When do you publish a stable version with this patch?

drcuong:
after version 0.61 this function doesn't work anymore, would you mind to fix this @skylot

PR #464, opened 2019-03-09
fix: Restore support for AAR files (issue #95)

jpstotz:


skylot:
@jpstotz thank you for the fix.

PR #466, opened 2019-03-13
fix: replace dot character with underscore in style resource name

tRuNKator:
Hello, current generated field name for style resources in R class
```
/* added by JADX */
/* renamed from: Base.Widget.AppCompat.ImageButton */
public static final int f3775BaseWidgetAppCompatImageButton = 2131624131;
```
Should be:
```
public static final int Base_Widget_AppCompat_ImageButton = 2131624131;
```


skylot:
@tRuNKator thank you!

PR #470, opened 2019-03-16
Fix anonymous inner class naming

mingyuan-xia:


skylot:
@mingyuan-xia I understand that splitting class names by '$' is not a good approach, but jadx will revert class from inner to top if expected parent class not really exists (see [here](https://github.com/skylot/jadx/blob/7654661b776c76abcb97915b46b9d1ace41c7fc9/jadx-core/src/main/java/jadx/core/dex/nodes/DexNode.java#L69)). Also classes names started with a number will be prefixed with 'AnonymousClass'.
So I don't see any improvements in your commit.

PR #477, opened 2019-03-20
GUI: disable package if "excluded"

asashour:
This is related to #423.

The changes does the following:
- On right-click of a package, it opens a menu to exclude/unexclude.
- Propagates this change to the settings.
- Refresh the tree.
- Excluded packages don't show children.


skylot:
@asashour thank you for your work, this is a nice feature and implementation!

PR #487, opened 2019-03-21
fix: resource configuration qualifiers

tRuNKator:
Currently Jadx does not properly generates configuration qualifiers. [Example apk](https://drive.google.com/file/d/1CwUTCvJZv4iWqVuKQ4aRNfZ94qAO-yLz/view?usp=sharing). In values-h1dp-land-v13 You can see duplicate resource values (there not only duplicates, but also resources from other configurations) like
```
    <dimen name="PopupLayoutList_height">270dp</dimen>
    <dimen name="PopupLayoutList_height">300dp</dimen>
```
I fixed it. Almost all the code for fix was copied from apktool

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ParserStream.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ParserStream.java
@@ -1,5 +1,6 @@
 package jadx.core.xmlgen;
 
+import java.io.EOFException;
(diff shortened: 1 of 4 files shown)


tRuNKator:
Sorry, without tests

skylot:
@tRuNKator thank you for your help!

PR #489, opened 2019-03-22
gui: better to show java version, instead of VM version

asashour:
In the GUI "About" dialog: so the users sees something like `1.8.0_202` instead of `25.202-b08`


skylot:
This is funny, thanks!
There also properties for `java.runtime.version` and `java.specification.version` - tons of versions in different formats. And in java 9+ you can do `Runtime.Version version = Runtime.version()` which help get parts of these strings )))

PR #492, opened 2019-03-22
fix: R class fields alias

tRuNKator:
Hello, when deobfuscation processes fields, it does not check whether alias is already set in `FieldInfo`. As a result, you can see that the class R fields are renamed twice.
Without deobfuscation (alias set by `AndroidResourcesUtils`):
```
    public static final class array {
        /* renamed from: enum */
        public static final int pref_widget_font_sizes = 2131689474;
        /* renamed from: ll1l */
        public static final int widget_theme_entries = 2131689473;
        /* renamed from: null */
        public static final int pref_widget_font_families = 2131689475;
    }
```
With deobfuscation (alias set by `AndroidResourcesUtils` and then replaced by `Deobfuscator`):
```
    public static final class array {
        /* renamed from: enum */
        public static final int f2654enum = 2131689474;
        /* renamed from: ll1l */
        public static final int widget_theme_entries = 2131689473;
        /* renamed from: null */
        public static final int f2655null = 2131689475;
    }
```

### DIFF
--- a/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java
+++ b/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 
+import jadx.core.dex.attributes.AFlag;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


skylot:
This will work, but I suggest a more clear approach: in AFlag add something like `DONT_RENAME`, for generated or renamed fields in AndroidResourcesUtils add this flag (after check `NameMapper.isValidIdentifier`), and in Deobfuscator don't run rename if this flag exists. For methods needed to work with flags (`add` and `contains`) see `IAttributeNode` interface which `FieldNode` implements.

tRuNKator:
> 
> 
> This will work, but I suggest a more clear approach: in AFlag add something like `DONT_RENAME`, for generated or renamed fields in AndroidResourcesUtils add this flag (after check `NameMapper.isValidIdentifier`), and in Deobfuscator don't run rename if this flag exists. For methods needed to work with flags (`add` and `contains`) see `IAttributeNode` interface which `FieldNode` implements.

I agree, this approach is better. In this PR, I can implement a check of the `DONT_RENAME` flag for fields only. For methods I think it is better to make a separate PR

PR #493, opened 2019-03-22
test: JUnit 5

asashour:


asashour:
This is now fine, but the only thing I couldn't do, is to remove JUnit 4.12 from the dependencies. It is still shown.

skylot:
I see, junit 4 it is fetched by Spock framework. Maybe I will remove it later, but this needs to rewrite some tests. I like Spock for writing tests with "tables" like [here](https://github.com/skylot/jadx/blob/9c90699c409e5455841868c7f338af49943b6ad8/jadx-core/src/test/groovy/jadx/tests/TestSignatureParser.groovy#L15) ))

asashour:
I tried to exclude JUnit 4, but the Spock framework fails.

At the current PR state, both JUnit 4 and JUnit 5 coexist, which is Ok. When Spock upgrades, it would be automatically solved.

PR #496, opened 2019-03-22
test: migrate all assertions to JUnit5

asashour:
And putting a test method to ensure this in the future.

This applied to only java files, not groovy.


skylot:
@asashour sorry I just pushed commits with tests and they violate your new checks. Can you fix them to pass build?

asashour:
My pleasure!

I was going crazy, since they work locally!

Thanks so much for your quick involvement.

skylot:
Thanks!

PR #503, opened 2019-03-23
style: use character instead of string, for better performance

asashour:
Please note I faced some issues updating local Git, and I am suspecting that the `master` history was changed (by forcing update).

There seems to be two commits with different IDs:

image

And one creating original PR, it says "can't merge":

image


skylot:
@asashour sorry for that. I know force update is bad and I will try to not use again (

asashour:
No worries, I thought fixing it would be very complex, but it wasn't.

As you know, you can revert any commit (by merge), because force update for a public repository makes all branches not usable.

PR #509, opened 2019-03-25
chore: add .gitattributes

asashour:
This ensures that the end of line of files is automatically handled.

This was detected during comparison of source code, and all lines are changed (due to EndOfLine).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*        text=auto


skylot:
@asashour I don't like auto conversion in any way. So I tried to describe and check style rules using [EditorConfig](https://editorconfig.org/) project see PR #510.

PR #510, opened 2019-03-25
style: enforce strict style rules with editorconfig

skylot:
* use [editorconfig](https://EditorConfig.org) for describe project format and style rules
* check these rules on build using [gradle plugin](https://github.com/ec4j/editorconfig-gradle-plugin), build fail on violations


asashour:
Any reason why this is not merged?

PR #512, opened 2019-03-25
fix: auto detect generics type

asashour:
A step forward to detect generics type, otherwise compilation errors occur from the generated code.

Overview of the changes:
1. `GenericsTypeVisitor` was introduced
2. Java Reflection was used to know the generic types.
3. `mth.unloadInsnArr();` was moved from `DebugInfoVisitor` to `GenericsTypeVisitor`

Appreciate looking into the changes, and suggest/commit any necessary modifications.


skylot:
@asashour sorry but I don't think I will merge this, here some reasons:
* **reflection**: it will be hard to add android classpath. I know we can just add stub android.jar into classpath but it will clash with original JRE classes and it will be a mess. Also, reflection is slow, so we will need some caching, etc. So it is not a good way. 
I was planning to add generic and other information for classes and methods into current [ClspGraph](https://github.com/skylot/jadx/blob/4ce5cc84924c90ec3d216495a72ebb9c130eef1f/jadx-core/src/main/java/jadx/core/clsp/ClspGraph.java#L23), now it contains only class hirarchy and stored in binary file which is small and fast to load.
* this **must** be integrated into type inference. Because cast doesn't mean that type is defined, we must check if this type is acceptable, in simple cases it looks easy, but it is not in general. Please trust me, I rewrite type inference algorithm from scratch already 4 times and it is still failing in some cases. Latest attempt you can see in `type-inference-wip` branch, just check the size of the [diff](https://github.com/skylot/jadx/compare/type-inference-wip).
* current implementation does not work with classes from decompiling code (need separate analyzer) - not a big deal due to above reasons, just mention )

Anyway, thank you for your attempt! Next time if you plan to fix any issue please tell me, I will try to describe how to better implement that and provide all needed information.

P.S. Looks like I need to write development plan because I have tons of ideas and features which I want to implement but some of them depends on each other. For example, I don't implement proper generic propagation, because previous type inference was so bad that base type was often incorrect (

asashour:
Thanks for feedback, I understand that I am just starting to know some part of the code and its design.

As you said, it would be a great step, to communicate your ideas, so whoever wants to help, can pick one peace.

Also, a single javadoc line for some classses (E.g. Visitors) would be also helpful.

Issue #409, opened 2018-12-10
Use of unnecessary "^ true" for boolean expressions


sergey-wowwow:
Hi,
I've noticed that thing in master and type-inference-wip branches. For example
``` com.paypal.android.p2pmobile.wallet.banksandcards.utils.EnterCardFragmentUtils ```
```java
    public static boolean isCardCSCValid(@android.support.annotation.Nullable com.paypal.android.foundation.wallet.model.FinancialInstrumentMetadataDefinition financialInstrumentMetadataDefinition, java.lang.String str) {
        boolean z = false;
        if (financialInstrumentMetadataDefinition == null) {
            return false;
        }
        com.paypal.android.foundation.wallet.model.FinancialInstrumentMetadataAttribute cvv = financialInstrumentMetadataDefinition.getCvv();
        if (cvv == null) {
            return true;
        }
        if (android.text.TextUtils.isEmpty(str)) {
            return cvv.isRequired() ^ true; // <<<
        }
```

It seems that ``` booleanExpr ^ true ``` is useless and can be removed, but ``` booleanExpr ^ false ``` could be replaced with ``` !booleanExpr ```

APK: https://drive.google.com/open?id=1pdmRyrp6rDEoYsTJn5RLhv39xXRq5Ksh

PR #514, opened 2019-03-25
test: add test case for #409 and "Xor with boolean"

asashour:


skylot:
Hm, why are you add two tests? I don't see a difference between them. I think bytecode in both cases will be same )

asashour:
Bytecode in .class is a little different.

Actually, the reason is there are two things in this test:

1. `^ 1` should be `^ true`
2. `test2() ^ true` should be `!test2()`

and thought having two tests to tackle each point separately.

Where should I ask for implementation hints, in the issue raised, or should I raise a new one?

skylot:
I think in both cases `expr ^ true`should be replaced to `!expr` if `expr` type is boolean. And it doesn't matter if `expr` is a variable or method invoke.

Fixing is rather simple: in `SimplifyVisitor` replace `ARITH` insn (where `op == XOR` and one of args is literal and equal 1) with `NOT` instruction (see [`simplifyArith`](https://github.com/skylot/jadx/blob/72b2663949555b03aeb864775d84f163b91bcea7/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java#L228) method as a reference)

PR #516, opened 2019-03-26
fix: xor with boolean (#409)

### LINKED ISSUE #409: Use of unnecessary "^ true" for boolean expressions

asashour:
Fixes #409


jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -245,9 +251,22 @@ private static InsnNode simplifyArith(InsnNode insn) {
 			// fix 'c + (-1)' => 'c - (1)'
 			if (arith.getOp() == ArithOp.ADD && lit < 0) {
 				return new ArithNode(ArithOp.SUB,
-						arith.getResult(), insn.getArg(0),
+						arith.getResult(), arith.getArg(0),
 						InsnArg.lit(-lit, litArg.getType()));
 			}
+			InsnArg firstArg = arith.getArg(0);
+			if (arith.getOp() == ArithOp.XOR && firstArg.getType() == ArgType.BOOLEAN) {
+				if (lit == 1 && firstArg.isInsnWrap()) {
```
skylot: So, you make it just for wrapped instructions, and it will not work for simple args (variables). There are test case:
```java
		public boolean test(boolean v) {
			return v ^ true; // expected 'return !v;'
		}
```
For this just do:
```java
if (lit == 1) {
  InsnNode not = new InsnNode(InsnType.NOT, 1);
  not.setResult(arith.getResult()); // note I return original result it is better because it will not broke SSA analysis
  not.addArg(firstArg);
  return not; 
}
```
For `lit == 0` you can return `MOVE` instruction

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -256,16 +257,12 @@ private static InsnNode simplifyArith(ArithNode arith) {
...
-					InsnNode wr = ((InsnWrapArg) firstArg).getWrapInsn();
-					InsnNode not = new InsnNode(InsnType.NOT, 1);
-					not.setResult(wr.getResult());
-					not.addArg(firstArg);
-					return not; 
-				}
-				if (lit == 0 && firstArg.isInsnWrap()) {
-					return ((InsnWrapArg) firstArg).getWrapInsn();
-				}
+				InsnNode node = new InsnNode(lit == 0 ? InsnType.MOVE : InsnType.NOT, 1);
```
skylot: Please ignore cases other then `lit == 0` and `lit == 1`.
I know such cases seem impossible, but better treat them well and don't produce wrong code :)

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -256,16 +257,12 @@ private static InsnNode simplifyArith(ArithNode arith) {
...
+				RegisterArg result = arith.getResult();
+				result.setType(ArgType.BOOLEAN);
```
skylot: Don't change types, made checks for this before modifications.

asashour: Actually, the type of the `firstArg` is checked, not `result`.

If this wasn't changed, then the type is `int`, and the code in `InsnGen.java` won't work:

    char op = insn.getResult().getType() == ArgType.BOOLEAN ? '!' : '~';

skylot:
@asashour thank you!
You changing code so fast, this is great! And I think I need to merge `type-inference-wip` branch or it will be hard to resolve conflicts. Be prepared for changes ))

asashour:
I would say, once you have a green build, then merge it.  There are not so many resources available, to support two evolving branches.

PR #520, opened 2019-03-26
test: make test @NYI instead of commenting assertions

asashour:
`private` was made `public` to fix Eclipse warning. It doesn't affect the test behavior.

### DIFF
--- a/jadx-core/src/test/java/jadx/tests/integration/conditions/TestConditions2.java
+++ b/jadx-core/src/test/java/jadx/tests/integration/conditions/TestConditions2.java
@@ -1,7 +1,12 @@
 package jadx.tests.integration.conditions;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.not;
+
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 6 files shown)


jadx-core/src/test/java/jadx/tests/integration/conditions/TestConditions2.java
```diff
@@ -24,11 +29,12 @@ public void testComplexIf(String a, int b) {
...
 	@Test
+	@NotYetImplemented
 	public void test() {
 		ClassNode cls = getClassNode(TestCls.class);
 		String code = cls.getCode().toString();
 
-//		assertThat(code, containsString("return;"));
-//		assertThat(code, not(containsString("else")));
+		assertThat(code, containsString("return;"));
+		assertThat(code, not(containsString("else")));
```
skylot: I think this is incorrect change, because original code checks only decompilation and recompilation without any manual code checks. But after change decompilation can fail but test will not show that.
I suggest just remove commented code, because checks is useless, and generated code is a same as original.
(NYI annotation also must be removed)

asashour: I see, I thought it is a complex `return` that should be inline into one line.

PR #525, opened 2019-03-27
fix(gui): add missing translations

asashour:
Google Translate was used for both Spanish and Chinese.

Tests are included to ensure all files exactly have match prefixes of all lines.


skylot:
@asashour I don't think Google Translate is a good source for translation. I think English is better than a bad translation. So I suggest you split this PR into fixes and new translations and for translations part, we will wait until someone with knowledge of these languages can review them.

asashour:
BTW, there is no fallback to English, if the value is not defined. I guess this should be changed.

skylot:
@asashour I checked and fallback is working. Looks like it doesn't work if you assign empty string instead translation, so it uses empty strings. You can fix it if you really want.
Also you can use `java.util.Properties` for parse files in tests and don't write a parser from scratch.

asashour:
I see.

Now, all untranslated values are commented, so it is easier for the translator to see what is missing.

Regarding using `java.util.Properties`, it won't help much to ensure one-to-one line comparison, and also considering comments.

PR #527, opened 2019-03-27
chore: use Path instead of File

asashour:
Java NIO Path is the new way to handle files.

This is a first step to move forward, more steps could be to change InputFile to have internal Path, but I am not sure about backward compatibility constraints for the API.

BTW, would it be helpful if Java 11 is used?


skylot:
@asashour thanks. New API is nice, but I don't think it will bring any improvements, errors after incorrect refactoring is more possible :)
About Java 11: some people still using Java 8 (because some apps not work on Java 11), also I know android projects which uses jadx (like [ShowJava](https://github.com/niranjan94/show-java/)) and they limited to java 8 api. So jadx is stick to Java 8 for a while.

Issue #530, opened 2019-03-28
String constructor with byte array


asashour:
As deduced from #529, there is also a potential to make String constructor beautiful.

```
	public static class TestCls {
		public String tag = new String(new byte[] {'a', 'b', 'c'});
	}

	@Test
	public void test() {
		ClassNode cls = getClassNode(TestCls.class);
		String code = cls.getCode().toString();

		assertThat(code, containsOne("abc"));
	}

```

asashour:
I am trying to fix this, but I counter issue, the corresponding output could be something like:

    new String("abc".getBytes(), StandardCharsets.UTF_8);

And I don't know how to construct a MethodInfo the the String(byte[]) constructor if it is not found in `dex`.

skylot:
Looks like you need to implement it youself. Add to `MethodInfo` class something like this:
```java
public static MethodInfo externalMth(ClassInfo declaredCls, String name, List<ArgType> args, ArgType retType) {
```

sergey-wowwow:
Seems that it would be better to begin from constant concatenations like
```java
String value = "a" + "b";
```
sometimes ``` "a" ``` could be a ``` static final String SMTH = "a"; ``` value, and when ``` --no-replace-consts ``` option is enabled, the given output will be generated. But it's nicer to see 
```java
String value = "ab";
```

asashour:
> Seems that it would be better to begin from constant concatenations like
> 
> ```java
> String value = "a" + "b";
> ```
> 
> sometimes `"a"` could be a `static final String SMTH = "a";` value, and when `--no-replace-consts` option is enabled, the given output will be generated. But it's nicer to see
> 
> ```java
> String value = "ab";
> ```

I guess this is a different case, since constants are usually `String`, but the one is question is a `byte array`.

PR #531, opened 2019-03-28
fix: better String constructor from byte array (#530)

asashour:
First part of fixing the string constructor

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
@@ -95,6 +95,23 @@ public class SimplifyVisitor extends AbstractVisitor {
 				}
 				break;
 
+			case CONSTRUCTOR:
+				ConstructorInsn constructor = (ConstructorInsn) insn;
+				if (constructor.getCallMth().getDeclClass().getType().equals(ArgType.STRING)) {
+					InsnArg arg = insn.getArg(0);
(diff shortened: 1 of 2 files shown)


asashour:
Handling of non-ASCII bytes should be also made.

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -95,6 +95,23 @@ private static InsnNode simplifyInsn(MethodNode mth, InsnNode insn) {
...
+				if (constructor.getCallMth().getDeclClass().getType().equals(ArgType.STRING)) {
+					InsnArg arg = insn.getArg(0);
+					if (insn.getArgsCount() == 1 && arg.isInsnWrap()) {
+						InsnNode node = ((InsnWrapArg) arg).getWrapInsn();
+						if (node.getArg(0).getType() == ArgType.BYTE) {
+							byte[] arr = new byte[node.getArgsCount()];
+							for (int i = 0; i < arr.length; i++) {
+								arr[i] = (byte) ((LiteralArg) node.getArg(i)).getLiteral();
+							}
+							insn.setArg(0, new InsnWrapArg(new ConstStringNode(new String(arr))));
```
skylot: You need to use here unescape method to support unprinted chars:
```java
mth.root().getStringUtils().unescapeString(new String(arr)))
```
Also if all char is not printable there is not point to convert them to string.

PR #533, opened 2019-03-28
fix: better String constructor from byte and char arrays (#530)

### LINKED ISSUE #530: String constructor with byte array

asashour:
Fixes #530

`test4()` fails, because I don't know how to get the actual value of the array.

In other words: the code generated is:

```
    public static class TestCls4 {
        public String tag = new String(new byte[] {0, 1, 2, 'a', 'b', 'c'});
    }

    public TestStringConstructor$TestCls4() {
        byte[] bArr = new byte[6];
        bArr[1] = (byte) 1;
        bArr[2] = (byte) 2;
        bArr[3] = (byte) 97;
        bArr[4] = (byte) 98;
        bArr[5] = (byte) 99;
        this.tag = new String(bArr);
    }
```

And in `SimplifyVisitor`, I don't know how to the the actual values of `bArr`, not its definition.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
@@ -8,6 +8,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import jadx.core.Consts;
+import jadx.core.deobf.NameMapper;
 import jadx.core.dex.attributes.AFlag;
 import jadx.core.dex.info.ClassInfo;
 import jadx.core.dex.info.FieldInfo;
@@ -26,12 +27,12 @@ import jadx.core.dex.instructions.args.FieldArg;
 import jadx.core.dex.instructions.args.InsnArg;
 import jadx.core.dex.instructions.args.InsnWrapArg;
 import jadx.core.dex.instructions.args.LiteralArg;
-import jadx.core.dex.instructions.args.RegisterArg;
 import jadx.core.dex.instructions.mods.ConstructorInsn;
 import jadx.core.dex.instructions.mods.TernaryInsn;
 import jadx.core.dex.nodes.BlockNode;
 import jadx.core.dex.nodes.InsnNode;
 import jadx.core.dex.nodes.MethodNode;
+import jadx.core.dex.nodes.RootNode;
 import jadx.core.dex.regions.conditions.IfCondition;
 
 public class SimplifyVisitor extends AbstractVisitor {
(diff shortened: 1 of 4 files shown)


jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -95,6 +98,44 @@ private static InsnNode simplifyInsn(MethodNode mth, InsnNode insn) {
 				}
 				break;
 
+			case CONSTRUCTOR:
+				ConstructorInsn constructor = (ConstructorInsn) insn;
+				if (constructor.getCallMth().getDeclClass().getType().equals(ArgType.STRING)) {
+					InsnArg arg = insn.getArg(0);
```
skylot: Check args count before getting first. Now fail on `new String()`
Also check type of arg, because there are many string constructors.
Or better check constructor signature (`constructor.getCallMth().getShortId()`)
Also you can easily support constructor with char array: `new String(new char[]{'a', 'b', 'c'});`

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -95,6 +98,44 @@ private static InsnNode simplifyInsn(MethodNode mth, InsnNode insn) {
...
+					InsnNode node = arg.isInsnWrap()
+							? ((InsnWrapArg) arg).getWrapInsn()
+							: ((RegisterArg) arg).getParentInsn();
```
skylot: `((RegisterArg) arg).getParentInsn()` is exactly `insn` and you don't need it, only wrapped insn

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -95,6 +98,44 @@ private static InsnNode simplifyInsn(MethodNode mth, InsnNode insn) {
...
+					ArgType argType = node.getArg(0).getType();
```
skylot: Add check `node.getType() == FILLED_NEW_ARRAY`

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -95,6 +98,44 @@ private static InsnNode simplifyInsn(MethodNode mth, InsnNode insn) {
...
+					if (argType == ArgType.BYTE
+							/*|| (argType.isArray() && argType.getArrayElement() == ArgType.BYTE)*/) {
+						int unreadable = 0;
+						byte[] arr = new byte[node.getArgsCount()];
+						for (int i = 0; i < arr.length; i++) {
+							arr[i] = (byte) ((LiteralArg) node.getArg(i)).getLiteral();
+							if (arr[i] < ' ' || arr[i] > '~') {
```
skylot: Here you can use `NameMapper.isPrintableChar`

jadx-core/src/test/java/jadx/tests/integration/others/TestStringConstructor.java
```diff
@@ -0,0 +1,63 @@
...
+	@Test
+	public void test3() {
+		ClassNode cls = getClassNode(TestCls3.class);
+		String code = cls.getCode().toString();
+
+		assertThat(code, containsOne("\\u0001\\u0002\\u0003"));
+	}
+
+	public static class TestCls4 {
+		public String tag = new String(new byte[] {0, 1, 2, 'a', 'b', 'c'});
```
skylot: Looks like this test is same as in test3

skylot:
Hm, I can't reproduce your case with not inlined array creation, it must be replaced [here](https://github.com/skylot/jadx/blob/17d8516d3b5561da669a1a87549c62ac3f171cab/jadx-core/src/main/java/jadx/core/dex/visitors/ModVisitor.java#L114).

I forget to mention in review, please move most of your code from switch to separate method.

P.S. Aaand Travis doesn't want to work today either, build won't start (((

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
asashour: Type type is checked later (byte or newly added char)

jadx-core/src/test/java/jadx/tests/integration/others/TestStringConstructor.java
asashour: Not in Eclipse, actually I started with `byte[] {0, 1, 2}`, then got a failure. so I moved to `1, 2, 3`. There is a difference in tests between Eclipse and gradle. I would open a new issue for that.

For now, `test4()` has been modified, to what the original test gets in Eclipse.

skylot:
@asashour almost fixed. Add checks to don't fail on this test
```java
	public static class TestClsNegative {
		public String f1 = new String();
	}

	@Test
	public void testNegative() {
		ClassNode cls = getClassNode(TestClsNegative.class);
		String code = cls.getCode().toString();

		assertThat(code, containsOne("f1 = new String();"));
	}
```

PR #540, opened 2019-03-29
test: warn about compiler not found

asashour:
Similar to #511, but with `StaticCompiler`

### DIFF
--- a/jadx-core/src/test/java/jadx/tests/api/compiler/StaticCompiler.java
+++ b/jadx-core/src/test/java/jadx/tests/api/compiler/StaticCompiler.java
@@ -1,16 +1,5 @@
 package jadx.tests.api.compiler;
 
-import javax.tools.FileObject;
-import javax.tools.ForwardingJavaFileManager;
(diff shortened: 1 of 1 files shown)


jadx-core/src/test/java/jadx/tests/api/compiler/StaticCompiler.java
```diff
@@ -22,11 +26,17 @@
...
 
+	private static final Logger LOG = LoggerFactory.getLogger(StaticCompiler.class);
+
 	private static final List<String> COMMON_ARGS = Arrays.asList("-source 1.8 -target 1.8".split(" "));
 
 	public static List<File> compile(List<File> files, File outDir, boolean includeDebugInfo) throws IOException {
 
 		JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+		if (compiler == null) {
+			LOG.error("Can not find compiler, please use JDK instead");
```
skylot: Please throw an exception instead of logging, otherwise assert error will raise but with a different reason.

PR #543, opened 2019-03-31
gui: support project (#526)

### LINKED ISSUE #526: gui: project settings
This is meant to capture the requirements of having projects settings in the `GUI` module, it is based on the discussion of #524.

Currently, the GUI opens a single API at a time, and I guess this won't change soon. So, accordingly a project is associated with one APK.

What is needed:
1. A way to store the project
2. Projects settings, e.g. APK, open tabs, 'renamed' methods are saved.
3. User can switch between projects (saved internally in Java Preferences)

The main questions are:
1. Should we have an actual file (e.g. with `jadx` extension), or should we store this as we are using Java Preferences for the settings?
2. If a file, what would be the format, XML/Json/custom?
3. When a user changes some setting, should it automatically be saved, or there is `save project` option.

**Update**

Steps:
1. `Recent Files` would be removed, and we would have `Recent Projects`
2. There would be an option to auto-save project settings, or need explicit `Save Project` action
3. <strike>`Exclude` packages feature would be moved from `Preferences` to `Project settings`</strike>
4. The main window title would change from `jadx-gui - name.apk` to `MyProject [full_path_to_MyProject_folder] - jadx-gui` with optional `*` at the beginning when the project is not saved

asashour:
Fix #526 

This is initial change to have basic project, other properties would be separately added.

Changes are:
- Open file can open either file or project
- A project contains currently `projectVersion`, and `filesPath` as a list to support future files without changing format.
- A setting option to `Auto save` projects or not.
- On window close, user is asked to save unsaved project, before proceeding
- `NLS.str` has now `varargs`
- Test was added to check if an i18n property is not used.

### DIFF
--- a/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
+++ b/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
@@ -231,10 +231,6 @@ public class JadxCLIArgs {
 		return deobfuscationUseSourceNameAsAlias;
 	}
 
-	public boolean escapeUnicode() {
-		return escapeUnicode;
-	}
-
 	public boolean isEscapeUnicode() {
 		return escapeUnicode;
 	}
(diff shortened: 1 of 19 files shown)


asashour:
Regarding the build error, seems that `editorconfig` eclipse plugin doesn't work. And even by changing the EOL locally, `git` and `egit` do not recognize it.

Appreciate committing the PR, and modify the EOL. We can investigate it later, and probably add the `lf` EOL to `.gitattributes` later.

skylot:
@asashour thanks, looks good. 
Can you remove trailing spaces in properties files to pass build?

PR #547, opened 2019-04-01
fix: variable usage with enhanced for loop (#535)

### LINKED ISSUE #535: Variable is declared inside synchronized statement, but used outside
Hi, variables are declared inside block on ``` synchronized ``` statement, but used outside what avoids all references to them, example in the class ``` com/paypal/android/foundation/auth/operations/AuthenticationChallengeManager.java ```
```java
    private void notifySwitchUserListeners() {
        synchronized (s_instance) {
            java.util.HashSet<com.paypal.android.foundation.paypalcore.OnSwitchUserListener> hashSet = new java.util.HashSet(switchUserListenerQueue);
        }
        for (com.paypal.android.foundation.paypalcore.OnSwitchUserListener onSwitchUser : hashSet) {
            onSwitchUser.onSwitchUser();
        }
    }
```
where correct code would be
```java
    private void notifySwitchUserListeners() {
        java.util.HashSet<com.paypal.android.foundation.paypalcore.OnSwitchUserListener> hashSet;
        synchronized (s_instance) {
            hashSet = new java.util.HashSet(switchUserListenerQueue);
        }
        for (com.paypal.android.foundation.paypalcore.OnSwitchUserListener onSwitchUser : hashSet) {
            onSwitchUser.onSwitchUser();
        }
    }
```

APK: https://drive.google.com/file/d/18ylxtApJeWdY6Xq9Kcj5vkiX_-mKIwPG/view?usp=sharing

asashour:
Fixes #535 

The root cause is: with `enhance for loop`, we have `InvokeInsn` with `DONT_GENERATE`, however its argument should be considered in usage.

I am not sure about the correctness of the implementation regarding the `InsnType.ARITH`, it is needed because of the `TestArrayForEach2`


skylot:
@asashour thank you for finding the root of the issue.
Of cause `insn.getType() != InsnType.ARITH` check must be removed because it can break everything else.
And to fix test we need to add specific processing for other loop type `for-each` (class `jadx.core.dex.regions.loops.ForEachLoop`). You can add another `loopType instanceof` case to force process args of ForEachLoop (`varArg` and `iterableArg`). Or (maybe it is a better solution) add one more flag like `COUNT_IN_VAR_USAGE` add it in `LoopRegionVisitor` for instructions and args, and force process them in `CollectUsageRegionVisitor`.

asashour:
How about `ITERATOR_IN_FOR`? As it is not related to `vararg`

skylot:
@asashour  sorry, I am not good at explaining things, for me, it is easier to write code :)
So I made commit d50d39e based on your tests with a new flag. This flag is not loop specific and allows to remove additional region processing.
I am glad that your approach also works, but it bothers me that you process instructions even if `DONT_GENERATE` flag is set. But it doesn't break tests, so maybe it doesn't matter.

And now I don't know which commit is better ))

asashour:
With no extensive test cases, we can't really know if a corner-case is broken or not. However, it is a step forward, and it is not the final commit.

Your commit seems better, as it simplifies `CollectUsageRegionVisitor`, but adds one flag.

Please go ahead, there are many other issues to tackle :)

skylot:
> Please go ahead, there are many other issues to tackle :)

I will, but it is hard to review commit if I don't understand how it works :)
And sometimes I need to debug it, check things around, and write code to find other approaches )))

PR #548, opened 2019-04-01
chore: use functional interface instead of ISettingsUpdater

asashour:
And make `JadxSettings.partialSync` private, as I guess it is meant to be internally used.


skylot:
Hm, actually I make `partialSync` method and interface public intentionally, but turns out it not needed.
Also I don't understand why consumer is better then separate interface, you don't like additional interfaces?
I can merge this PR, but for me it is the same code, so I am just curious why you make such changes?

asashour:
Well, it is the same code, but lesser class is always better I guess, unless for backward compatibility, or more readability.

PR #550, opened 2019-04-01
fix: field increment

asashour:
I am not very comfortable with the code, because it can get easily messy, but it gives an idea of what should be done.

The fix is done by merging two `InsnNode`s to make one.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
@@ -68,6 +68,8 @@ public class SimplifyVisitor extends AbstractVisitor {
 
 	private static InsnNode simplifyOneArgConsecutive(InsnNode insn1, InsnNode insn2, ArithNode modInsn) {
 		if (insn1.getType() == InsnType.IGET
+				&& insn2.getType() == InsnType.IPUT
+				&& insn1.getResult().getSVar().getUseCount() == 2
 				&& insn2.getArg(1).equals(insn1.getResult())) {
 
 			FieldInfo field = (FieldInfo) ((IndexInsnNode) insn2).getIndex();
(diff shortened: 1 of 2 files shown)


jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -49,12 +49,34 @@ public void visit(MethodNode mth) {
 			for (int i = 0; i < list.size(); i++) {
 				InsnNode modInsn = simplifyInsn(mth, list.get(i));
 				if (modInsn != null) {
+					if (i != 0 && modInsn.contains(AFlag.ARITH_ONEARG)) {
+
+						InsnNode mergedNode = simplifyOneArgConsecutive(
+								list.get(i - 1), list.get(i), (ArithNode) modInsn);
+
+						if (mergedNode != null) {
+							list.remove(i - 1);
```
skylot: You can't just remove insn, because it results can be used somewhere else.
Check that it used only by your insn like this `list.get(i - 1).getResult().getSVar().getUseCount() == 2`

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
```diff
@@ -49,12 +49,34 @@ public void visit(MethodNode mth) {
...
 				}
 			}
 		}
 	}
 
+	private static InsnNode simplifyOneArgConsecutive(InsnNode insn1, InsnNode insn2, ArithNode modInsn) {
+		if (insn1.getType() == InsnType.IGET
+				&& insn2.getArg(1).equals(insn1.getResult())) {
+
+			FieldInfo field = (FieldInfo) ((IndexInsnNode) insn2).getIndex();
```
skylot: Check type of `insn2` before cast

jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
asashour: Done in method, although we do the check for `modInsn.contains(AFlag.ARITH_ONEARG)`, which won't happen unless we pass through https://github.com/skylot/jadx/blob/master/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java#L341

PR #554, opened 2019-04-02
fix: remove duplicate variable

asashour:
`VarUsage` has new `getUseNodes` and `getAssignNodes`, I guess this should be moved with the `UsePlace` to a combined new class.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/nodes/InsnNode.java
+++ b/jadx-core/src/main/java/jadx/core/dex/nodes/InsnNode.java
@@ -124,7 +124,7 @@ public class InsnNode extends LineAttrNode {
 		int count = getArgsCount();
 		for (int i = 0; i < count; i++) {
 			InsnArg arg = arguments.get(i);
-			if (arg == from) {
+			if (arg.equals(from)) {
 				InsnRemover.unbindArgUsage(null, arg);
 				setArg(i, to);
 				return true;
--- a/jadx-core/src/main/java/jadx/core/dex/nodes/InsnNode.java
+++ b/jadx-core/src/main/java/jadx/core/dex/nodes/InsnNode.java
@@ -124,7 +124,7 @@ public class InsnNode extends LineAttrNode {
 		int count = getArgsCount();
 		for (int i = 0; i < count; i++) {
 			InsnArg arg = arguments.get(i);
-			if (arg.equals(from)) {
+			if (arg == from) {
 				InsnRemover.unbindArgUsage(null, arg);
 				setArg(i, to);
 				return true;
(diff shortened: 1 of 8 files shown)


skylot:
Do you intentially make this "remove duplicate variable" only for method arguments?
(by this check: `varUsage1.getAssignNodes().isEmpty()`)
If yes, it can be done in much shorter way (insert before `checkCodeVars` method [here](https://github.com/skylot/jadx/blob/bae7f1b09cef15e22a7f472c9ca9a06a8705574c/jadx-core/src/main/java/jadx/core/dex/visitors/regions/variables/ProcessVariables.java#L47)):
```java
		for (RegisterArg arg : mth.getArguments(false)) {
			SSAVar ssaVar = arg.getSVar();
			List<RegisterArg> useList = ssaVar.getUseList();
			if (useList.size() == 1) {
				RegisterArg use = useList.get(0);
				InsnNode insn = use.getParentInsn();
				if (insn != null
					    && insn.getResult() != null
					    && insn.getResult().getType().equals(arg.getType())) {
					// merge code vars
					SSAVar replaceSsaVar = insn.getResult().getSVar();
					codeVars.remove(replaceSsaVar.getCodeVar());
					replaceSsaVar.setCodeVar(ssaVar.getCodeVar());
				}
			}
		}
```
Note that I made `CodeVar` class especially for this case: merge several SSAVars into one named variable.

P.S. I already tried to make this in general case: merging variables by name if debug info attached. But this needed a lot of checks for java scopes ([this](https://github.com/skylot/jadx/blob/bae7f1b09cef15e22a7f472c9ca9a06a8705574c/jadx-core/src/main/java/jadx/core/dex/visitors/regions/variables/ProcessVariables.java#L48)  todo comment)

asashour:
Oh, that is getting behind my comprehension. I guess I need to study this area first.

After the suggested changes, there is a failure of compiling `TestTryCatch.test6`, there is `PhiInsn`.

As a side note, I guess to attract more people to be involved, we need some description/javadoc/wiki of major areas.

Failed test: from:

```
	public static boolean test6(Object obj) {
		boolean res = false;
		while (true) {
			try {
				res = exc2(obj);
				return res;
			} catch (IOException e) {
				res = true;
			} catch (Throwable e) {
				if (obj == null) {
					obj = new Object();
				}
			}
		}
	}
```

To:

```
    public static boolean test6(Object obj) {
        obj = obj;
        while (true) {
            try {
                return exc2(r0);
            } catch (IOException e) {
            } catch (Throwable th) {
                if (r0 == null) {
                    r0 = new Object();
                }
            }
        }
    }

```

skylot:
@asashour sorry about failed tests, this can be fixed by adding `&& !replaceSsaVar.isUsedInPhi()` like:
```java
// merge code vars
SSAVar replaceSsaVar = insn.getResult().getSVar();
if (replaceSsaVar != null && !replaceSsaVar.isUsedInPhi()) {
	codeVars.remove(replaceSsaVar.getCodeVar());
	replaceSsaVar.setCodeVar(ssaVar.getCodeVar());
}
```

skylot:
Also I don't think this is a good change anyway, because it applied to a very rare case (one time use of a method argument) and replace any insn result regardles type, this case:
```java
		public int test3(int a) {
			int b = mth(a);
			use(b);
			return b;
		}

		private static int mth(int i) {
			return 7;
		}

		private static void use(int i) {}
```
decompiled as:
```java
    public int test3(int a) {
        a = mth(a);
        use(a);
        return a;
    }
```
And this looks strange.
I want to make more general fix. Like for this case:
```java
public int test4() {
	int b = mth(0);
	use(b);
	b++;
	use(b);
	return b;
}
```

PR #558, opened 2019-04-03
fix: conditions in ternary if (#449)

### LINKED ISSUE #449: type-inference-wip: invalid not-equal operator is generated
Hi, file ``` com/paypal/android/sdk/contactless/reader/emv/EmvCardTrack2Data.java ```
String field is declared
```java
    private java.lang.String f15781a;
```
```java
    public boolean equals(java.lang.Object obj) {
        //...
        if (!this.f15781a == null ? this.f15781a.equals(emvCardTrack2Data.f15781a) : emvCardTrack2Data.f15781a == null) {
            return false;
        }
```
so this expression ``` (!this.f15781a == null ? this.f15781a.equals(emvCardTrack2Data.f15781a) : emvCardTrack2Data.f15781a == null) ``` seems invalid, it should be wrapped with ``` (...) ```

APK: https://drive.google.com/file/d/16O--dtXt35zB1Ahx-9qEYvCmyeF1n0AG/view?usp=sharing

asashour:
Fix #449

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/regions/conditions/IfCondition.java
+++ b/jadx-core/src/main/java/jadx/core/dex/regions/conditions/IfCondition.java
@@ -6,6 +6,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 
 import jadx.core.dex.instructions.ArithNode;
 import jadx.core.dex.instructions.ArithOp;
@@ -118,7 +119,7 @@ public final class IfCondition {
 			case COMPARE:
 				return new IfCondition(cond.getCompare().invert());
 			case TERNARY:
-				return ternary(not(cond.first()), cond.third(), cond.second());
+				return ternary(cond.first(), not(cond.second()), not(cond.third()));
 			case NOT:
 				return cond.first();
 			case AND:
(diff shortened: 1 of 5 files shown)


asashour:
Just a hint, to fix this particular case, we need to change this [line](https://github.com/skylot/jadx/blob/master/jadx-core/src/main/java/jadx/core/dex/visitors/regions/IfMakerHelper.java#L218)

from:

    nextElse = IfInfo.invert(nextElse);

to:

    nextThen = IfInfo.invert(nextThen);

So far, I have no idea when to use which statement.

skylot:
@asashour I don't think this is a correct way to fix this. The root of this issue somewhere inside invert or simplify, like additional/missing wrapping. Can you check this?

asashour:
I guess that:

    return ternary(cond.first(), not(cond.second()), not(cond.third()));

could also be:

    return ternary(not(cond.first()), not(cond.third()), not(cond.second()));

skylot:
@asashour I checked your test and on master branch it doesn't reproduce original issue. I mean I expect decompiled code contains a string like `!this.a == null ` and not compilable.

asashour:
Well, yes, the original case is not reproducible `exactly` by the test case, but on fixing the test case, the original APK is decompiled correctly.

asashour:
I tried with different ways, but couldn't find an exact match.  We need bytecode changes (or smali), since it seems `javac` and eclipse compilers don't give the same smali output.

skylot:
I understand, java test is not possible sometimes, but you can always make smali test.
Because without test you did not make a fix of the issue, you just find a way to avoid an issue.

asashour:
Ummm, I am not very convinced, since there is a test, that passes when the original case passes, and fails when it is not.

Anyhow, adding the original smali.

asashour:
Please note, `IfCondition` `equals` and `hashCode()` were added because of `TestIfCondition.testSimplify()` and `testSimplify()2`, but I am not sure about the test case purpose.

PR #560, opened 2019-04-04
fix: move return to inside TryCatchRegion, if possible

asashour:


skylot:
Sorry, but to resolve this issue we need to fix the root cause and in this case, it is in [`CodeShrinkVisitor`](https://github.com/skylot/jadx/blob/cf79a519d3f7fbd6d260eeaead449b473ce76698/jadx-core/src/main/java/jadx/core/dex/visitors/shrink/CodeShrinkVisitor.java#L25).
The purpose of a `CodeShrinkVisitor` is to inline variables with one-time usage, but for doing this there are a lot of checks like: we can't reorder insns with side-effects (invoke for example). And in your test, there is such invoke instruction from finally block which incorrectly prevents inlining because it moved and marked as `DONT_GENERATE` and `CodeShrinkVisitor` don't handle this case well (inline canceled [here](https://github.com/skylot/jadx/blob/cf79a519d3f7fbd6d260eeaead449b473ce76698/jadx-core/src/main/java/jadx/core/dex/visitors/shrink/CodeShrinkVisitor.java#L149))
CFG of the test method (problem insn in block 3):
cfg-560

To fix that we can change [`canReorder`](https://github.com/skylot/jadx/blob/17d8516d3b5561da669a1a87549c62ac3f171cab/jadx-core/src/main/java/jadx/core/dex/nodes/InsnNode.java#L197) method in `InsnNode` like this:
```java
	public boolean canReorder() {
		if (contains(AFlag.DONT_GENERATE)) {
			if (getType() == InsnType.MONITOR_EXIT) {
				return false;
			}
			return true;
		}
		switch (getType()) {
...
```

asashour:
Very nice, it is a complex logic, but I learn gradually, thanks a lot for your help.

Issue #390, opened 2018-11-10
Jadx-gui 0.8.0-b569-20b03aa fails on framework.jar with java classes inside


S-trace:
In https://github.com/skylot/jadx/issues/388#issuecomment-437601188 @skylot says that jadx can process java .jar with .class files inside.
I tried to do this on out/target/common/obj/JAVA_LIBRARIES/framework_intermediates/classes.jar (from LineageOS 14.1 after ROM build), and jadx-gui failed on this file:
```
INFO  - output directory: classes
INFO  - loading ...
INFO  - converting to dex: classes.jar ...
WARN  - dx message: 
trouble writing output:
com.android.dex.DexIndexOverflowException: Too many method references to fit in one dex file: 102035; max is 65536.
You may try using multi-dex. If multi-dex is enabled then the list of classes for the main dex list is too large.
References by package:
    27 android
   304 android.accessibilityservice
   685 android.accounts
   774 android.animation
    28 android.annotation
  6966 android.app
  1034 android.app.admin
   153 android.app.assist
   504 android.app.backup
   190 android.app.job
    78 android.app.trust
   185 android.app.usage
   168 android.appwidget
  2280 android.bluetooth
   232 android.bluetooth.le
  2016 android.content
  2286 android.content.pm
    55 android.content.pm.permission
   792 android.content.res
   535 android.database
   499 android.database.sqlite
   116 android.ddm
   169 android.drm
    23 android.filterfw
   817 android.filterfw.core
    26 android.filterfw.format
    42 android.filterfw.geometry
    46 android.filterfw.io
    83 android.filterpacks.base
   410 android.filterpacks.imageproc
     6 android.filterpacks.numeric
    17 android.filterpacks.performance
    18 android.filterpacks.text
    29 android.filterpacks.ui
    18 android.filterpacks.videoproc
    21 android.filterpacks.videosink
    86 android.filterpacks.videosrc
   223 android.gesture
  1515 android.graphics
  1898 android.graphics.drawable
    41 android.graphics.drawable.shapes
   108 android.graphics.pdf
   709 android.hardware
   358 android.hardware.camera2
    19 android.hardware.camera2.dispatch
   427 android.hardware.camera2.impl
   483 android.hardware.camera2.legacy
    25 android.hardware.camera2.marshal
   208 android.hardware.camera2.marshal.impl
   192 android.hardware.camera2.params
   122 android.hardware.camera2.utils
   252 android.hardware.display
   270 android.hardware.fingerprint
[...]

PR #561, opened 2019-04-04
chore: write to standard input and output if there is an error in dex

asashour:
And rename parameter

### DIFF
--- a/jadx-core/src/main/java/jadx/core/utils/files/JavaToDex.java
+++ b/jadx-core/src/main/java/jadx/core/utils/files/JavaToDex.java
@@ -31,14 +31,16 @@ public class JavaToDex {
 
 	private String dxErrors;
 
-	public byte[] convert(String javaFile) throws JadxException {
+	public byte[] convert(String jarFile) throws JadxException {
 		try (ByteArrayOutputStream out = new ByteArrayOutputStream();
 		     ByteArrayOutputStream errOut = new ByteArrayOutputStream()) {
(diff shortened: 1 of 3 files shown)


jadx-core/src/main/java/jadx/core/utils/files/JavaToDex.java
```diff
@@ -31,14 +31,16 @@ public DxArgs(DxContext context, String dexFile, String[] input) {
...
 		try (ByteArrayOutputStream out = new ByteArrayOutputStream();
 		     ByteArrayOutputStream errOut = new ByteArrayOutputStream()) {
 			DxContext context = new DxContext(out, errOut);
-			DxArgs args = new DxArgs(context, "-", new String[]{javaFile});
+			DxArgs args = new DxArgs(context, "-", new String[]{jarFile});
 			int result = (new Main(context)).runDx(args);
 			dxErrors = errOut.toString(CHARSET_NAME);
 			if (result != 0) {
+				System.out.println(out);
+				System.err.println(errOut);
```
skylot: I don't think this is a good solution. Yes, this is help in development, but users will don't see this anyway.
I already tried to process errors (you can see `dxErrors` field) and output them in exception and logs. Sometimes it works like in #390 (`dx message:`).
Also `out` can contains binary data.

asashour: ok, thanks.

Issue #552, opened 2019-04-02
ClspGraph to include generic details


asashour:
As communicated in #512, what would be the way to add the generic details to `ClspGraph`?

Appreciate providing sample/pseudo code of the expected entry of one of the methods/classes.

I guess we should start with Collections.

skylot:
At first, I will describe the current `Clsp` concept and implementation:
* Current Java and Android classpath is really big and contains many classes. Jadx needs some information from those classes for correct decompilation (like class hierarchy for type inference) and need to load it at every start, so loading and processing must be really fast. For this jadx store minimum needed information in a binary format close to resulting class structure for minimum conversion.
* Work with binary file (`.jcst`) implemented in `jadx.core.clsp.ClsSet` class, initial conversion from input classpath to binary format  done in `ConvertToClsSet`. Next classes from `ClsSet` added to maps in `ClspGraph`. 
* `ClsSet` stores array of all classes full names and their "parents" (combined array of super class and interfaces), this is enough to build class hierarchy and implement methods like `instanceof` in `ClspGraph`.
* At first I was planning bundle two classpaths (`.jcst` files) for Java and Android. But apparently Android classpath just adds an additional class and core classes are almost the same. So now only Android is bundled. 
* For build `.jcst` file you need to use `ConvertToClsSet` cli tool and as input provide something like `android.jar` from AndroidSDK (this is "stub" jar and all methods are empty).

Next about new features scope:
* generic map for classes and methods
* `throws` list for methods 
* `vararg` info (variable arguments count)
* maybe method access modifiers

To extend the current implementation we need:
* add `ClspGenericMap` class to store a mapping of generics mapping (generic name to `extends` types)
* add `ClspMethod` with fields same as in `MethodInfo` plus reference to `ClspGenericMap` and info about `throws` list and `vararg` (just boolean).

Binary serialization: I prefer manual serialization, it is harder to write but very flexible and efficient. Please note that all saved types just a reference by class id, because we have a full classpath, so there are no external classes.

Also for work with generics, we need some helper class like (`ClspGraph`) which will collect and simplify access to generic information (some sort of api methods). I think we need methods for map generic types to actual types in implementation class. Like for `class A<T>` and `class B extends A<String>` make mapping `T` -> `String` in `A`. These methods will be used in type inference.

asashour:
How was the current `core.jcst` generated?

When I run the following (with latest `Android Studio`)

```
RootNode root = new RootNode(new JadxArgs());
[...]

PR #564, opened 2019-04-08
feat: add generic method information to .jcst

asashour:
This change writes the following method information to `.jcst`, which is the first part of #552.

- Method if any of its arguments is `GenericObject` or `GenericType`, and only those arguments are saved.
- Method if it has `varargs`.

Other options:
- Currently method `shortId` is saved, however we can save its `hashCode` instead (we can also verify if two methods have same `hashCodes` before writing).
- Writing method `throws`, will make many more methods are saved, not sure if this is needed.


skylot:
Great work!
One issue: in new `android-5.1.jar` you removed some files used in resource decoding (`android` folder inside jar). I don't remember if they really needed because resource decoding is a little messy right now and don't have any tests. And btw which android version you use to build new `jcst` file? I think we need the latest version, but in this case, we need to rename jar and update these resource files.

About using hashCode as method id: it is possible, but we can't use java built-in string hashCode, because it can change. And we need to support collision case because we can't "rename" methods. Also in new Android can be added a new method which can collide with old one and jadx will find old instead of missing a new one, but this is very unlikely :).
So I think string `shortId` is easier to implement and use.

`throws` info needed to fix cases of not needed try/catch wrapping like in test [`TestTryAfterDeclaration`](https://github.com/skylot/jadx/blob/1e5541175e33d01d3b0a6a41b28bbd54ca10bced/jadx-core/src/test/java/jadx/tests/integration/trycatch/TestTryAfterDeclaration.java#L16) (PR #497). In this test generated two `try` blocks and this is exactly what written in bytecode. One of the blocks wraps `System.out.println` which not throws any exception and this lead to compilation error. So if jadx knows `throws` it can eliminate such incorrect `try` wrapping.

asashour:
> One issue: in new `android-5.1.jar` you removed some files used in resource decoding (`android` folder inside jar).

Actually, it wasn't deliberate. I just used the tool `ConvertToClsSet` to point to `clsp-data\android-5.1.jar`, it seems to delete it, I will change this behavior, not to remove other files.

> And which android version you use to build new `jcst` file? I think we need the latest version, but in this case, we need to rename jar and update these resource files.

It is the latest, SDK tools says the below, but I don't know which is the actual android version.

    tools   | 26.1.1  | Android SDK Tools 26.1.1 | tools\


> For `throws`, I know the test method, but I would leave this for further PR/discussion, since I think there is a way to tackle it, without looking into the throw possibilities.

PR #566, opened 2019-04-08
fix: remove redundant array type when initialized with declaration

asashour:


skylot:
Nice improvement, thanks!
Also, this can be done for fields initialization (NYI test candidate)

PR #567, opened 2019-04-08
fix: remove field redundant array type when initialized

asashour:


skylot:
@asashour it looks a little hacky, but I think it is ok :)

PR #575, opened 2019-04-09
test: add two cases for switch-try-break combination

asashour:
The first one gives error:

    error: variable res might not have been initialized

The second one gives:

    JadxRuntimeException: Can't find immediate dominator


skylot:
@asashour thanks!! Very nice cases :+1:

PR #582, opened 2019-04-10
feat: add options to configure "renaming" (#570)

### LINKED ISSUE #570: Add options to disable "renaming"
This issue discussion started in #158.
Sometimes it is useful to view names "as-is" without any modifications.
So jadx must have an option for that.

For now, jadx has many "renaming" types besides deobfuscation:
* to support system case sensitivity
* to make valid java identifiers (to make code compilable)
* to remove not "printable" characters 
"printable" - here means only Latin chars (32 to 126 codes in [ascii table](http://www.asciitable.com/))

Options must allow to disable/enable such types separately:
* In `JadxArgs` it must be enum set (full set is default)
* In jadx-cli list of lower case enum names with comma separator, plus shortcuts like `all` and `none`
Examples: `--rename-flags case,valid,printable` or `--rename-flags all`
* In jadx-gui in must be several checkboxes under common label.
And if some options is disabled before source export warning notification about changing names can be shown.

asashour:
Fix #570

### DIFF
--- a/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
+++ b/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
@@ -320,7 +320,7 @@ public class JadxCLIArgs {
 				set.add(RENAME.CASE);
 				set.add(RENAME.VALID);
 				set.add(RENAME.PRINTABLE);
-			} else if (!value.equals("NONE")) {
+			} else if (!value.equalsIgnoreCase("NONE")) {
 				for (String s : value.split(",")) {
 					try {
 						set.add(RENAME.valueOf(s.toUpperCase(Locale.ROOT)));
@@ -333,7 +333,6 @@ public class JadxCLIArgs {
 								+ ", possible values are " + values.toLowerCase(Locale.ROOT));
 					}
 				}
-
 			}
 			return set;
 		}
--- a/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
+++ b/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
@@ -1,7 +1,7 @@
 package jadx.cli;
 
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
@@ -17,6 +17,7 @@ import com.beust.jcommander.Parameter;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.Appender;
 import jadx.api.JadxArgs;
+import jadx.api.JadxArgs.RENAME;
 import jadx.api.JadxDecompiler;
 import jadx.core.utils.exceptions.JadxException;
 import jadx.core.utils.files.FileUtils;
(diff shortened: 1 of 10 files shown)


jadx-gui/src/main/java/jadx/gui/settings/JadxSettingsWindow.java
```diff
@@ -165,6 +193,49 @@ private SettingsGroup makeDeobfuscationGroup() {
...
+		JLabel renameCaseSensitiveLabel
+			= group.addRow(NLS.str("preferences.rename_case"), renameCaseSensitive);
+		group.addRow(NLS.str("preferences.rename_valid"), renameValid);
+		group.addRow(NLS.str("preferences.rename_printable"), renamePrintable);
+
+		File dir = FileUtils.createTempDir("jadx");
+		if (FileUtils.isCaseSensitiveFS(dir)) {
+			renameCaseSensitive.setSelected(false);
+			renameCaseSensitive.setEnabled(false);
+			renameCaseSensitiveLabel.setEnabled(false);
```
skylot: Don't do this, leave selection choice to the user. 
Also, this check will not work for masOS, because HFS case sensitivity determined by user and can be different for different drives.

jadx-core/src/main/java/jadx/api/JadxArgs.java
```diff
@@ -45,7 +45,11 @@
 	private boolean respectBytecodeAccModifiers = false;
 	private boolean exportAsGradleProject = false;
 
-	private boolean isFsCaseSensitive;
+	private boolean fsCaseSensitive;
+
+	private boolean renameCaseSensitive = true;
+	private boolean renameValid = true;
+	private boolean renamePrintable = true;
```
skylot: Please move `RENAME` enum to this class and make instead three fields one using Set.
Also for this set always use `EnumSet` class instead `HashSet`.

jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
```diff
@@ -254,4 +269,59 @@ public boolean isRespectBytecodeAccessModifiers() {
...
+		public Set<RENAME> convert(String value) {
+			value = value.toUpperCase(Locale.ROOT);
+			Set<RENAME> set = new HashSet<>();
+			if (value.equals("ALL")) {
+				set.add(RENAME.CASE);
+				set.add(RENAME.VALID);
+				set.add(RENAME.PRINTABLE);
+			} else if (!value.equals("NONE")) {
+				for (String s : value.split(",")) {
+					set.add(RENAME.valueOf(s));
```
skylot: Throw a detailed exception like `<value> is unknown, possible values list: <list of values>`, because user can input incorrect names.
Also, add some tests to check this converter.

asashour:
I guess we should change `NameMapper.isValidIdentifier()` to either have `JadxArgs` or `boolean isPrintable`, but this would be changed for all caller.

In other words, `isValidIdentifier` already includes `isPrintable` in its checking.

skylot:
I think we can remove "printable" checks from `NameMapper.isValidIdentifier()` because allowed chars must be checked by `VALID_JAVA_IDENTIFIER` regexp. And use both checks in new method with rename set args.

Issue #385, opened 2018-11-07
Support Dex 39


skylot:
New instructions:
- const-method-handle
- const-method-type

[Nice description](https://www.pnfsoftware.com/blog/dex-version-39-new-dalvik-opcodes/)

Issue #384, opened 2018-11-07
Support Dex 38


skylot:
New instructions:
- invoke-polymorphic
- invoke-custom

[Good description](https://www.pnfsoftware.com/blog/android-o-and-dex-version-38-new-dalvik-opcodes-to-support-dynamic-invocation/)

skylot:
Done.

Issue #224, opened 2018-02-03
Anti-disassemble tricks with illegal opcodes totally break JADX decompilation


enovella:
This protector destroys all the JADX decompilation. It may be interesting to take a peek at how this protector confuses DEX decompilers by adding junk opcodes.

**APK:** https://koodous.com/apks/a5722d6ab8a2de93f94901129f9793a4b898a31fe33250e2466d4d877af1dcd3

**More info:** 
https://github.com/rednaga/APKiD/issues/52
https://github.com/JesusFreke/smali/issues/562#issuecomment-340075562

**APKiD output**:
```
$ apkid /tmp/GODDESS_KISS_v1.040.12_apkpure.com.apk 
[+] APKiD 1.0.0 :: from RedNaga :: rednaga.io
[*] /tmp/GODDESS_KISS_v1.040.12_apkpure.com.apk!classes.dex
 |-> anti_vm : Build.FINGERPRINT check, Build.MANUFACTURER check, Build.MODEL check, Build.PRODUCT check, Build.TAGS check, network operator name check, subscriber ID check
 |-> compiler : dexlib 2.x
 |-> protector : CNProtect (anti-disassemble)
```

enovella:
Any workaround on this topic?

manizzle:
https://github.com/JesusFreke/smali/issues/562#issuecomment-340124461 looks like a way smali deals with it?

zhangmengfeigithub:
mistake

skylot:
I made incorrect instructions removing in branch `issue-224`. In most cases this is enough, for others I need to implement proper instructions checking and dead code elimination. This is not hard but need to change a lot of code.
Also I don't see any correct and usable code in these classes (correct me if I wrong), so I think these classes can be just ignored.
Another problem: if I enable this 'remove bad instructions' behavior by default it will hide jadx internal errors and you will see incorrect code instead error :(

P.S. Did anyone trying to use [simplify](https://github.com/CalebFenton/simplify) or [dex-oracle](https://github.com/CalebFenton/dex-oracle) on this apk?

enovella:
Hi all,

**@manizzle:**

> Basically, uses the concept of dead code which is bad (from Dex Education presentation) to insert bad opcodes/junk.
> Baksmali handles this correctly by ignoring the class.

You suggest to skip the class decompilation too?

**@skylot:**
Are you saying that these classes are intentionally included as a bait and they are pointless? If so, can I see the decompilation of any of them?
[...]

PR #583, opened 2019-04-10
fix: handle unknown opcodes (#342)

### LINKED ISSUE #342: decompile static block throw exception
static {
        /* JADX: method processing error */
/*
Error: jadx.core.utils.exceptions.DecodeException: Load method exception: bogus opcode: 0073 in method: com.android.mms.attachment.datamodel.FileProvider.<clinit>():void, dex: 
	at jadx.core.dex.nodes.MethodNode.load(MethodNode.java:117)
	at jadx.core.dex.nodes.ClassNode.load(ClassNode.java:249)
	at jadx.core.ProcessClass.process(ProcessClass.java:29)
	at jadx.api.JadxDecompiler.processClass(JadxDecompiler.java:286)
	at jadx.api.JavaClass.decompile(JavaClass.java:62)
	at jadx.api.JavaClass.getCode(JavaClass.java:48)
Caused by: java.lang.IllegalArgumentException: bogus opcode: 0073
	at com.android.dx.io.OpcodeInfo.get(OpcodeInfo.java:1217)
	at com.android.dx.io.OpcodeInfo.getName(OpcodeInfo.java:1224)
	at jadx.core.dex.instructions.InsnDecoder.decode(InsnDecoder.java:590)
	at jadx.core.dex.instructions.InsnDecoder.process(InsnDecoder.java:74)
	at jadx.core.dex.nodes.MethodNode.load(MethodNode.java:103)
	... 5 more
*/
        /*
        // Can't load method instructions.
        */
        throw new UnsupportedOperationException("Method not decompiled: com.android.mms.attachment.datamodel.FileProvider.<clinit>():void");
    }
[test.zip](https://github.com/skylot/jadx/files/2304296/test.zip)

asashour:
Fix #342 

If an opcode is incorrect, then we shouldn't throw Exception, because this will make the method has no blocks.

Instead, print warning, and replace it with `NOP`, this is also what apktool does.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java
+++ b/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java
@@ -579,7 +579,14 @@ public class InsnDecoder {
 						InsnArg.reg(insn, 0, ArgType.UNKNOWN_OBJECT));
 
 			default:
-				throw new DecodeException("Unknown instruction: '" + OpcodeInfo.getName(insn.getOpcode()) + '\'');
+				String opcode;
+				try {
+					opcode = OpcodeInfo.getName(insn.getOpcode());
+				} catch (IllegalArgumentException e) {
+					opcode = "0x" + Integer.toHexString(insn.getOpcode());
+				}
+				LOG.warn("Unknown instruction: '" + opcode + "', replaced with NOP");
+				return new InsnNode(InsnType.NOP, 0);
 		}
 	}
 
--- a/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java
+++ b/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java
@@ -579,14 +579,13 @@ public class InsnDecoder {
 						InsnArg.reg(insn, 0, ArgType.UNKNOWN_OBJECT));
 
 			default:
-				String opcode;
 				try {
-					opcode = OpcodeInfo.getName(insn.getOpcode());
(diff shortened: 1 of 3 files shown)


skylot:
@asashour can you describe some case which you trying to fix?
Because code you changed will be executed only for instructions not yet supported by jadx (#384 #385). And removing such insns will produce incorrect code, which is bad. 
Also if you want to show warning please use `mth.addWarn("Unsupported insn");` this will add a comment into generated code and mark method as "inconsistent"

P.S. explicitly throw an exception in try/catch block is a terrible idea

asashour:
There are two cases here:

1. Correct Opcodes, recognized by `OpcodeInfo.getName()`, but not handled in Jadx, previously `DecodeException` was thrown, and they remain also thrown.
2. Incorrect Opcodes, not recognized by `OpcodeInfo.getName()`, and should be simply ignored.

I guess `OpcodeInfo` is the reference, and if we have new opcodes, then we should upgrade the dependency.

------
If `OpcodeInfo` is not the reference, then in this case how to handle `0x73`, which is marked 'unused' in the [doc](https://source.android.com/devices/tech/dalvik/dalvik-bytecode)

skylot:
Incorrect Opcodes will throw exception from parser [here](https://github.com/skylot/jadx/blob/e1f4955286f73a6ca4525fc27b11fd02106f7f08/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java#L57).
For unsupported add method warning as I suggest above.

skylot:
As a example of bad opcodes you can use apk from this issue #224

bobolau:
when to support  instruction: 'invoke-custom'?

asashour:
@bobolau : please subscribe to #384

PR #590, opened 2019-04-12
fix: don't use anonymous type as type

asashour:
Is there a better way to get `RootNode`?


skylot:
Another way is to pass `mth` as argument and use `mth.root()` but I am not sure it is better.

asashour:
I guess `root` this way, is for better performance, with minimal memory effect (since we have a single `TypeInferenceVisitor`.

I guess the `!type.isGeneric()` is incorrect, since we need to check the classpath, but that's another investigation (for now).

asashour:
Should we remove assignments with 0 usages instead? It would be better I guess.

asashour:
I guess `TestConstructorInvoke` lost its need now.

skylot:
I agree, `TestConstructorInvoke` can be removed.
And remove unused variable also was a good decision, thanks!

PR #602, opened 2019-04-15
fix: cast of int-to-(number) when int is boolean (#596)

### LINKED ISSUE #596: Invalid casts among primitive types
Hi, in some cases casts are invalid. For example in class ``` com/paypal/android/foundation/presentation/model/ConsentUriParams.java ``` type ``` boolean ``` is converted to ``` byte ``` using explicit cast
```java
    public void writeToParcel(android.os.Parcel parcel, int i) {
        parcel.writeByte((byte) this.showConsent);
        parcel.writeByte((byte) this.showOneTouch);
```
but the following code should be generated instead
```java
    public void writeToParcel(android.os.Parcel parcel, int i) {
        parcel.writeByte(this.showConsent ? 1 : 0);
        parcel.writeByte(this.showOneTouch ? 1 : 0);
```
APK: https://drive.google.com/file/d/16O--dtXt35zB1Ahx-9qEYvCmyeF1n0AG/view?usp=sharing

asashour:
Fixes #596 

Please merge #601 first, since it is already included.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/instructions/args/InsnArg.java
+++ b/jadx-core/src/main/java/jadx/core/dex/instructions/args/InsnArg.java
@@ -10,8 +10,12 @@ import org.slf4j.LoggerFactory;
 
 import jadx.core.dex.attributes.AFlag;
 import jadx.core.dex.instructions.ArithNode;
+import jadx.core.dex.instructions.IfNode;
+import jadx.core.dex.instructions.IfOp;
 import jadx.core.dex.instructions.InsnType;
+import jadx.core.dex.instructions.mods.TernaryInsn;
 import jadx.core.dex.nodes.InsnNode;
+import jadx.core.dex.regions.conditions.IfCondition;
 import jadx.core.utils.InsnUtils;
 
 /**
@@ -110,8 +114,14 @@ public abstract class InsnArg extends Typed {
 		if (i == -1) {
 			return null;
 		}
+
+		InsnArg arg = processCast(insn, i);
+		if (arg != null) {
+			return arg;
+		}
+
 		insn.add(AFlag.WRAPPED);
-		InsnArg arg = wrapArg(insn);
+		arg = wrapArg(insn);
 		parent.setArg(i, arg);
 
 		if (insn.getType() == InsnType.ARITH && parent.getType() == InsnType.ARITH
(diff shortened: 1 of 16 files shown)


jadx-core/src/main/java/jadx/core/dex/instructions/args/InsnArg.java
```diff
@@ -122,6 +132,32 @@ public InsnArg wrapInstruction(InsnNode insn) {
 		return arg;
 	}
 
+	private InsnArg processCast(InsnNode insn, int i) {
```
skylot: Move this code to `ModVisitor`

jadx-core/src/main/java/jadx/core/dex/visitors/ModVisitor.java
```diff
@@ -155,6 +159,22 @@ private static void replaceStep(MethodNode mth, InsnRemover remover) {
...
+								InsnArg zero = new LiteralArg(0, ArgType.BYTE);
+								InsnArg one = new LiteralArg(1, ArgType.BYTE);
+								IfNode ifNode = new IfNode(IfOp.NE, -1, InsnArg.wrapArg(pi), LiteralArg.FALSE);
+								IfCondition condition = IfCondition.fromIfNode(ifNode);
+								TernaryInsn ternary = new TernaryInsn(condition, insn.getResult(), one, zero);
+								replaceInsn(block, i, ternary);
+								remover.addAndUnbind(pi);
+							}
+						}
+						break;
```
skylot: I don't know why you need previous instruction, try to limit modification scope to minimum.
So I can suggest do this:
```java
					case CAST:
						// replace boolean to byte cast with ternary
						if (insn.getResult().getType() == ArgType.BYTE
								&& insn.getArg(0).getType() == ArgType.BOOLEAN) {
							IfNode ifNode = new IfNode(IfOp.EQ, -1, insn.getArg(0), LiteralArg.TRUE);
							IfCondition condition = IfCondition.fromIfNode(ifNode);
							InsnArg zero = InsnArg.lit(0, ArgType.BYTE);
							InsnArg one = InsnArg.lit(1, ArgType.BYTE);
							TernaryInsn ternary = new TernaryInsn(condition, insn.getResult(), one, zero);
							replaceInsn(block, i, ternary);
						}
						break;
```

sergey-wowwow:
@asashour @skylot 
I would also recommend to do the same for boolean -> int casts, because such cases are also generated by a wrong way

jadx-core/src/main/java/jadx/core/dex/visitors/ModVisitor.java
asashour: Sure, this is much better, I guess I was distracted by the "previous Instruction" in the previous code location.

asashour:
> I would also recommend to do the same for boolean -> int casts, because such cases are also generated by a wrong way

I thought about `int-to-char`, `int-to-short`, but I guess there is no boolean to int cases [here](http://pallergabor.uw.hu/androidblog/dalvik_opcodes.html), I would try to make some test cases and fix them in other PRs.

I am not experienced in davlik, so appreciate if you can provide some tests.

jadx-core/src/main/java/jadx/core/dex/visitors/ModVisitor.java
asashour: Also, should we just use `.isPrimitive()`, instead of the `BYTE||CHAR||SHORT||LONG||DOUBLE||FLOAT`, or even remove it completely, since no casting can be done except to these things (`VOID||BOOLEAN` are the remaining in `primitive`).

skylot: You can use  `.isPrimitive()` I think it is ok.
But for me looks like you overdid this ))
Yeah sure, you can write these weird conversions in smali but I don't think we can find all of them in real APKs.
Anyway I think this is good work, thanks!

PR #605, opened 2019-04-17
feat(gui): save tree expansions in project

asashour:
- Tree expansions, and expanding them back when the project is open.

### DIFF
--- a/jadx-core/src/main/java/jadx/api/JavaClass.java
+++ b/jadx-core/src/main/java/jadx/api/JavaClass.java
@@ -68,7 +68,7 @@ public final class JavaClass implements JavaNode {
 		cls.unload();
 	}
 
-	ClassNode getClassNode() {
+	public ClassNode getClassNode() {
 		return cls;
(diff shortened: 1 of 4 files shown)


skylot:
@asashour it works, but I don't like implementation it is too "heavy". I mean on every tree change you traverse all leaves and creating new arrays of arrays and it can be a lot of classes and packages.
I think it is better to use incremental changes like add item on expanding and remove on collapse without rebuild full info (for this you can use temporal optimized storage and convert it to "save" format only on save).
Another approach is to save this tree info only on project save or file reload.

Also, it will be nice if this work on deobfuscation switch, so you need to store and search by original names.

And you need to decide with @jpstotz  who implements saving of splitter location :))

asashour:
I don't know how to get the original name of `JPackage`.

skylot:
Yeah, looks like adding original package name to `JPackage` will need a lot of changes. So I think this change is good start and we will fix it later.

PR #606, opened 2019-04-17
feat(gui): save the class tree width and restore it upon jadx-gui start

jpstotz:


jpstotz:
At a first glance #605 does more or less the same (saw #605 right after I created this PR) . The only difference is that this PR saves the tree width not project based. 
For cases where a new APK file is loaded and therefore no project file is available #605 is not effective, in such a case this PR would help. Therefore I think both PRs their justifications.

asashour:
This proves there is a need for this feature :)

I guess since the tree width is more related to the window size, I am now inclined to have it as a global setting, not a project-specific.

skylot:
I agree, tree size is better to store in global settings.
@jpstotz for save/change tree size please use methods in `JSplitPane` component instead changing size of it children. Also you need to use some kind of listener to save size on changes.

asashour:
Great, I will remove the width from the other PR then

asashour:
> Also you need to use some kind of listener to save size on changes.

I guess no need for a listener, since we also save the window size only in `closeWindow()`

skylot:
@jpstotz will you finish this PR according to suggestions about using `JSplitPane` to change split size?

jpstotz:
@skylot Sorry I did not realize that this PR is still open. I changed the code to use `JSplitPane`. I also added some minimum size restrictions to make sure users can't bring jadx into a state where important components are not visible.

PR #612, opened 2019-04-20
fix: remove redundant cast over similar conditions

asashour:


jadx-core/src/main/java/jadx/core/codegen/ConditionGen.java
```diff
@@ -144,12 +144,18 @@ private void addNot(CodeWriter code, CondStack stack, IfCondition condition) thr
...
+		String modeStr = mode == Mode.AND ? " && " : " || ";
 		Iterator<IfCondition> it = condition.getArgs().iterator();
 		while (it.hasNext()) {
-			wrap(code, stack, it.next());
+			IfCondition next = it.next();
+			if (next.getMode() == mode) {
+				add(code, stack, next);
+			} else {
+				wrap(code, stack, next);
+			}
```
skylot: Can you check: is it possible to move this mode check inside `wrap` method because it already contains call: `boolean wrap = isWrapNeeded(cond);` or maybe inside `isWrapNeeded` method?

I prefer to localize all logic in code generators or better - made such optimizations in visitors before codegen by using flags (like `DONT_WRAP`).

asashour:
This is now based on #627

Issue #165, opened 2017-03-10
Proper search command for macOS


krokofant:
Search is usually binds to CMD+F instead of CTRL+F on macOS. While copying and pasting things into search and closing the search one has to do a bit of a gymnastics with their hands.

jadx version 0.6.1
jadx-gui version 1.0 (?)

asashour:
For macOS, should the other shortcuts be also with <kbd>⌘</kbd>?

As a side note, should we change the Windows search to be <kbd>Ctrl </kbd>+<kbd>F</kbd> instead of <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd>?

skylot:
@asashour sorry I don't have macOS so I don't know which shortcuts are common and must be used. If you know how to make correct shortcuts and can test it - PR are welcome.

As for search, jadx use both shortcuts: <kbd>Ctrl </kbd>+<kbd>F</kbd> for search in current file and <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd> for global text search (same shortcuts used in IntelliJ Idea and Android Studio).

asashour:
I don't have a macOS either.

I see now the shortcut (in the code), but couldn't figure it out from the menu.

sergey-wowwow:
@skylot 
Why do you use in ``` jadx-gui/src/main/java/jadx/gui/ui/MainWindow.java ```
```java
		textSearchAction.putValue(Action.SHORT_DESCRIPTION, NLS.str("menu.text_search"));
		textSearchAction.putValue(Action.ACCELERATOR_KEY,
				getKeyStroke(KeyEvent.VK_F, KeyEvent.CTRL_DOWN_MASK | KeyEvent.SHIFT_DOWN_MASK));
```
with Shift pressed, but for other actions don't?

I have a Mac, so I'm doing to implement it (requirement to use Ctrl instead on Cmd makes me nervous for a long time :D)

skylot:
@sergey-wowwow because jadx use both shortcuts: 
* <kbd>Ctrl </kbd>+<kbd>F</kbd> for search in current file
* <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd> for global text search

same shortcuts used in IntelliJ Idea and Android Studio

PR #616, opened 2019-04-21
Implements command (CMD) button for MacOS

sergey-wowwow:
Issue https://github.com/skylot/jadx/issues/165


sergey-wowwow:
@skylot 
Let's merge that and we're gonna get only 3 pages of issues, but not 4 :D

PR #632, opened 2019-04-25
fix: loading aar/jar files (#631)

### LINKED ISSUE #631: Support for (specific?) AAR files
I found out that this commit:

SHA-1: 79ccaadaff18de7443491dbc333e71cfa6e3e3d5
* fix: handle big .jar files (using multi-dex option) (#390) (PR #568)

breaks decompilation for AAR libraries that I use. The result is that when loading those libraries into jadx, after a split-second it shows just empty "Source code" folder.

Unfortunately I cannot share with you the specific AARs. Is there any other way I can contribute to fixing this?

asashour:
By overwriting the created temp file

Fix #631

### DIFF
--- a/jadx-core/src/main/java/jadx/core/utils/files/InputFile.java
+++ b/jadx-core/src/main/java/jadx/core/utils/files/InputFile.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jadx-core/src/main/java/jadx/core/utils/files/InputFile.java
```diff
@@ -134,7 +135,7 @@ private boolean loadFromZip(String ext) throws IOException, DecodeException {
 							case ".jar":
 								index++;
 								Path jarFile = Files.createTempFile(entryName, ".jar");
```
skylot: I just notice that in previous commits you replace many calls from `FileUtils.createTempFile` to `Files.createTempFile` and now temp files not deleted on exit and pollute temp dir, because `FileUtils.createTempFile`  invoke `file.deleteOnExit();` to remove temp file, but now it is not :(
Also names was nice like starting with `jadx-tmp-` so it easier to understard which app created it:
```
before: /tmp/jadx-tmp-8207354968595972366-classes.jar
after: /tmp/classes.jar8207354968595972366.jar
```
So, please, next time try to not change current behavior heavily, because sometimes it is hard to notice such changes on review :((

asashour: You are right, I have a bad habit of mixing multiple changes.

I thought using the new API would be better.

A new PR is underway.

PR #634, opened 2019-04-25
fix: always use FileUtils.createTempFile

asashour:


jadx-core/src/main/java/jadx/core/utils/files/FileUtils.java
```diff
@@ -65,24 +65,13 @@ public static void makeDirs(@Nullable File dir) {
...
-		} catch (IOException e) {
-			throw new JadxRuntimeException("Failed to create temp file with suffix: " + suffix);
-		}
-		return temp;
-	}
-
-	public static File createTempDir(String suffix) {
-		try {
-			Path path = Files.createTempDirectory("jadx-tmp-" + System.nanoTime() + '-' + suffix);
+			Path path = Files.createTempFile("jadx-tmp-", System.nanoTime() + '-' + suffix);
```
skylot: Looks like `System.nanoTime()` not needed here, and was not needed in the previous version :)

PR #635, opened 2019-04-25
feat(gui): show smali (#197)

### LINKED ISSUE #197: Have an option to display smali code side-by-side with decompiled code
Since some decompilation might fail or be inaccurate, it would be nice to have an option to view the original smali code of a class/method.

asashour:
Fix #197 

@jpstotz has kindly contributed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,31 +54,38 @@ Run **jadx** on itself:
 
 ### Usage
 ```
-jadx[-gui] [options] <input file> (.dex, .apk, .jar or .class)
+jadx[-gui] [options] <input file> (.apk, .dex, .jar, .class, .smali, .zip, .aar, .arsc)
 options:
- -d,  --output-dir           - output directory
- -ds, --output-dir-src       - output directory for sources
- -dr, --output-dir-res       - output directory for resources
- -j,  --threads-count        - processing threads count
- -r,  --no-res               - do not decode resources
- -s,  --no-src               - do not decompile source code
- -e,  --export-gradle        - save as android gradle project
-      --show-bad-code        - show inconsistent code (incorrectly decompiled)
-      --no-imports           - disable use of imports, always write entire package name
-      --no-replace-consts    - don't replace constant value with matching constant field
-      --escape-unicode       - escape non latin characters in strings (with \u)
-      --deobf                - activate deobfuscation
-      --deobf-min            - min length of name
-      --deobf-max            - max length of name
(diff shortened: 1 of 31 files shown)


jpstotz:
Hi Ahmed,

I just gave your smali branch a try and have to say this looks quite useful. However I noticed that when a class is decompiled also automatically the smali code is generated. In general I would recommend to generate the smali code on-demand when the smali tab is clicked the first time (should be possible via `addChangeListener` on the `areaTabbedPane`). Then the user don't have to spend time and memory for generating and storing the smali code that may be never displayed.

asashour:
Thanks for trying it out. Yes, it should be shown on demand.

jpstotz:
I managed to eliminate one major draw-back of your implementation: The forced decompilation of all classes to temporary files. My implementation isn't perfect, but at least doe snot require any temporary files and it decompiles only one class.

To tried to make a pull request for your smali branch, but somehow GitHub refuses to do so (does not allow to select your jadx fork as pull request target). Therefore here is the link to my [smali branch](https://github.com/jpstotz/jadx/tree/smali) and my commit:
https://github.com/jpstotz/jadx/commit/840b5f284edf07420f995c5635ec399ab1833901

asashour:
Very nice, that is indeed the major issue, I will include your changes.

asashour:
I was able to make PR to the fork, from the web, you can select the source fork and destination fork.

image

jpstotz:
@asashour Your screenshot shows exactly the point where I had to stop, because Github does not allow me to select your `asashour/smali` branch as base repository on the left side. 
I get a long list of jadx forks as base repository but yours is missing in the list. When I use the integrated search and search for "asa" or "asashour" no entry is left.

asashour:
@jpstotz thanks for pointing this out, would [this](http://srcmvn.com/blog/2013/03/14/sending-a-pull-request-on-github-even-when-the-base-repository-you-want-is-missing/) help?

cobalt232:
How about showing the byte code for each instruction in the small tab. Maybe in a row on the left?

asashour:
> How about showing the byte code for each instruction in the small tab. Maybe in a row on the left?

Can you elaborate more? Why the actual byte code would be important?

Also, we use Baksmali, and it would be difficult to implement that, without existing options.

cobalt232:
It would be useful for dex modifications on byte code level, as used in my project "Blackberry Manager".

Issue #532, opened 2019-03-28
References to not existing classes (deobfuscator doesn't rename outer class names)


sergey-wowwow:
Hi, 
Open class ``` com.paypal.merchant.client.features.webview.PPBWebViewActivity ```
```java
public class PPBWebViewActivity extends p000.C5744Dy {
```
then open ``` p000.C5744Dy ```
```java
/* renamed from: Dy */
public abstract class C5744Dy extends p000.C1430y {
```
and finally open ``` p000.C1430y ```
```java
/* renamed from: y */
public class C1430y extends p000.Hh implements p000.C1439z, p000.je.C1314a, p000.C1331n {
```
class ``` p000.je.C1314a ``` doesn't exist as tons of other classes referenced by ``` p000.**.C***** ```

Actually that interface was moved to
```java
package p000;

/* renamed from: je */
public final class C5243je implements java.lang.Iterable<android.content.Intent> {
    //...

    /* renamed from: je$a */
    public interface C1314a {
        android.content.Intent getSupportParentActivityIntent();
    }
```

Please use this APK: https://drive.google.com/file/d/1_8--CfspEnugSTGjF3Krkt4r3SEV6vq7/view?usp=sharing

skylot:
@sergey-wowwow I can't reproduce this issue on latest master, maybe it was fixed in #474?
Can check it again.

sergey-wowwow:
@skylot
No, even in jadx-gui Search feature shows not deobfuscated class names (when deobfuscation is enabled) for the given example

sergey-wowwow:
@asashour,
Can you please check it too, it that reproducible on your side?

asashour:
It works fine with me:

The implemented interface is an inner class, check the the import statements of the my `C2658y`

```
import p000.C2885Dy;
public class PPBWebViewActivity extends C2885Dy {
```

```
/* renamed from: Dy */
public abstract class C2885Dy extends C2658y {
```

```
import p000.C2198je.C2199a;
public class C2658y extends C0234Hh implements C2690z, C2199a, C2305n {
```


```
/* renamed from: je */
public final class C2198je implements Iterable<Intent> {

    ...

    /* renamed from: je$a */
    public interface C2199a {
        Intent getSupportParentActivityIntent();
    }

```

sergey-wowwow:
[...]

PR #649, opened 2019-04-29
fix: use simple class name for sibling class (#648)

### LINKED ISSUE #648: Don't use fully qualified class name when it is in the same package.
Even if we have `use imports` to false, we shouldn't write the fully qualified class name when it is a sibling class (both are in the same package).

asashour:
Fix #648


skylot:
Issue #532 fixed, so we don't need this workaround anymore

PR #650, opened 2019-04-29
Enforce code style using checkstyle and spotless with eclipse formatter

skylot:
To check and force use one code style I add [checkstyle](https://checkstyle.org/index.html) and [spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle) with eclipse formatter.
These tools work in tandem and help maintain same code style using config which can be used in all IDEs or from command line.

I will add wiki page about using and configuring these tools later.
Here I will give only short description to get started:
- spotless - a gradle plugin checks formatting by reformating all code and fail build it diff exists.
For format code spotless use Eclipse formatter using config located in `config/code-formatter`. 
In Eclipse these files can be just imported.
 In Idea [Eclipse formatter plugin](https://github.com/krasa/EclipseCodeFormatter#instructions) can be used. 
Also, all code can be reformated with: `gradle spotlessApply`
- checkstyle - check other code style issues like unused imports, bad naming, etc.
Config located in `config/checkstyle` folder. Many IDEs have plugins to run checkstyle on the fly.

All these checks can be executed with `gradle check -x test` and also will be executed on build (`gradle build`).
So before pushing commit anyone can run:
```bash
gradle spotlessApply
gradle check -x test
```
Then fix all issue and run these command again until all checks passed.


asashour:
Nice, please go ahead, I can update my PRs. You don't to write wikis, the internet is full of resources.

skylot:
Also from now, I will reject any PRs which change only code style without fixing any bugs and forbid any other style changes without automated checks. There are many reasons for this decision:
1. I don't like change code just for change code without any real improvements
2. Style changes break `git blame` and destroy history of changes and relation to issues
3. There is no **universal** code style for java, so I don't want open dispute how to better write code.
4. Checkstyle and other automated tools provide tons of checks and it is enough, other style issues which are not automated not make code worst or better, so it is not worth changing.

skylot:
@asashour my previous comment means that I will close all your `chore` PRs. Because they contains a lot of changes which don't needed or I don't like. And I don't want to open discussion on every small change, sorry :(

asashour:
No worries, but how about sonar issues?

skylot:
Only issues in category 'bug' and 'vulnerability' is worth fixing. Others are mostly style checks and not really harm.
I think, I will try to add `findbugs` or `pmd` with config which will fail only on really important issues, like possible NPE or similar.

jpstotz:
I had some problems with `gradle spotlessApply` on Windows. By default I use git with `core.autocrlf true` therefore my local files use CRLF for newlines. After `gradle spotlessApply` all files were changed to LF only - therefore git recognized each and every file as modified.

I changed to `core.autocrlf input` and reset and re-implement my changes.

asashour:
Just a hint: even if you add all files to the index and commit, the EOL is automatically handled by git(hub), at least with me

skylot:
@jpstotz thanks for the report, looks like this must be mentioned in contributing guide (after I made one).
Anyway, I made this intentionally because I don't like auto changes in any way. So you are forced to use LF even on Windows.
Also, I can suggest committing changes before `gradle spotlessApply` and after amending fixes.

jpstotz:
@skylot Line endings are a common problem for me - the whole git system never worked very reliable for me. Sometime the local Git client (Eclipse, TortoiseSVN, IntelliJ) do only show the real changed lines and when I push the commit to Github all lines are marked changed in the PR.

This time I ended up in hundred of files marked as "modified" by git but in the IntelliJ compare UI no change was visible (even the line endings were correct). I don't know exactly how SVN managed to solve the line-ending problem but it worked much more reliable then the system introduced by git.

> Also, I can suggest committing changes before gradle spotlessApply and after amending fixes.

But then the PR would contain two commits, one for the changes and one for the style changes. I assumed that this is unwanted. Or is there a simple and fast way to handle this situation?

skylot:
> the whole git system never worked very reliable for me

That's why I don't like `autocrlf` and always disable it. But doing so needs to adjust IDE you are using :(

> But then the PR would contain two commits, one for the changes and one for the style changes

I think you are mistaken 'commit' and 'push'. Git allows to change local (not yet pushed) commit locally as you want. So you can use `git commit --amend` to change local commit before pushing it to remote repository.

Anyway, commit count in PR doesn't matter because I always use GitHub 'squash merge'. This merge (squash) all commits in one and rebase result on latest master. This is most cleaner way - no merge commits and easier history (no branching), also this allows me to fix resulting commit message if needed.

jpstotz:
@skylot can we please apply `gradle spotlessApply` once on the master branch (and commit the result)?
As it modifies nearly every java file in the repo, applying it on a branch is not a good idea as afterwards you can't distinguish between style changes and real code changes.

skylot:
@jpstotz sorry but I don't understand what you mean.
`gradle spotlessApply` already applied to the master and I don't have any changes after running it again.
What kind of changes do you see?

jpstotz:
@skylot Sorry my mistake. I was just trying to prepare a PR that extends the code base from #652 and in the end GitHub showed me that the branch can't be merged into your master branch. Therefore I thought this is because of the style changes (but that wasn't the problem). 

I now created it as PR #653 which has some conflicts I need to resolve first.

PR #652, opened 2019-04-30
feat(gui): make search bar usable for smali code

jpstotz:
Additionally this PR includes a small fixe (start with line 0) for the smali tab and some unnecessary code in JadxDecompiler.java is removed.


skylot:
I think search bar must be moved inside `*Area` so we must have two independent search bars.
But your implementation work, so we can make these improvements later.

jpstotz:
@skylot I agree with you in the end there should be one search for one RSyntaxTextArea/CodeArea. 
Anyway there is some more work needed as I noticed that the Code/Smali tab on the bottom currently also appears for text based JResource nodes like AndroidManifest.xml which may be confusing for the users.

PR #653, opened 2019-05-03
Separate SearchBar for Java code area and smali code area

jpstotz:
For doing so some classes were renamed and the class hierarchy was changed:

1. The `CodePanel` has been split up into one panel for Java Classes (with smali tab) and one for text based reources (without smali tab).
2. `CodePanel` has been renamed to `CodeContentPanel` (as the new `CodePanel` now is only a CodeArea + scrollable SearchBar)
3. The CodeArea for Java code and Smali code now use the same abstract base class `AbstractCodeArea`.


skylot:
Good. One thing: looks like settings (like font size) not apply to smali tab on creation. To reproduce: change font size in settings-> open another not yet opened class -> go to smali tab -> still old font size.

P.S. Is it really usefull to view smali in separate tab? For me side by side view will be much better because it allows to see and compare at a same time both java code and smali.
Like in [Bytecode Viewer](https://github.com/Konloch/bytecode-viewer) (jadx was added recently :+1:)
Or maybe we need option to switch from tabs to split view.

jadx-core/src/main/java/jadx/api/JadxDecompiler.java
```diff
@@ -325,7 +326,7 @@ void generateSmali(ClassNode cls) {
 			if (!decompiled) {
 				LOG.error("Failed to find smali class {}", className);
 			}
-		} catch (Exception e) {
+		} catch (IOException e) {
```
skylot: Can you leave `Exception` for catch possible NPE :)

jadx-gui/src/main/java/jadx/gui/ui/codearea/SmaliArea.java
```diff
@@ -1,21 +1,22 @@
...
-import org.fife.ui.rsyntaxtextarea.RSyntaxTextArea;
-
 import jadx.gui.treemodel.JNode;
+import jadx.gui.ui.ContentPanel;
 
-public final class SmaliArea extends RSyntaxTextArea {
+public final class SmaliArea extends AbstractCodeArea {
 	private static final long serialVersionUID = 1334485631870306494L;
 
 	private final JNode node;
```
skylot: `node` field already exists in `AbstractCodeArea` so can be removed here

jadx-gui/src/main/java/jadx/gui/ui/codearea/ClassCodeContentPanel.java
```diff
@@ -0,0 +1,74 @@
...
+	public ClassCodeContentPanel(TabbedPane panel, JNode jnode) {
+		super(panel, jnode);
+
+		javaCodePanel = new CodePanel(this);
+		smaliCodePanel = new CodePanel(this) {
+			@Override
+			protected AbstractCodeArea createCodeArea(ContentPanel contentPanel) {
+				return new SmaliArea(contentPanel);
+			}
+		};
```
skylot: Can we replace this method override with additional parameter in `CodePanel` constructor?

jadx-gui/src/main/java/jadx/gui/ui/codearea/CodePanel.java
```diff
@@ -5,60 +5,57 @@
...
-	private final SmaliArea smaliArea;
+	private final AbstractCodeArea codeArea;
 	private final JScrollPane codeScrollPane;
-	private final JScrollPane smaliScrollPane;
-	private JTabbedPane areaTabbedPane = new JTabbedPane(JTabbedPane.BOTTOM);
 
-	public CodePanel(TabbedPane panel, JNode jnode) {
-		super(panel, jnode);
+	public CodePanel(ContentPanel contentPanel) {
+		super();
```
skylot: Super calls without arguments can be removed

PR #657, opened 2019-05-08
feat(cli): decompile only a single class

jpstotz:
Decompilation a single single class only was not yet possible. In the backend a flexible Predicate<String> (String=full class name) is used. Therefore implementing different filters would be possible.


skylot:
Thanks! Nice implementation!

And I want this feature also in jadx-gui it will be very helpful for testing :)
(I mean for class passed using this option open new tab and sync with class tree)

xzxiaoshan:
How to use the parameter "-- single class" ?

PR #658, opened 2019-05-09
feat(gui): select a class via command-line to open in Jadx

jpstotz:
Works only for non-inner classes

### DIFF
--- a/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
+++ b/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
@@ -130,12 +130,16 @@ public class JadxCLIArgs {
 	 * Used to merge saved options and options passed in command line.
 	 */
 	public boolean overrideProvided(String[] args) {
-		JCommanderWrapper<JadxCLIArgs> jcw = new JCommanderWrapper<>(new JadxCLIArgs());
+		JCommanderWrapper<JadxCLIArgs> jcw = new JCommanderWrapper<>(newInstance());
 		if (!jcw.parse(args)) {
 			return false;
 		}
 		jcw.overrideProvided(this);
 		return process(jcw);
+	}
+
+	protected JadxCLIArgs newInstance() {
+		return new JadxCLIArgs();
 	}
 
 	private boolean process(JCommanderWrapper<JadxCLIArgs> jcw) {
--- a/jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
+++ b/jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
@@ -15,12 +15,14 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 
-import javax.swing.*;
+import javax.swing.JFrame;
 
 import org.fife.ui.rsyntaxtextarea.RSyntaxTextArea;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.beust.jcommander.Parameter;
 
 import jadx.api.JadxArgs;
 import jadx.cli.JadxCLIArgs;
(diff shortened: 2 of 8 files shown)


jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
```diff
@@ -130,14 +130,18 @@ public boolean processArgs(String[] args) {
...
 		if (!jcw.parse(args)) {
 			return false;
 		}
 		jcw.overrideProvided(this);
 		return process(jcw);
 	}
 
+	protected JadxCLIArgs newInstance() {
+		return new JadxCLIArgs();
+	}
```
skylot: This method not needed.
And please, don't change code not needed for your fix.

jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
```diff
@@ -65,6 +67,10 @@
 
 	private int settingsVersion = 0;
 
+	@JadxSettingsAdapter.GsonExclude
+	@Parameter(names = { "-sc", "--select-class" }, description = "GUI: Open the selected class and show the decompiled code")
+	private String cmdSelectClass = null;
```
skylot: You don't need to add this field, it is already inherited from `JadxCLIArgs`

jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
```diff
@@ -412,4 +422,8 @@ private void upgradeSettings(int fromVersion) {
 		sync();
 	}
 
+	@Override
+	protected JadxCLIArgs newInstance() {
+		return new JadxSettings();
+	}
```
skylot: Remove this also

jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
jpstotz: I think this is a misunderstanding. This option can not be inherited from `JadxCLIArgs` because it does not exist in jadx-cli as it has no GUI to select a class.

I don't want to misuse one of the existing command-line arguments as this option is GUI only.

jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
jpstotz: This change is required to make the parameter `cmdSelectClass` in JadxSettings work (used by  `JadxCLIArgs.overrideProvided(..)` . 

Or do you prefer creating a new instance via reflection or may be `clone()`?

jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
skylot: Hm, sorry, I thought you will use `--single-class` option (easier way), but you add a new one just for jadx-gui. And now for me unclear what `--single-class` will mean for jadx-gui :)

jpstotz: I have not checked it but from my perspective there are other command-line options like `--export-gradle` or `--no-src` that should not have an effect in jadx-gui. 

To make it clean there would be still the possibility to split `JadxCLIArgs` into two classes, one base class for argument that affect settings settings (and as base class for `JadxSettings`) and a second class that inherits base class and contains the cli-only arguments.

skylot: Yeah, you are right, these must be different classes.
I just too lazy and don't want to write more code and more classes :)

skylot: Hm, I have merged PR, but just notice that `jadx-gui --help` don't show new option :(

jpstotz: I managed to fix this issue. See PR #660

Issue #692, opened 2019-06-20
1.0.0版本闪退


xmhwws:
使用不包含jre的版本，运行jadx-gui.bat闪退。
使用含有jre的版本，正常运行jadx-gui.bat。
我的java version是 "1.8.0_181"，系统为windows10 x64 专业版。

jpstotz:
I am sorry please use English as language for writing issues and include a concrete description of your problem. From the online translated version of your issue I don't understand what your problem is.

xmhwws:
> I am sorry please use English as language for writing issues and include a concrete description of your problem. From the online translated version of your issue I don't understand what your problem is.

Because my English is not good, so I used Chinese previously, I am sorry.

Run jadx-gui.bat crash with a version that does not contain jre.
Use the jadx version with JRE to run jadx-gui.bat properly.

My java version is "1.8.0_181" and the system is Windows 10 x64 Professional Edition.

Can you understand what I said?

jpstotz:
Open a command-line (`cmd.exe`) an test the following commands:

1. `java -version` --> output should include the version of your java runtime "1.8.0_181"
2. `where java.exe`  -> list of available java executables
3. `echo %JAVA_HOME%`

Alternatively you can start jadx using `jadx-gui.exe` launcher. It searches automatically for an appropriate installed JRE.

xmhwws:
> Open a command-line (`cmd.exe`) an test the following commands:
> 
> 1. `java -version` --> output should include the version of your java runtime "1.8.0_181"
> 2. `where java.exe`  -> list of available java executables
> 3. `echo %JAVA_HOME%`
> 
> Alternatively you can start jadx using `jadx-gui.exe` launcher. It searches automatically for an appropriate installed JRE.

Commands doesn't work，jadx-gui.bat still crash with a version that does not contain jre.
I use jadx-gui-1.0.0-with-jre-windows now,it can work properly.
I hope that jadx-1.0.0.zip  can also work properly.
If there is no solution to this problem for the time being, ignore it. 
I'm sorry to disturb you,thank you very much.

jpstotz:
The commands were not to fix the problem but to identify it. Please post the output.

13-beta2:
Most probably caused by introduced bug in jadx-gui.bat - JAVA_EXE quoted twice. To fix just replace 
```
:findJavaFromJavaHome
set JAVA_HOME=%JAVA_HOME:"=%
[...]

Issue #686, opened 2019-06-09
Not finding JAVA path in windows batch script


tvalle:
Commenting here because the jadx-gui.bat file doesn't seem to be in the repo, so I couldn't send a pull request.
On the bundle with the JRE, I was having an issue with the path not being found, after messing with it I discovered the problem:

Instead of:
```
19 set JAVA_HOME="%APP_HOME%\jre"
20 set JAVA_EXE=%JAVA_HOME%\bin\javaw.exe
```

It should be:
```
19 set JAVA_HOME="%APP_HOME%\jre"
20 set JAVA_EXE="%JAVA_HOME%\bin\javaw.exe"
```

On line 19 it was making the expansion all right but when it passed down to 20 because it didn't have double quotes both cmd and powershell were unable to find the path because mine had spaces in it.

skylot:
@tvalle thanks! I apply your fix in release bundle and in master branch.

PR #694, opened 2019-06-21
fix(gui): jadx-gui.bat does not work (#692)

jpstotz:
In `jadx-gui.bat` the command for starting `javaw.exe` contains in the end double double quotes which prevents starting jadx-gui.

This is the fix for issue #692

### DIFF
--- a/jadx-gui/build.gradle
+++ b/jadx-gui/build.gradle
@@ -51,7 +51,7 @@ startScripts {
         def str = windowsScript.text
         str = str.replaceAll('java.exe', 'javaw.exe')
         str = str.replaceAll('"%JAVA_EXE%" %DEFAULT_JVM_OPTS%', 'start "jadx-gui" /B "%JAVA_EXE%" %DEFAULT_JVM_OPTS%')
-        str = str.replaceAll('set JAVA_EXE=%JAVA_HOME%/bin/javaw.exe', 'set JAVA_EXE="%JAVA_HOME%/bin/javaw.exe"')
+        str = str.replaceAll('set JAVA_EXE=%JAVA_HOME%/bin/javaw.exe', 'set JAVA_EXE=%JAVA_HOME%/bin/javaw.exe')
         windowsScript.text = str
     }
 }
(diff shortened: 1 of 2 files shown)


jadx-gui/build.gradle
```diff
@@ -51,7 +51,7 @@ startScripts {
         def str = windowsScript.text
         str = str.replaceAll('java.exe', 'javaw.exe')
         str = str.replaceAll('"%JAVA_EXE%" %DEFAULT_JVM_OPTS%', 'start "jadx-gui" /B "%JAVA_EXE%" %DEFAULT_JVM_OPTS%')
-        str = str.replaceAll('set JAVA_EXE=%JAVA_HOME%/bin/javaw.exe', 'set JAVA_EXE="%JAVA_HOME%/bin/javaw.exe"')
+        str = str.replaceAll('set JAVA_EXE=%JAVA_HOME%/bin/javaw.exe', 'set JAVA_EXE=%JAVA_HOME%/bin/javaw.exe')
```
skylot: Looks like we can remove this line completely. Also please apply this change also in `jadx-cli/build.gradle` .
And for reference I made these changes here: a051ce6cf4ec2f9a431366658eebbcc988d9d11d, sorry :(

jpstotz: LOL you are right. I did not recognize that the line is now totally useless.

skylot:
@jpstotz thank you!
But it bothers me why I made such changes, the original discussion started in #686, and looks like we need to review that issue again, because just now we reverted suggested changes :(

jpstotz:
My JAVA_HOME is set to `C:\Program Files\AdoptOpenJDK\jdk-11.0.3.7-hotspot` - therefore it contains spaces. And before my last commit I tested both batch files (jadx-gui.bat and jadx.bat) and they worked without problems.
When I enable debug mode I can see that the final call of java.exe (for jadx.bat) is `"C:\Program Files\AdoptOpenJDK\jdk-11.0.3.7-hotspot/bin/java.exe" "-Xms128M" "-Xmx4g" "-XX:+UseG1GC" ...`. Hence it is correctly quoted.

Was any related component (gradle or gradeplugin) updated in the meantime? If yes this could have fixed this issue at a different point so that we ended up in a "double fix".

PR #733, opened 2019-07-31
Add unit tests for StringUtils, OverlayIcon and PrimitiveType

shafi-1:
I've analysed your codebase and noticed that jadx.tests.functional.StringUtils, jadx.gui.utils.OverlayIcon and jadx.core.dex.instructions.args.PrimitiveType are not fully tested.
I've written some tests for the methods in these classes with the help of Diffblue Cover.

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important in a subsequent PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,6 +41,9 @@ allprojects {
 		testCompile 'org.hamcrest:hamcrest-library:2.1'
 		testCompile 'org.mockito:mockito-core:2.28.2'
 
+		testCompile group: 'junit', name: 'junit', version: '4.8.2'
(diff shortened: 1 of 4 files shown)


skylot:
@shafi-1 
Are these tests were generated automatically?
Because for me they look useless, they do not cover most cases just weird ones, uses reflection (which is bad), also you add an old version of JUnit and this is unacceptable.
So I will don't merge this, sorry.

Issue #763, opened 2019-10-15
[core] Jadx does not load all classes (IndexOutOfBoundsException)


jpstotz:
When loading apk of com.whatsapp 2.19.274 (452992) not all classes within the package `X` are loaded (and are therefore missing in the class tree). There is an exception which silently stops the loading process:

```
jadx.core.utils.exceptions.JadxRuntimeException: Error decode class: X.1cX
	at jadx.core.dex.nodes.ClassNode.<init>(ClassNode.java:126)
	at jadx.core.dex.nodes.DexNode.loadClasses(DexNode.java:58)
	at jadx.core.dex.nodes.RootNode.load(RootNode.java:85)
	at jadx.api.JadxDecompiler.load(JadxDecompiler.java:91)
	at jadx.gui.JadxWrapper.openFile(JadxWrapper.java:42)
	at jadx.gui.ui.MainWindow.open(MainWindow.java:281)
	at jadx.gui.ui.MainWindow.processCommandLineArgs(MainWindow.java:164)
	at jadx.gui.ui.MainWindow.init(MainWindow.java:157)
	at java.desktop/java.awt.event.InvocationEvent.dispatch$$$capture(InvocationEvent.java:313)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:770)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:715)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:85)
	at java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:740)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:203)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:124)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:113)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:109)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
[...]

PR #764, opened 2019-10-15
fix: don't stop loading classes in case of an error

jpstotz:
As described in #763 Jadx silently stops loading classes form an APK like com.whatsapp 2.19.274 (452992) which causes one class loading to fail.

This PR changes the class loading behavior in a way that only the affected class is not loaded. In case of an error the error is logged and class loading continues.


skylot:
@jpstotz thanks!

PR #773, opened 2019-11-01
feat(gui): Support for APK signature v3 added

jpstotz:
The old apk signature test had problems if the APK contained a v3 signature. 
The updated apksig library can handle v1-v3 signatures correctly. 
Furthermore I moved the verification warnings below the signature block as there are usually a lot of them, but they are not more important that the signature verification result.


skylot:
@jpstotz thanks!

PR #779, opened 2019-11-14
Fix: escapeUnicode always-on for `c < 32`

nis-spiir:
The escape unicode logic checks `c < 32` OR `c >= 127 && escapeUnicode`.

Maybe the feature should be on by default and the flag should be changed to `--no-escape-unicode` ?

I haven't run `gradle spotlessApply` as I can't get gradle to work probably.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/utils/StringUtils.java
+++ b/jadx-core/src/main/java/jadx/core/utils/StringUtils.java
@@ -64,7 +64,7 @@ public class StringUtils {
 				break;
 
 			default:
-				if (c < 32 || c >= 127 && escapeUnicode) {
+				if ((c < 32 || c >= 127) && escapeUnicode) {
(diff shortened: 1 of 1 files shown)


nis-spiir:
The tests that fail is because they assume "encode unicode" is on by default.

skylot:
@nis-spiir I think the original check was made like this intentionally and not by mistake.
I was trying to hide not printable chars by showing as char code and as soon as all chars < 32 is not printable that was the desired output. But for higher codes it is harder to decide is char printable or not, so I add an option `--escape-unicode`. And that option was set to false by default because strings mostly contain printable characters like Unicode chars from different languages, so users can read these strings. Binary data in strings are less common but still can be treated with option set to true.
Anyway, thank you for this suggestion, look like I need to add some comments here :)

Issue #791, opened 2019-12-10
[feature] Support for renaming classes, methods and fields


S-trace:
Hello.

I added a Rename popup menu item that allows the user to rename classes, methods and fields.
It is useful for obfuscated code, it makes it much easier to read.
The implementation is very basic and perhaps not optimal, but it works.
I pushed it to https://github.com/S-trace/jadx/tree/rename
Could you please review it and allow me to open a pull request?

Thanks.

skylot:
@S-trace good implementation, thanks!
Although your approach is a little different from what I describe in #524 [in this comment](https://github.com/skylot/jadx/issues/524#issuecomment-477120938), I think it is a good start, so we can improve this later. Please, open a PR :)
Also, this implementation will close #88 and this is great!

S-trace:
I'm working right now for a more advanced implementation of this feature, which will not reload whole file, but now I'm have a problem with renaming inner classes renaming and with renaming classes in methods parameters (regular classes renaming, methods renaming and fields renaming works fine).

I hope i can finish this update and push another commit to the same branch.
But it's possible to merge current implementation now and wait for updated implementation (no feature is worse than slow but working feature).

Thank you.

S-trace:
I pushed updated rename implementation to https://github.com/S-trace/jadx/tree/rename2
Could you please review and test it?

It still have a problem with renaming classes from import statements and method arguments list, but all other usecases seems working fine.
I'm not too familiar with jadx architecture, so any tips and hints are welcome.

Thank you!

S-trace:
About performance:
Tried to rename 2 classes in the project containing 6622 classes, 1249 of them excluded. classes.dex size is 7665080 bytes.
Rename was performed twice: first time in clean project with only renaming class open (best case, only renamed class will be redecompiled), second time after performing "Find usages" for the class (worst case, all classes referring to renamed class will be redecompiled).

Results:

Class1: 217725 bytes, 5079 lines
Clean state:
refreshState() took 1692 ms to update state, 0 classes to be upgraded in background (1768 ms total)

After "Find usages"
refreshState() took 1801 ms to update state, 121 classes to be upgraded in background (1875 ms total)
INFO  - Memory usage: Before refresh: heap: 238MB, allocated: 795MB, free: 556MB, total free: 3857MB, max: 4096MB
INFO  - Memory usage: After refresh: heap: 239MB, allocated: 795MB, free: 555MB, total free: 3856MB, max: 4096MB, refresh took 989 ms

[...]

PR #794, opened 2019-12-10
Add support for renaming methods, classes and fields

S-trace:
This request adds a Rename popup menu item that allows the user to rename classes, methods and fields.
It is useful for obfuscated code, it makes it much easier to read.
The implementation is very basic and perhaps not optimal, but it works.

I'm working for a better implementation which will not reload whole file at every rename, but it's still in progress.

This request is related to issue #791 but does not completely closes it - there is a place for improving this implementation,
However, no feature is worse than slow, but working feature - so, it's will be good to merge current implementation and then improve it.

### DIFF
--- a/jadx-core/src/main/java/jadx/api/JadxDecompiler.java
+++ b/jadx-core/src/main/java/jadx/api/JadxDecompiler.java
@@ -281,7 +281,7 @@ public final class JadxDecompiler {
 		root.getErrorsCounter().printReport();
 	}
 
-	RootNode getRoot() {
+	public RootNode getRoot() {
 		return root;
 	}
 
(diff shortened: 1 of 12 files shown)


ppwwyyxx:
Hi, thanks for adding this feature!
However, when I clicked rename, a window like this popup:

and nothing happened after I clicked rename in the window. Looks like something isn't working well

jpstotz:
That seems to be a problem with the GUI of the rename dialog. The dialog itself can be resized. If you do so and increase the height the text field where you can change the name becomes visible.

S-trace:
Just for anyone interested in the advanced rename feature - please check out the https://github.com/skylot/jadx/tree/rename branch to test it.

Please report any issues to the https://github.com/skylot/jadx/issues/791 issue (if any).

PR #813, opened 2019-12-23
Fix

Henry2o1o:


skylot:
@Henry2o1o thanks for noticing this, but I think I will add these dots in code because I plan to add some descriptions for additional actions like task canceling [here](https://github.com/skylot/jadx/blob/9c88f70740b7e8b326534afada316720f48a38a8/jadx-gui/src/main/java/jadx/gui/jobs/BackgroundExecutor.java#L128).

PR #839, opened 2020-01-27
Re-ordering Stream APIs

HirotoTanaka114:
We are conducting a research project on re-ordering Stream APIs for improving program execution speed.

We executed this command before and after re-ordering : 
`time ./gradlew -Dtest.single=TestInlineVarArg test`

and the result is :
```
the order of Stream API      test running time
===============================================
distinct().sorted()                 10.640s
sorted().distinct()                 10.610s
```

### DIFF
--- a/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
+++ b/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
@@ -328,7 +328,7 @@ public class ClassGen {
 	public void insertDecompilationProblems(CodeWriter code, AttrNode node) {
 		List<JadxError> errors = node.getAll(AType.JADX_ERROR);
 		if (!errors.isEmpty()) {
-			errors.stream().distinct().sorted().forEach(err -> {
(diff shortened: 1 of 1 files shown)


skylot:
@HirotoTanaka114 can you share your research? Or a reason why such reordering can improve performance? I found [this](https://stackoverflow.com/questions/43806467/java-streams-how-to-do-an-efficient-distinct-and-sort) discussion, checked JRE source code and such claim still not clear to me.
Also, your performance test is bad and doesn't show anything. Performance testing in java is hard, so you need to use tools like [jmh](https://openjdk.java.net/projects/code-tools/jmh/) on a small part of the code and try different inputs.

Anyway thanks for the advice, I can merge your PR, but I am just curious about reasons for this change, so any analysis is welcome :)

HirotoTanaka114:
@skylot 
I'm sorry for not replying sooner.

I have not enough knowledge for JMH.
So, I can't do the analysis immediately.

PR #865, opened 2020-02-23
chore: add inlined class name as comment

jpstotz:
When classes are inlined it is not that easy to get the original class name, e.g. if you want to compare it in a different decompiler or if you need the class name for a different purpose like an Xposed module or a Frida script.

Therefore I changed the code generator to add a comment line containing the full class name of the inlined class.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
+++ b/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
@@ -223,10 +223,25 @@ public class ClassGen {
 	}
 
 	public void addClassBody(CodeWriter clsCode) throws CodegenException {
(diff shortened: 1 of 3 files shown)


skylot:
@jpstotz please fix [TestAnonymousClass14](https://github.com/skylot/jadx/blob/c7890f2468a14147d80b151dcac17a7875648fea/jadx-core/src/test/java/jadx/tests/integration/inner/TestAnonymousClass14.java#L48) test. It expects that inner class name not appear in decompiled code. Just made a more strict check, like this:
```java
assertThat(code, containsString("use(new Thread(this) {"));
assertThat(code, not(containsString("class AnonymousClass1 extends Thread {")));
```

PR #878, opened 2020-03-06
Fixed adding this. instead of super. when super member is shadowed

acsbendi:
#### Description
The library erroneously used `this.` modifier before a member variable even if the member variable was declared in the super class (for an example, see the [comment](https://github.com/acsbendi/jadx/blob/88e1e7277b9ec7d53749011761b7c89c1795f6d6/jadx-core/src/test/java/jadx/tests/integration/others/TestShadowingSuperMember.java#L14) in the test I added). The modifications I made will put replace `this.` with the correct `super.` keyword in these cases. I found this issue in a real, obfuscated Android APK where the member variable of the super class was shadowed by that of the subclass.

I added a test including smali files. These present the issue quite well, since without the changes I made, the library fails to even create code that compiles.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/Jadx.java
+++ b/jadx-core/src/main/java/jadx/core/Jadx.java
@@ -7,11 +7,11 @@ import java.util.Enumeration;
 import java.util.List;
 import java.util.jar.Manifest;
 
-import jadx.core.dex.visitors.*;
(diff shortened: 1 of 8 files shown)


skylot:
@acsbendi thank you! Nice fix!

PR #880, opened 2020-03-09
Allow to search for full method/field signature

jpstotz:
The search for methods/fields currently only allows to search for the method/field name, but not for the type or the method arguments.

I extended the text index so that the search text is identical to the displayed text (just without the HTML formatting). This allows to search for methods with a specific method signature.

Because the `JMethod` and `JField` `makeLongString()` implementations were returning HTML formatted text I has to introduce the new methods, `makeStringHtml()` and `makeLongStringHtml()` can be used to return a html formatted representation. Therefore the default implementations now only return plain text data. I only needed `makeLongString()` / `makeLongStringHtml()` but for consistency I also applied this scheme to `makeString()` / `makeStringHtml()`.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/treemodel/JMethod.java
+++ b/jadx-gui/src/main/java/jadx/gui/treemodel/JMethod.java
@@ -2,7 +2,8 @@ package jadx.gui.treemodel;
 
 import java.util.Iterator;
 
-import javax.swing.*;
+import javax.swing.Icon;
+import javax.swing.ImageIcon;
 
 import jadx.api.JavaMethod;
 import jadx.api.JavaNode;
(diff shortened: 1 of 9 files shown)


jadx-gui/src/main/java/jadx/gui/treemodel/JMethod.java
```diff
@@ -90,7 +91,11 @@ String makeBaseString() {
 
 	@Override
 	public String makeString() {
-		return UiUtils.typeFormat(makeBaseString(), getReturnType());
+		return UiUtils.typeFormatHtml(makeBaseString(), getReturnType());
```
skylot: Should this be just a `typeFormat` without `Html`?
```suggestion
		return UiUtils.typeFormat(makeBaseString(), getReturnType());
```
And also overriden makeStringHtml() method:
```java
	@Override
	public String makeStringHtml() {
		return UiUtils.typeFormatHtml(makeBaseString(), getReturnType());
	}
```

jpstotz: Looks like I forgot to adapt those methods. Should be addressed in the latest commit.

skylot:
@jpstotz thanks! Nice feature.

PR #888, opened 2020-03-15
chore(gui): preferences dialog changed to two-column mode

jpstotz:
During the last months I had problems with the preferences dialog on my Full-HD monitor on Windows because of it's height:

The button bar at the bottom of the preferences dialog was always shown behind the Windows task bar and therefore inaccessible. I changed the layout to a two-column layout that IMHO fits better to the most monitors.
Additionally the ESC key allows to abort and close the preferences dialog.


skylot:
@jpstotz thanks! Nice improvement!

Issue #890, opened 2020-03-18
Respect JAVA_HOME environment variable in Windows intead of Windows Registry


iyh:
I did set  JAVA_HOME & PATH environment for the right java path (tried both jdk8 and jdk11)
in Windows10 64
```shell
C:\Users\iyh>java -version
openjdk version "11.0.6" 2020-01-14
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.6+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.6+10, mixed mode)
```

still` jadx-gui-1.1.0-no-jre.exe` told me could not find JRE PATH, 
after some digging I managied add fake "JavaHome" 1.8(openjdk11 in fact) path to the Registry "[HKEY_LOCAL_MACHINE\SOFTWARE\JavaSoft\Java Runtime Environment\1.8"  
then works fines.

So how to tell the program use JAVA_HOME environment variable over Windows Registry

iyh:
Maybe it's a launch4j issue
https://github.com/mirror/launch4j/issues/8

jpstotz:
Which java distribution do you use? I use https://adoptopenjdk.net (Java 11) and it's installer has an option to automatically register the `JavaSoft (Oracle) registry keys`. Using this option the Jadx Launch4j exe runner works for me.

iyh:
I confirmed this is a launch4j issue,(also tried running build dev jadx-gui-dev.exe)
 I am using  https://adoptopenjdk.net (Java 11)   as well ,the zip one (OpenJDK11U-jdk_x64_windows_hotspot_11.0.6_10.zip), guess no registry is the problem, wonder why a java program(launch4j)  ingnore  JAVA_HOME environment variable

jpstotz:
Try the installer version, it sets the correct registry keys.

Because the latest Launch4j version is from 2018 where Java on Windows was synonym for Oracle JRE/JDK and those versions always set the registry keys correctly but did not set the JAVA_HOME environment variable. Unfortunately from my perspective it looks like Launch4j project is dead (no commit since last August).

skylot:
Must be fixed in PR #389. 
[AdoptOpenJDK .msi installer](https://adoptopenjdk.net/releases.html?variant=openjdk11&jvmVariant=hotspot) is the easiest way to get java runtime and setup JAVA_HOME environment variable.

PR #895, opened 2020-04-01
Use env %JAVA_HOME% as the JrePath

YunlongYang:
In this way, people can execute jadx-gui-1.1.0-no-jre.exe easily.  
@see Respect JAVA_HOME environment variable in Windows intead of Windows Registry #890


skylot:
Thanks!

Issue #844, opened 2020-02-04
[core] XML attributes contain unprintable bytes


sergey-wowwow:
APK: https://drive.google.com/file/d/1msaMxk32XHruCtEaQDJNk0IGnb7oOWBo/view?usp=sharing

Open ``` AndroidManifest.xml ``` and check attributes e.g. 
```xml
    <application android:theme="@style/" android:label="@string/" android:icon="@mipmap/i_launcher_app" android:name="com.drweb.p000s.App" android:persistent="true" android:allowBackup="false" android:hardwareAccelerated="true" android:supportsRtl="true" android:roundIcon="@mipmap/i_launcher_app_round" android:appCategory="7" android:appComponentFactory="androidx.core.app.CoreComponentFactory">
```

sergey-wowwow:
Seems that those values are parsed in ``` /jadx-core/src/main/java/jadx/core/xmlgen/ResTableParser.java ```
```java
		String[] keyStrings = null;
		if (keyStringsOffset != 0) {
			is.skipToPos(keyStringsOffset, "Expected keyStrings string pool");
			keyStrings = parseStringPool(); // <<<
		}

		PackageChunk pkg = new PackageChunk(id, name, typeStrings, keyStrings);
```
That package is definitely obfuscated, @skylot do you have any ideas
1. how to restore valid values
2. at least where to filter original values? When written in file ``` BinaryXMLParser.java ```, or when they are decoded?

sergey-wowwow:
@skylot 
Can you please suggest something here? I meet this bug pretty often, so I want to fix it asap

jpstotz:
The resources are heavily obfuscated and the key-strings are meaningless - Android just ignores them. Android only uses the resourceIds when loading a resource string, hence the only chance I see here is to check if the resource key name is an obfuscated string like in this case the character 0x1 and then ignore the string and generate a key name from it's resourceId instead.

sergey-wowwow:
@jpstotz 
Yea, that's why I asked for an advise where is the best place to modify code and generate own resource names

jpstotz:
I made some tests and it seems like the best way to handle such situations is to correct the resource names in the resources.arsc reader. 

In detail I modified the end of the method `private static String extractString8(byte[] strArray, int offset)`
https://github.com/skylot/jadx/blob/master/jadx-core/src/main/java/jadx/core/xmlgen/CommonBinaryParser.java#L62:

[...]

PR #909, opened 2020-04-18
feat(res): rename resources keys if contains unprintable chars or duplicates (#844)

skylot:
Implement resources key renaming suggested by @jpstotz in #844


sergey-wowwow:
Thanks, seems to work perfect for two affected apps!

PR #910, opened 2020-04-19
Run indexJob before rename

### LINKED ISSUE #907: Rename can't work correctly after jadx start
I use jadx built from the `rename` branch.
When I use `Rename` instantly after jadx start, only the current name is changed, all the other references of this name remain as before.
After I use  `Find Usage` or `Go to declaration`, `Rename` can work correctly.

bet4it:
The `Rename` function needs `usageInfo` to change all the references of the name changed. We need to make sure `indexJob` has completed before `Rename`.
Fix #907.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/CommonSearchDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/CommonSearchDialog.java
@@ -516,17 +516,15 @@ public abstract class CommonSearchDialog extends JDialog {
 		}
 	}
 
-	protected void loadStartCommon() {
+	private void loadStartCommon() {
 		setCursor(Cursor.getPredefinedCursor(Cursor.WAIT_CURSOR));
 		progressPane.setIndeterminate(true);
 		progressPane.setVisible(true);
-		resultsTable.setEnabled(false);
 		warnLabel.setVisible(false);
 	}
 
 	private void loadFinishedCommon() {
 		setCursor(null);
-		resultsTable.setEnabled(true);
 		progressPane.setVisible(false);
 
 		TextSearchIndex textIndex = cache.getTextIndex();
--- a/jadx-gui/src/main/java/jadx/gui/ui/RenameDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/RenameDialog.java
@@ -1,7 +1,6 @@
 package jadx.gui.ui;
 
 import java.awt.*;
-import java.awt.event.KeyEvent;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -42,7 +41,7 @@ import jadx.gui.ui.codearea.CodeArea;
 import jadx.gui.ui.codearea.CodePanel;
 import jadx.gui.utils.*;
 
-public class RenameDialog extends JDialog {
+public class RenameDialog extends CommonSearchDialog {
 	private static final long serialVersionUID = -3269715644416902410L;
 
 	private static final Logger LOG = LoggerFactory.getLogger(RenameDialog.class);
@@ -55,6 +54,8 @@ public class RenameDialog extends JDialog {
 
 	private CodeArea codeArea;
 
+	private JButton renameBtn;
+
 	public RenameDialog(CodeArea codeArea, JNode node) {
 		super(codeArea.getMainWindow());
 		mainWindow = codeArea.getMainWindow();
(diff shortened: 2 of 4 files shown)


bet4it:
@S-trace

skylot:
@bet4it looks like your change fixes the issue, but I don't think that decompile all classes to just get dependencies is a good idea. It is better to just reload code for all currently decompiled classes and as I remember that was a goal @S-trace tried to achieve.
Anyway, thanks for the help! I will keep this PR open until @S-trace reviews it or I made a fix for that.

S-trace:
I don't think we really need to decompile classes to get usage information.

This should be possible via parsing raw DEX code, just like simple `grep -Flr 'Lsome/FancyClass;'` command in the apktool project gives the list of classes which have the references to the some.FancyClass class.

@skylot What do you think about this approach?

skylot:
@S-trace I agree. Actually, right now I am working on my own dex parsing library which will allow fast traversing through instructions and decode only interesting ones (which contains references to other classes/methods/fields) so we can collect dependencies before decompilation.

S-trace:
@skylot Great! 

How do you think, should I try to implement dependencies collection for current dx-based implementation, or just wait for your new DEX parser? 

We don't need full search index for renaming, we just need the list of classes having references to the class which has the renamed entity (class/method/field), so I think this can be implemented using two `Map<Long, int[][]>` fields somewhere for each DEX. 
Those maps will contain `DexID<<32 + ClassID` as a key, and 2d array of DEX IDs and Class IDs inside each DEX. as a value.

The first Map will contain forward references for each class (which classes do each class reference), and the second Map will contain reverse dependencies for each class (which classes/methods/fields references to this class).

So, the forward/reverse dependencies lookup it will be possible just by Map.get(DexID<<32 + ClassID), and storage will be very compact and fast because it will be based on int arrays, not Objects.

Those maps can be filled using temporary `Map<Long, Set<Pair<Integer, Integer> >` Maps and then (when all DEX files were loaded) those temporary Maps could be flattened to final `Map<Long, int[][]>` Maps.

What do you think about this way to store deps for renaming functionality?

skylot:
@S-trace 

> How do you think, should I try to implement dependencies collection for current dx-based implementation, or just wait for your new DEX parser?

I don't want to implement the same thing twice, so there is no point in doing it right now.

`Map<Long, int[][]>` - is very low level and has no advantage over simple `List<ClassNode>` inside `ClassNode` class like it is already done for ["direct" dependencies](https://github.com/skylot/jadx/blob/e09e933f9ca719c05d647367198878ddc70f02ac/jadx-core/src/main/java/jadx/core/dex/nodes/ClassNode.java#L72). The main issue is that we need to convert from ids to ClassNode node objects every time and it kills all performance gain from using primitive arrays.
Also, I want to move all raw dex info and methods into new lib, so I will remove DexNode class and all low-level dex ids.

> Could you please review it and maybe give me a bit of advice about tracking refs inside methods bodies?

Currently, such tracking implemented in [DependencyCollector](https://github.com/skylot/jadx/blob/f283ef4342fcbeb82c76602a0e23d1a64c225ddd/jadx-core/src/main/java/jadx/core/dex/visitors/DependencyCollector.java#L28) but it works on high-level instruction objects which require to load all method instructions, so now it runs as a visitor in class decompilation.

Also, I want to mention that dependencies info will be used not only for rename feature but to resolve other issues like:
- detect and inline anonymous classes (to be sure it used only in one place)
- classes decompilation order, especially to solve indeterministic results in multi-threaded decompilation
- maybe "find usage" in jadx-gui
- other issues which require class or methods modifications

S-trace:
Ok, so I think this pull request resolves the issue for now.

Attached rebased over https://github.com/skylot/jadx/pull/935 patch from this pull request. 
@bet4it  - could you please apply it to avoid merge conflicts? 
You can use git push --force to update the branch for this pull request - it will update the pull request automatically.

But I have one question about the visual part: 
For now, rename dialog looks really weird for me (screenshot attached).
I think it should not be that high - @bet4it  maybe, you can make it slimmer?

Thank you.

[0001-Run-indexJob-before-rename.patch.txt](https://github.com/skylot/jadx/files/4630536/0001-Run-indexJob-before-rename.patch.txt)
Screenshot_20200514_224517

bet4it:
I didn't change the size of the dialog, jadx uses the size of the dialog at last time.
If you switch to your commit, the dialog should still be so high.
And you can resize this dialog to make it smaller.
Maybe we need to lock the size of dialog so user can't resize it, but I have no idea how to do it.

PR #923, opened 2020-04-29
XML parsing: handling of TYPE_DYNAMIC_REFERENCE entries

### LINKED ISSUE #919: [core] Bug: showing invalid resource names in some specific APK files
**Describe error**
- full name of method or class with error
- full java stacktrace (no need to copy method fallback code (commented pseudocode))
I used the online version, so I don't know what causes it:
http://www.javadecompilers.com/result?currentfile=resources/AndroidManifest.xml

- **IMPORTANT!** attach or provide link to apk file (double check apk version)
Attached here:
[GoogleExtServices.zip](https://github.com/skylot/jadx/files/4551000/GoogleExtServices.zip)

The bug is that I get these:

```
    <application android:label="  ?0x7 131072" android:defaultToDeviceProtectedStorage="true" android:directBootAware="true">
```

```
        <service android:label="  ?0x7 131073" android:name="android.ext.services.notification.Assistant" android:permission="android.permission.BIND_NOTIFICATION_ASSISTANT_SERVICE" android:exported="true">

```

image

The label is invalid.

jpstotz:
A first attempt to fix #919

In the end `TYPE_DYNAMIC_REFERENCE` seem to work identical to `TYPE_REFERENCE`. Not sure if Jadx already resolves those resources and therefore we don't have to make a difference...

The type `TYPE_DYNAMIC_ATTRIBUTE` has been added, but resources of that type can't be processed yet. An example APK would be helpful for this type.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ParserConstants.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ParserConstants.java
@@ -48,6 +48,18 @@ public class ParserConstants {
 	protected static final int TYPE_DIMENSION = 0x05;
 	// The 'data' holds a complex number encoding a fraction of a container.
 	protected static final int TYPE_FRACTION = 0x06;
+
+	/**
+	 * The 'data' holds a dynamic reference, a reference to another resource table entry.
+	 * See https://github.com/skylot/jadx/issues/919
+	 */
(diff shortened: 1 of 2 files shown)


skylot:
@jpstotz thanks!

Issue #931, opened 2020-05-13
[core] Decompilation res, the style have no parent info.


weimings:
Decompilation res, open values/styles.xml, the style have no parent info.
Like 

    <style name="CustomizableHomeTheme" parent="@android:style/Theme.DeviceDefault">

use apktool_2.4.1.jar to decompilation the apk , the parent is complete, use jadx, only have 

    <style name="CustomizableHomeTheme">

I find some drawable by xml is lost.

jpstotz:
Have you already tried the latest unstable version of Jadx? 
Can you please provide the APK file this problem occurs?

weimings:
> Have you already tried the latest unstable version of Jadx?
> Can you please provide the APK file this problem occurs?

I was already pull the the latest unstable version, build it by idea and tried it.
https://apkpure.com/wear-os-by-google-smartwatch-was-android-wear/com.google.android.wearable.app

jpstotz:
Sorry, but at the moment Jadx does not consider the `ResourceEntry.parentRef` data for generating those values.

weimings:
> Sorry, but at the moment Jadx does not consider the `ResourceEntry.parentRef` data for generating those values.

Please explain this problem in the readme, it can reduce misleading users.

jpstotz:
@weimings Sorry but I am too lazy for that. Writing a few lines of code that add generation of parent attribute to Jadx is easier ;)

skylot:
@weimings this issue is fixed in PR #933 by @jpstotz.
Please check the [latest unstable build](https://bintray.com/skylot/jadx/unstable/_latestVersion#files).

weimings:
> @weimings Sorry but I am too lazy for that. Writing a few lines of code that add generation of parent attribute to Jadx is easier ;)

Thanks, I will try this patch.
But I decompilation the ClockworkHomeGoogleLocalEdition.apk（com.google.android.wearable.app ）pull from xiaomi watch, found res/drawable/list_divider.xml was lost.

The apk in this zip, zip pwd is 123.
[com.google.android.wearable.app.zip](https://github.com/skylot/jadx/files/4625621/com.google.android.wearable.app.zip)

jpstotz:
The posted APK has no `drawables` subdirectory unter `/res`, hence Jadx has never and does not display it.

Just open it with a ZIP utility and see it yourself.

weimings:
[...]

PR #933, opened 2020-05-13
chore(core): resource XML generation for parent attribute added

jpstotz:
This PR added the missing functionality identified in #931 

@skylot I am not sure why but I had some serious problems with the `gradlew check -x test` test. Unless I make a clean before executing this the test very often fails because of format violations in the `:spotlessMisc` test for the files `jadx-cli\build\reports\checkstyle\*.xml` (e.g. main.xml). 

I am not sure if the test is designed to also include the build directory (hence generated files), I would have assumed it is only for checking src files. Is it possible to exclude the build directories from code style tests?


skylot:
Thanks for the fix!
For fix checks try to add `**/build/**` to [this line](https://github.com/skylot/jadx/blob/2dce1c0ad920c4cc3f6aaa84cf3f2124dd97ad2d/build.gradle#L102). Hope it helps.

jpstotz:
@skylot adding `"*/build/**"` to the `targetExclude` solves the problem for me, thanks!

skylot:
@jpstotz can you make a PR for that fix?

PR #935, opened 2020-05-14
gui: RenameDialog: Warn user if deobfuscation settings are invalid.

S-trace:
Done.

### Description
    
The rename functionality relies on deobfuscation for now - so let the user know
this and ask the user to enable deobfuscation to get rename working.

The "Force rewrite deobfuscation map file" option effectively disables
renaming, because renaming relies on deobfuscation map modification for now,
but Force rewrite rewrites the map on each file reload, destroying changes.
So. let the user know this issue instead of silent failure.

TODO: Split rename map and deobfuscation map (for example, load the rename map
file after deobfuscation map file to override auto-generated aliases).

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/RenameDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/RenameDialog.java
@@ -31,6 +31,7 @@ import jadx.core.utils.files.InputFile;
 import jadx.gui.jobs.IndexJob;
 import jadx.gui.jobs.RefreshJob;
 import jadx.gui.jobs.UnloadJob;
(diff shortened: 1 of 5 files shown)


skylot:
> TODO: Split rename map and deobfuscation map (for example, load the rename map
file after deobfuscation map file to override auto-generated aliases).

I was planning to save renames in the project file ([this class](https://github.com/skylot/jadx/blob/d1ac43de33aede078592eed864da88540d9c4af1/jadx-gui/src/main/java/jadx/gui/settings/JadxProject.java#L19) serialized to json), so it can be restored after rewrite.

And thanks for the fix!

PR #959, opened 2020-07-03

CirQ:
modifying error of positional argument in Chinese search result text

### Description
The second positional argument is mistakenly set to the third one in Chinese text


skylot:
@CirQ thank you!

PR #968, opened 2020-08-16
Copy preferences values to clipboard

jpstotz:
For issues it is often relevant which preferences values are set in jadx. Currently the most users create a screen shot and paste it into the issue if needed. However having all values as text is often the better solution.

Therefore I added a button that to the preferences dialog that allows to copy the preferences (except for some non-relevant properties) to clipboard in JSON form.


skylot:
@jpstotz thanks!

Issue #980, opened 2020-09-22
Prevent zipbomb forged headers attacks


eric-therond-sonarsource:
The [isZipBomb](https://github.com/skylot/jadx/blob/1560284831aec19433043f9b818fe4daf46ef607/jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java#L56) method relies on [ZipEntry.getSize](https://docs.oracle.com/javase/7/docs/api/java/util/zip/ZipEntry.html#getSize()) to retrieve the size of an uncompressed entry, but getSize is unreliable to protect against zipbomb attacks, indeed, it gets the size from zip file headers which can be forged by an attacker.

For instance, this file [zipBombForgedHeaders.zip](https://github.com/eric-therond/zipbomb/blob/master/src/main/resources/zipBombForgedHeaders.zip) is the same than [zipBomb.zip](https://github.com/eric-therond/zipbomb/blob/master/src/main/resources/zipBomb.zip) except that the headers have been forged (compressed size of an entry = uncompressed size of an entry):

![image](https://user-images.githubusercontent.com/56015232/93878867-0cbd6e80-fcdb-11ea-868f-afcff10a86b2.png)

![image](https://user-images.githubusercontent.com/56015232/93878828-fca58f00-fcda-11ea-9a2e-561b98d13301.png)

The both archives after being uncompressed take several GiB on the disk.

The first zip (ZipBomb.zip) is correctly blocked by jadx but the second not:
`./build/jadx/bin/jadx ./zipBombForgedHeaders.zip -d zipbombtest2`

![image](https://user-images.githubusercontent.com/56015232/93879132-7b9ac780-fcdb-11ea-962b-29432b8edac1.png)

jpstotz:
Do you know how likely ZIP files with manipulated decompressed size values are for non-ZIP-bomb ZIP files?

Because if the uncompressed size value is usually correct (or zero) it would be possible to wrap the `InputStream´ we get from `zipFile.getInputStream(entry)` by a FilterStream that counts the number of bytes read and compares it to the expected decompressed size and a fixed maximum value.

eric-therond-sonarsource:
I guess no legit ZIP files should have forged headers, most unzip tools will likely treat these files as corrupt or unzip will not work.

Exactly, for a more precise solution see:
https://wiki.sei.cmu.edu/confluence/display/java/IDS04-J.+Safely+extract+files+from+ZipInputStream
or
[...]

PR #982, opened 2020-09-22
fix: prevent zipbomb forged headers attacks

skylot:
Fix to prevent zipbomb forged headers attacks. See issue #980.
Input stream for zip entry wrapped by new LimitedInputStream which don't allow to read more than the expected size of entry.
Also, zip entries count is limited to prevent file creation spam.

### DIFF
--- a/jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java
+++ b/jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java
@@ -16,7 +16,7 @@ public class ZipSecurity {
 
 	// size of uncompressed zip entry shouldn't be bigger of compressed in MAX_SIZE_DIFF times
 	private static final int MAX_SIZE_DIFF = 100;
-	private static final int MAX_ENTRIES_COUNT = 10_000;
+	private static final int MAX_ENTRIES_COUNT = 100_000;
 
 	private ZipSecurity() {
 	}
--- a/jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java
+++ b/jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java
@@ -1,8 +1,12 @@
 package jadx.api.plugins.utils;
 
+import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.util.function.BiConsumer;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -12,6 +16,7 @@ public class ZipSecurity {
 
 	// size of uncompressed zip entry shouldn't be bigger of compressed in MAX_SIZE_DIFF times
 	private static final int MAX_SIZE_DIFF = 100;
+	private static final int MAX_ENTRIES_COUNT = 10_000;
 
 	private ZipSecurity() {
 	}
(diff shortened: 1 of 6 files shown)


jpstotz:
I will review your patch, soon. 

From what I have seen the limitation regarding the maximum files in the APK file is set to 10000 which may be a bit low. Especially apps that make use of web technology can contain a large number of files.

I checked by app archive for the apps with more than 10k files, the following app list is just a partial result:

* com.enhanced.skineditorstudio 23k
* com.amazon.dee.app 20k
* com.facebook.games 18k
* com.kakao.talk 15k
* com.microsoft.office.outlook 10k
* com.xiaomi.smarthome 14k
* com.wix.android 12k
* cn.wps.moffice_eng 11k
* com.cyberlink.powerdirector 13k
* com.getmimo 11k
* com.facebook.pages.app 14k
* com.facebook.analytics 18k
* com.tencent.mobileqq 13k
* de.check24.check24 11k

Therefore I would strongly recommend to increase this limit to may be 50k or may be 100k considering that apps usually grow in size (e.g the com.amazon.dee.app grew over last last months from 17k to 20k).

skylot:
@jpstotz thanks! I updated entries limit to 100k.

jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java
```diff
@@ -73,4 +78,28 @@ public static boolean isValidZipEntry(ZipEntry entry) {
...
+		LimitedInputStream limited = new LimitedInputStream(in, entry.getSize());
+		return new BufferedInputStream(limited);
+	}
+
+	public static void visitZipEntries(File file, BiConsumer<ZipEntry, InputStream> visitor) {
+		try (ZipFile zip = new ZipFile(file)) {
+			zip.stream()
+					.filter(entry -> !entry.isDirectory())
+					.filter(ZipSecurity::isValidZipEntry)
+					.limit(MAX_ENTRIES_COUNT)
```
jpstotz, line 93: Using the streams `limit` functionality is very clean, but if we hit the limit it will be not noticeable by the user.
Therefore in this case a manual counter + Exception if the limit is hit may be better.

jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java
```diff
@@ -73,4 +78,28 @@ public static boolean isValidZipEntry(ZipEntry entry) {
...
+					.forEach(entry -> {
+						try (InputStream in = getInputStreamForEntry(zip, entry)) {
+							visitor.accept(entry, in);
+						} catch (Exception e) {
+							throw new RuntimeException("Error process zip entry: " + entry.getName());
+						}
+					});
+		} catch (Exception e) {
+			throw new RuntimeException("Failed to process zip file: " + file.getAbsolutePath());
```
jpstotz, line 102: I made a short test with this branch and it seems that this exception at least in the GUI version is never shown to the user. Therefore a user may open a APK file that triggers the ZipSecurity measures but will never know and in the end work with the APK in Jadx that was not fully loaded.

skylot: You are right, actually, this is a big problem in `jadx-gui`, none of errors shown to the user.
I plan to make some kind of a panel with errors and warnings count if clicked it will open a log window with needed filter.
But showing stacktraces to users is not very good idea, so it is better to also add an additional node in project tree and collect all errors and warnings in a nice report. Anyway, these changes will go in separate commits :)

jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java
skylot: I agree, looks like it is better to rewrite with a simple loop

jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/utils/ZipSecurity.java
skylot: I made a separate issue for this: #986

PR #992, opened 2020-10-17
fix(gui): use correct charset when writing mapping file

coin3x:

### Description
On Windows, the default charset depends on system settings 🤕


skylot:
@coin3x thanks for the fix!

Issue #987, opened 2020-09-30
[feature] Ability to rename package names.


alissonlauffer:
Currenly, Jadx is able to rename classes and field names, but not packages.

## Usage cases:
- With deobfuscation enabled, sometimes Jadx wrongly renames some packages, like `android.support.v4` to `android.support.p000v4`, so we could be able to revert it back to `v4`.
- Sometimes, package names are obfuscated, but its inner classes not. This feature could help to manually rename the obfuscated package name to a more descriptive name.

skylot:
@alissonlauffer I have added rename for packages.
Although if package consist of serveral parts you can rename only one at a time:
![package-rename](https://user-images.githubusercontent.com/118523/94864652-0058b480-0445-11eb-8a79-a0883cc5f0c0.png)
Not sure how to implement it better.

alissonlauffer:
I've been testing it for a while and found a regression: When renaming some packages, the renaming will happen in some of its subpackages instead of the package itself.

Try to rename `android.support.v7` to something else in the following APK: https://drive.google.com/file/d/1RQesxeywrcHQveXEZGFiv9EfIczF2Q3X/view

skylot:
@alissonlauffer I have fixed several issues. Please check :)

alissonlauffer:
Okay, it's now working fine. Thanks!

One last thing, I think would be great if we could rename class files by right clicking their name (just like for packages) instead of opening the file and renaming main class name.

pkilller:
I implemented the rename function。[https://github.com/pkilller/super-jadx/raw/super-jadx/demo.gif](url)

pkilller:
I implemented the rename function
![img](https://github.com/pkilller/super-jadx/raw/super-jadx/demo.gif)

alissonlauffer:
@pkilller I see you created a JADX fork (apparently based on version 1.0.0) with renaming feature, but that feature is already present in dev version and will be released in v1.2.0.

skylot:
Looks like this issue can be closed now

ed9w2in6:
Sorry if this is not the right place for such questions.

I have been trying this feature but I see that this only allows renaming that conforms the existing structure.  That is if a class is located under `a.b.c` I cannot rename it to `a.b.c.d`.

May I ask if there are any existing method that can allow me to do so?

ed9w2in6:
My previous comment can be ignored, sorry for bothering.
I think my issue is described in #1590.
[...]

PR #993, opened 2020-10-20
Rename classes and fields by right clicking their name in left panel.

alissonlauffer:
image

### Description
Adds support for renaming classes and fields by right clicking their name in the left panel. 
As requested in my comment https://github.com/skylot/jadx/issues/987#issuecomment-703030051.


skylot:
@alissonlauffer thanks!

Issue #994, opened 2020-10-25
[core] strings.xml contains duplicated names


sergey-wowwow:
Google Chrome app
![image](https://user-images.githubusercontent.com/15173437/97096930-b6de3c80-1673-11eb-927a-e7896e64a730.png)

Issue #952, opened 2020-06-12
[feature] Detect correct resource type in obfuscated resource folder names


alissonlauffer:
Actually, if an app have its resource folder names obfuscated, Jadx won't try to decode/correct them.

Take a look at https://github.com/alissonlauffer/via-browser-dump/tree/master/src/main: Obfuscated resources are under the `r` folder with obfuscated subfolders, while others (Which are under `resources.arsc`) are under the `res` folder with non-obfuscated subfolders.

Note: If I decompile and rebuild the app with apktool, resource folder names will be decoded and jadx will decompile them fine, but this could be done by Jadx directly.

APK: https://apkpure.com/via-browser-fast-light-geek-best-choice/mark.via.gp/download

liatsineam:
I think this feature is very useful, Jadx should add this feature like apktool.

sergey-wowwow:
@alissonlauffer 
Can you recheck please?

alissonlauffer:
Ok, I've built jadx with the bug_994 branch and tested saving the project as a Gradle project, but resource files are being saved inside `src/main` instead of `src/main/res`:
![image](https://user-images.githubusercontent.com/29029632/97116762-1bce7c80-16de-11eb-88d7-3efc9844898e.png)

sergey-wowwow:
@alissonlauffer 
Updated the PR

sergey-wowwow:
Should be fixed

PR #995, opened 2020-10-25
Fixes duplicates entries in XML resources

sergey-wowwow:

### Description
#994

### DIFF
--- a/jadx-core/src/main/java/jadx/api/ResourceFile.java
+++ b/jadx-core/src/main/java/jadx/api/ResourceFile.java
@@ -4,6 +4,7 @@ import java.io.File;
 
 import jadx.api.plugins.utils.ZipSecurity;
 import jadx.core.xmlgen.ResContainer;
+import jadx.core.xmlgen.entry.ResourceEntry;
 
 public class ResourceFile {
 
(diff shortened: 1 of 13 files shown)


sergey-wowwow:
image

skylot:
@sergey-wowwow, unfortunately, you can't use binary search on that list because it must be sorted before search with the same comparator, and that not true here. I'm not yet sure how to better fix this so I just replace search with plain loop. Please check that it is still working as intended.

skylot:
I just notice that we need to add `config` variable to name comparator. Othewise it renames all entries in other configs even they are not obfuscated. Also this make entries with same id have different names in different configs.
As soon this issue become more complex here some constraint we must met:
- entry names in one config and type must be unique
- entry name for same id must be preserved across different configs 
- notice that different configs can have different entry sets

I tried to fix this PR according to these constatraints, but not sure if it cover all possible cases. Please check and review :)

sergey-wowwow:
> @sergey-wowwow, unfortunately, you can't use binary search on that list because it must be sorted before search with the same comparator, and that not true here. I'm not yet sure how to better fix this so I just replace search with plain loop. Please check that it is still working as intended.

Late night coding issues :D

skylot (approved):

skylot:
@sergey-wowwow great, thanks!
Is this PR complete? Can I merge it?
Also, which issues can be closed after merge?  All these: #994 #996 #952 right?

sergey-wowwow:
@skylot 
Yeah, I think it's complete and ready to be merged. I closed by issues, but not sure if #952 is also done, waiting for confirmation from @alissonlauffer

PR #1001, opened 2020-10-29
Only show renaming option for JClass, JField and JMethod

alissonlauffer:
### Description
My previous PR (#993) erroneously used a null-check for determining if the Rename option should be shown. As result, all nodes had the Rename option shown (but of course it does not work for nodes which are not class/method/field instances).
As this is just a quick fix for an issue in my last PR, I don't think it needs an issue attached (If it does, please tell me).

I'm sorry for pushing broken code to the project. In my future contributions I will try to pay more attention to possible regressions, lol.


skylot:
@alissonlauffer thanks! It is always ok to fix a bug :smile:

skylot (approved):

PR #1027, opened 2020-11-21
Renames invalid res keys

bagipro:
### Description
I've noticed an apk with invalid resource keys. E.g.:

```xml
    <application android:theme="@style/%!" android:label="@string/~{" android:icon="@mipmap/~" android:allowBackup="true" android:supportsRtl="true">
```

```xml
        <PreferenceScreen android:title="@string/!&" android:key="expset" android:summary="@string/~;"/>
```

the last example prevents the XML file from being parsed.


skylot:
@bagipro thank you for the fix!

PR #1035, opened 2020-11-30
res-map.txt updated

jpstotz:
I want to make sure that we also have all resource names from older API levels. Therefore I regenerated the res-map.txt (the number of missing entries was very low). While doing so I noticed small problems ResTableParser.java that prevented me from getting the correct result (renaming of resource names because of not matching the pattern). In the end I added a "raw" mode for building the res-map.txt to avoid future problems.  

fix: do not rename resources names for building res-name.txt
allow loading of resources.arsc from android.jar files
res-map.txt bases on resources.arsc from API 3, 4, 7-30 (taken from https://github.com/Sable/android-platforms)


skylot:
@jpstotz thank you! I totally forgot about renaming :(

PR #1044, opened 2020-12-09
fix: do not remove method start block when it is referenced from dead code

yawkat:
The old code would ignore the entry block for the first pass of dead code removal but would still remove that block if it was referenced by another dead block, because the entry block has no predecessors.

This patch solves the immediate problem, but I'm not convinced by this approach to dead code elimination. I'm not sure it works reliably for complicated dead code structures. Maybe a tracing GC algorithm would be better? Tracing all the blocks that *are* reachable from the entry node instead of trying to determine which blocks are not. This should be possible in O(n).


skylot:
@yawkat thanks for spot such case.

> Maybe a tracing GC algorithm would be better?

Maybe. Current approach very similar, and also such unreachable blocks are very rare, so now we don't allocate any memory if there is nothing to remove. But in terms of code readability and reliable results more general approach is better.

skylot (approved):

PR #1049, opened 2020-12-12
fix: preserve original method details in inlined invocation

yawkat:
`replaceInsn` copies the attributes from the replaced instruction to the new instruction. This included the `METHOD_DETAILS` attribute. For synthetic accessor methods used for accessing private methods of outer classes, this gives an invoke node that calls the outer method but has a `METHOD_DETAILS` attribute for the synthetic accessor method.

Because of this dissonance, overload processing further downstream would use the method details for the synthetic accessor to process overloads of the normal outer class method. The synthetic accessor includes an additional leading argument: the `Outer.this` argument. In overload processing, this argument would be resolved when it shouldn't be, leading to a cast to the outer class type being inserted for the first argument (i.e. the casts appear one position to the right in the argument list of where they should be).

In the test case added in this PR, this bug would manifest as the decompilation `outerMethod((X.this) "");` and `outerMethod((X.this) "", (String) "");` respectively.

With this fix, the `METHOD_DETAILS` are preserved during inlining. This ensures that the invariant that the method detail attribute matches the actual method called by the invoke is maintained.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/InlineMethods.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/InlineMethods.java
@@ -7,6 +7,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import jadx.core.dex.attributes.AFlag;
+import jadx.core.dex.attributes.AType;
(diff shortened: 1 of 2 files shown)


skylot (approved):

skylot:
@yawkat thank you!

PR #1054, opened 2020-12-16
fix: properly transform array creation with constant field length to filled-array

yawkat:
To properly decompile enums, the $VALUES initializer must be a filled-array instruction. To create a filled-array instruction, `ReSugarCode` needs a constant array length. Because of constant inlining, this constant array length can be replaced by an sget if the enum is large (>100 elements). Before this patch, this would break filled-array resugaring and thus enum decompilation.


skylot:
@yawkat thanks!

Issue #1021, opened 2020-11-18
[feature]chat


xmhwws:
I just saw a post where the author introduced his own improved version of JADX, and I was wondering what the jADX author thought of it?
https://bbs.pediy.com/thread-263562.htm
https://github.com/pkilller/super-jadx
![image](https://user-images.githubusercontent.com/42903751/99511386-28837f00-29c3-11eb-9cdc-bfe0cd4a20ae.png)

jpstotz:
@xmhwws I only have contributed very few changes to Jadx, hence I would not call me an author, but in my opinion this is the way open source works. From what I see this "super-jadx" bases on Jadx 1.0.0 and an renaming feature added. Jadx also has a renaming feature now, which one works better, I don't know. As the "super-jadx" project is not a proper Github fork of jadx it seems like the author is not interested in contributing changes back into Jadx which is a valid decision. As the project name has been changed the probability that somebody mixes up both projects is not very high. 

I also did some time ago create a fork + release of Jadx as an experiment for adding a Jimple/[Soot](https://github.com/soot-oss/soot) based output. But that was just a fun project more or less for personal usage, with no real benefit for most of the Jadx users.

xmhwws:
Oh, I'm sorry, there might be a problem with my presentation.
I felt good about some of the features in this enhanced version and wanted to ask the authors of JADX if they would consider integrating them.

The following is from the translation:
![image](https://user-images.githubusercontent.com/42903751/99521419-5b337480-29cf-11eb-9ef6-248d262718b1.png)
What are the shortcomings of the latest JADX 1.2.0?
Today I briefly tried the latest JADX and found the following shortcomings:
1.Renaming variables within a method is not supported.
2.Automatic renaming of override's other methods in parent and subclasses is not supported.(If you rename only 1 method, it will cause a logical error in the code.)
3.The partial Override method is buggy when it is renamed and will not be reflected.
![image](https://user-images.githubusercontent.com/42903751/99521576-8c13a980-29cf-11eb-8793-cb47b13a2e7f.png)
What is super-Jadx?
uper-jadx is my enhanced version based on native JADX and is intended for use by The Android reverse engineer.The following features have been added:
1.Added support for renaming classes, methods, fields, and variables (not supported by native JADx).
2.Rename the override, interface method by association (not supported by native JADx).
3.Duplicate name checks (automatically correlate checks on the override, interface methods of other classes and interfaces).Native JADX is not supported.
[...]

PR #1055, opened 2020-12-18
Renaming causes various concurrency issues

jpstotz:
fix: prevent NullPointerException and ConcurrentModificationException when renaming something

Unfortunately even after this fix renaming still doesn't work for me (tried to rename a method of an Interface). With this fix no error is shown, just the renaming does not happen.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/utils/search/CodeIndex.java
+++ b/jadx-gui/src/main/java/jadx/gui/utils/search/CodeIndex.java
@@ -23,7 +23,7 @@ public class CodeIndex {
 		values.add(value);
 	}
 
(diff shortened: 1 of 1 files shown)


skylot (approved):

skylot:
@jpstotz thanks!
Right now I am working on issue in overridden methods renaming, maybe it is your case. Similar issue was reported in #1021 .

skylot:
@jpstotz in recent commits I fix some issues with the renaming of overridden methods. Can you check your case? If it is not fixed please open a new issue with steps to reproduce.

jpstotz:
@skylot Thanks. Unfortunately method renaming still does not work for me. I have opened issue #1058

Issue #1043, opened 2020-12-06
[core] generated resource names use capital letters


hedwinkoning:
Resource file names are created using a capital "RES_" as a prefix. From what I know capital letters are not allowed for resource file names.

From looking at the code using "RES_" is hardcoded in ./xmlgen/ResTableParser.java

jpstotz:
Android Studio does not complain about resource names with capital letters. Also apps that use such resource names seem to work without problems, at least for those resources I have tested. 
Or may be this was an limitation of older versions of Android resource compiler?

ANR2ME:
I had similar issue using Android Studio 4.1.1 with gradle 6.1.1
```
'R' is not a valid file-based resource name character: File-based resource names must contain only lowercase a-z, 0-9, or underscore
```

MrIkso:
@skylot, I think this issue is not relevant now, it should be closed.

anr2me:
> Hey, I'm using jadx 1.4.0 from homebrew with `--export-gradle` and face a similar issue - I get issues like the identifier "do" or "if" in a strings file (`if is not a valid resource name (reserved Java keyword)`) and names like tG.png `('G' is not a valid file-based resource name character: File-based resource names must contain only lowercase a-z, 0-9, or underscore).`
> 
> Is there any good workaround for this? It seems like the naming rules can't easily be circumvented. Could this be an obfuscation/annoyance technique by the app's vendor?

These are reserved words that's not allowed to be used as identifier: https://stackoverflow.com/questions/53328395/import-is-not-a-valid-resource-name-reserved-java-keyword

If this issue is still happening on the latest version, you probably want to open a new issue, so those words can be blacklisted from the identifier generator.

PR #1057, opened 2020-12-20
fix: use lowercase for resource filename and only use underscore

ANR2ME:
Fix: use lowercase for resource filename and only use underscore for compatibility with newer android studio (#1043)
```
'R' is not a valid file-based resource name character: File-based resource names must contain only lowercase a-z, 0-9, or underscore
```

PS: I don't know whether this commit could cause a problem on older android studio or not tho, since i don't have an old one. 
Anyone willing to test this out?

PPS: Even on newer android studio there will be more resource issues that need to be resolved in order to build the APK, but it seems to be unrelated to resource filename rules (something like: `\src\main\res\values\styles.xml:933:5-946:13: AAPT: error: expected flags but got (raw string) 30.`)

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ResTableParser.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ResTableParser.java
@@ -7,6 +7,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
@@ -27,7 +28,7 @@ import jadx.core.xmlgen.entry.ValuesParser;
 public class ResTableParser extends CommonBinaryParser {
 	private static final Logger LOG = LoggerFactory.getLogger(ResTableParser.class);
 
-	private static final Pattern VALID_RES_KEY_PATTERN = Pattern.compile("\\$+[\\w\\d-_.]+");
+	private static final Pattern VALID_RES_KEY_PATTERN = Pattern.compile("[\\w\\d_]+");
 
 	private static final class PackageChunk {
 		private final int id;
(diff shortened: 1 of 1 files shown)


jadx-core/src/main/java/jadx/core/xmlgen/ResTableParser.java
```diff
@@ -315,7 +316,14 @@ private String getResName(int resRef, String origKeyName) {
...
 		}
-		return "RES_" + resRef; // autogenerate key name
+		// Making sure origKeyName compliant with resource file name rules
+		Matcher m = VALID_RES_KEY_PATTERN.matcher(origKeyName);
+		StringBuilder sb = new StringBuilder();
+		while (m.find()) {
+			sb.append(m.group());
+		}
+		// autogenerate key name, appended with converted origKeyName to be human-friendly
+		return ("res_" + resRef + "_" + sb.toString()).toLowerCase();
```
jpstotz, line 326: Effectively you are cleaning out all disallowed characters from the resource name. In my opinion it is better for readability to have a separator between the different cleaned sections. 

Also keep in mind that on heavily obfuscated apps the cleaned resource name may be empty as all characters were invalid. Therefore I would propose to generate the new resource name this way:

		// Making sure origKeyName compliant with resource file name rules
		Matcher m = VALID_RES_KEY_PATTERN.matcher(origKeyName);
		StringBuilder sb = new StringBuilder();
		boolean first = true;
		while (m.find()) {
			if (!first) {
				sb.append("_");
			}
			sb.append(m.group());
			first = false;
		}
		// autogenerate key name, appended with cleaned origKeyName to be human-friendly
		String newResName = "res_" + resRef;
		String cleanedResName = sb.toString();
		if (!cleanedResName.isEmpty()) {
			newResName += "_" + cleanedResName.toLowerCase();
		}
		return newResName;

ANR2ME: updated, thanks :)

skylot (approved):

jadx-core/src/main/java/jadx/core/xmlgen/ResTableParser.java
ANR2ME: btw, is there any chance for the origKeyName to have non-latin text? (ie. chinese/japanese/korean font)

jadx-core/src/main/java/jadx/core/xmlgen/ResTableParser.java
```diff
@@ -27,7 +28,7 @@
 public class ResTableParser extends CommonBinaryParser {
 	private static final Logger LOG = LoggerFactory.getLogger(ResTableParser.class);
 
-	private static final Pattern VALID_RES_KEY_PATTERN = Pattern.compile("\\$+[\\w\\d-_.]+");
+	private static final Pattern VALID_RES_KEY_PATTERN = Pattern.compile("[\\w\\d_]+");
```
ANR2ME, line 31: Btw, is there any chance for the `origKeyName` to have non-latin text? (ie. chinese/japanese/korean font)
Should i use `a-z` instead of `\w` ?

jpstotz (approved):

skylot:
Merged!
@ANR2ME, @jpstotz thanks!

Issue #454, opened 2019-02-22
setting page can not scroll to bottom


DesLin:
description: setting page can not scroll to bottom
version: 0.9.0
machine: MacBook Pro (Retina, 13-inch, Early 2015)
os: MacOS Mojave 10.14.3 (18D109)
<img width="421" alt="setting" src="https://user-images.githubusercontent.com/3025666/53214416-769d0e80-3687-11e9-8e84-97297da43906.png">

skylot:
@DesLin this issue already fixed in [latest unstable build](https://bintray.com/skylot/jadx/unstable/_latestVersion#files).

jpstotz:
@skylot From a long-term perspective it may be better to introduce multiple tabs here instead of scrolling or do you disagree?

skylot:
@jpstotz I agree, tabs will be good or maybe categories list at left like in idea or [here](https://github.com/dlemmermann/PreferencesFX#preferencesfx), it is hard to decide ))

PR #1059, opened 2020-12-20
Increase scroll speed in settings

alissonlauffer:
### Description
Currently, the scroll speed in settings is very slow (At least in Linux, I couldn't test in other platforms). This PR increases its speed.

Before:

https://user-images.githubusercontent.com/29029632/102722836-9a145d00-42e2-11eb-8fb8-53b247c48726.mp4

After:

https://user-images.githubusercontent.com/29029632/102722826-8832ba00-42e2-11eb-8b0e-20024eae1393.mp4

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/settings/JadxSettingsWindow.java
+++ b/jadx-gui/src/main/java/jadx/gui/settings/JadxSettingsWindow.java
@@ -150,6 +150,7 @@ public class JadxSettingsWindow extends JDialog {
 
 		Container contentPane = getContentPane();
 		JScrollPane scrollPane = new JScrollPane(panel);
+		scrollPane.getVerticalScrollBar().setUnitIncrement(16);
 		scrollPane.setHorizontalScrollBarPolicy(ScrollPaneConstants.HORIZONTAL_SCROLLBAR_AS_NEEDED);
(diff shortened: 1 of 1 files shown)


skylot:
@alissonlauffer thanks!
I hope someday we will implement categories in preferences, like  [here](https://github.com/dlemmermann/PreferencesFX#preferencesfx), discussed in https://github.com/skylot/jadx/issues/454#issuecomment-466529000 :rofl: 

Can anyone confirm that solution works on other platforms?

skylot (approved):

alissonlauffer:
> I hope someday we will implement categories in preferences, like [here](https://github.com/dlemmermann/PreferencesFX#preferencesfx), discussed in [#454 (comment)](https://github.com/skylot/jadx/issues/454#issuecomment-466529000)

Yeah, I agree with the idea of a preference panel just like in IDEA. Using PreferencesFX wouldn't force us to switch GUI from swing to javafx?

skylot:
> Using PreferencesFX wouldn't force us to switch GUI from swing to javafx?

I don't mean switching, I mean just similar appearance ))

jpstotz:
@alissonlauffer Even if the rest of Jadx would still be using Swing we would have to ship JavaFX with Jadx as JavaFx was removed from JRE/JDK.

PR #1061, opened 2020-12-21
fix: do not show empty RenameDialog

jpstotz:
If DeobfuscationForceSave settings is enabled the RenameDialog shows a dialog and asks the user to change this setting. 
If the user denies the change the rename dialog is still shown, but without title and any control.


skylot (approved):

skylot:
@jpstotz thanks! Nice fix!

Issue #954, opened 2020-06-17
[feature] Multi keyword search


dubey-amit:
*Describe your idea:*
Does JADX-GUI have an option to search multiple keywords at once? 
Eg    Debug|root|users

MrIkso:
@skylot cloese it, regex search supported

PR #1069, opened 2020-12-27
Implements the option to use Regex on the Search Dialog

green9317:
This will implement the option to use Regex when searching with the text search dialog.  

Addresses issue #954


jadx-gui/src/main/java/jadx/gui/utils/search/CodeIndex.java
```diff
@@ -27,15 +29,35 @@ public synchronized void removeForCls(JavaClass cls) {
 		values.removeIf(v -> v.getJavaNode().getTopParentClass().equals(cls));
 	}
 
-	private boolean isMatched(StringRef key, String str, boolean caseInsensitive) {
-		return key.indexOf(str, caseInsensitive) != -1;
+	private boolean isMatched(StringRef key, String str, boolean caseInsensitive, boolean useRegex) {
+		if (caseInsensitive && useRegex) {
+			try {
+				Pattern pattern = Pattern.compile(str, Pattern.CASE_INSENSITIVE);
```
jpstotz, line 35: `isMatched` is called for each `CodeNode`, hence it is called several thousand times . For efficiency it would be good if you could manage to compile the regex pattern only once for a complete search. 

May be it is time to switch from search method(s) that get the search options as parameters to a search predicate that is created once and then applied to all items to be searched?

green9317: Thanks for the quick response.  That is a good point about the efficiency issues with compiling in the isMatched function.  Ill make some changes to fix this.

green9317: I just created a new class SearchSettings to store all of the options provided by the user and pass it as the input to the search methods.  I also created a SearchImpl class to check if a match was found.  I think this is a bit cleaner and will enable the Regex to be compiled prior to running the search.  It should also make it easier to add new options in the future should they be needed.  

Its worth pointing out that the SearchEngine API used for highlighting appears to compile Regex at search time as well so there might still be a hit to efficiency.  It may be possible extend the new search classes i made to perform the highlighting function which will enable the regex to be compiled ahead of time.  Let me know if that is a concern and I can look into updating it.

jpstotz: Thanks. Your implementation is not exactly what I had in mind but the major point regarding the regex compilation is solved.

jadx-gui/src/main/java/jadx/gui/utils/search/SearchImpl.java
```diff
@@ -0,0 +1,54 @@
...
+	 */
+	public static boolean isMatch(StringRef searchArea, final SearchSettings searchSettings) {
+		return isMatch(searchArea.toString(), searchSettings);
+	}
+
+	/*
+	 * Checks if searchArea matches the searched string found in searchSettings
+	 */
+	public static boolean isMatch(String searchArea, final SearchSettings searchSettings) {
+		return find(searchArea, searchSettings) == -1 ? false : true;
```
jpstotz, line 20: This code is very likely to be misinterpreted by a developer. Why not clean and simple 

`return find(searchArea, searchSettings) != -1;`

jadx-gui/src/main/java/jadx/gui/utils/search/SearchSettings.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.regex.Pattern;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SearchSettings {
+
+	private static final Logger LOG = LoggerFactory.getLogger(SearchSettings.class);
+
+	private String searchString;
```
jpstotz, line 12: For efficiency I would recommend to make the fields final and provide for each use case an own constructor. As final fields can't change the Java code optimizer can generate faster code as several if branches can be omitted.

jadx-gui/src/main/java/jadx/gui/utils/search/SearchImpl.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+public class SearchImpl {
+
+	/*
+	 * Checks if searchArea matches the searched string found in searchSettings
+	 */
+	public static boolean isMatch(StringRef searchArea, final SearchSettings searchSettings) {
+		return isMatch(searchArea.toString(), searchSettings);
```
jpstotz, line 13: All your  `isMatch` and `find` implementations are static and have an `SearchSettings searchSettings` argument. 
So effectively they are handled like a virtual method of `SearchSettings`. I assume you wanted make a split between the data model and the implementation. Was this the reason or are there other reasons for doing so? 

Because moving the code into `SearchSettings` would in the end allow to use `SearchSettings` as interface and provide for each major search type an own optimized implementation.

jadx-gui/src/main/java/jadx/gui/utils/search/SearchImpl.java
green9317: Thanks not sure what I was thinking

jadx-gui/src/main/java/jadx/gui/utils/search/SearchSettings.java
green9317: I just implemented this for the searchString, ignoreCase, and useRegex option.  It should work with a single constructor as all 3 options should always have a value.

jadx-gui/src/main/java/jadx/gui/utils/search/SearchImpl.java
green9317: Your correct in that my goal was to separate the data model from the implementation with the idea to build a settings class to store all options which can be expanded in the future if needed.  Originally, I was thinking about just passing the settings class into the existing isMatched functions and doing the comparison there as was originally the case.  Although, it appeared that codeIndex, simpleIndex, and the searchNext function all implement the same algorithm with minor changes so I just moved that into the searchImpl class rather than updating it in all 3 places.

I just submitted another commit combining SearchImpl and SearchSettings as you suggested and agree that it is cleaner than my original implementation.  I still need to implement a search interface as recommended so will be making another commit.

What did you have in mind with doing an optimized implementation for each search type? Are you recommending to create a search interface then implement it in the index classes rather than using a generic find function implemented in the searchSettings class which is what I currently have.

skylot (approved):

jpstotz (approved):

Issue #1066, opened 2020-12-25
[code] can't decompile spring boot jar


lizhibin7530:
I download from the Latest release and decompiler a springboot project jar, the source code is blank. the log is :

```java
ERROR: Failed to convert file: C:\Users\lzb\Desktop\excutor\xxl-job-executor.jar
java.lang.RuntimeException: Java to dex conversion error, code: 1
 errors: 
PARSE ERROR:
com.android.dx.cf.iface.ParseException: class name (com/xxl/job/executor/XxlJobExecutorApplication) does not match path (BOOT-INF/classes/com/xxl/job/executor/XxlJobExecutorApplication.class)
	at com.android.dx.cf.direct.DirectClassFile.parse0(DirectClassFile.java:557)
	at com.android.dx.cf.direct.DirectClassFile.parse(DirectClassFile.java:434)
	at com.android.dx.cf.direct.DirectClassFile.parseToInterfacesIfNecessary(DirectClassFile.java:416)
	at com.android.dx.cf.direct.DirectClassFile.getMagic(DirectClassFile.java:256)
	at com.android.dx.command.dexer.Main.parseClass(Main.java:773)
	at com.android.dx.command.dexer.Main.access$1600(Main.java:85)
	at com.android.dx.command.dexer.Main$ClassParserTask.call(Main.java:1766)
	at com.android.dx.command.dexer.Main.processClass(Main.java:756)
	at com.android.dx.command.dexer.Main.processFileBytes(Main.java:724)
	at com.android.dx.command.dexer.Main.access$1200(Main.java:85)
	at com.android.dx.command.dexer.Main$FileBytesConsumer.processFileBytes(Main.java:1719)
	at com.android.dx.cf.direct.ClassPathOpener.processArchive(ClassPathOpener.java:285)
	at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:165)
	at com.android.dx.cf.direct.ClassPathOpener.process(ClassPathOpener.java:143)
	at com.android.dx.command.dexer.Main.processOne(Main.java:678)
	at com.android.dx.command.dexer.Main.processAllFiles(Main.java:575)
	at com.android.dx.command.dexer.Main.runMultiDex(Main.java:365)
	at com.android.dx.command.dexer.Main.runDx(Main.java:286)
	at jadx.plugins.input.javaconvert.DxConverter.run(DxConverter.java:40)
[...]

PR #1070, opened 2020-12-28
Replace dx with d8 for dex code generation

jpstotz:
dx library has been replaced with d8 for dex code generation.

Additionally support for embedded jar files within jar files has been added.

d8 seem to be much more relaxed when it comes to class files not in the main class-path like it is the case in #1066. 
Together with the decompilation of embedded jar files this should fix issue #1066.

I had to set the `setMinApiLevel(28)` for d8, otherwise dex generation failed as we have to set `setDisableDesugaring(true)`. 
With desugaring code generation would be easier, but if I understand d8 correctly this would require not only the Jar file(s) to be converted to be present but also all of their dependencies.


jpstotz:
Looks like d8 generated totally different code for certain code structures. The reason for the build to fail is that d8 does convert switch - case blocks with an enum as switch value to an anonymous inner class that uses a `int[] ` array as lookup table:

dx based version:

```
    public String testEnumSwitch(Direction direction) {
        switch (direction) {
            case NORTH:
                return "N";
            case SOUTH:
                return "S";
            default:
                return "<>";
        }
    }
```

d8 based version:
```
    /* renamed from: jadx.samples.TestEnum$1  reason: invalid class name */
    static /* synthetic */ class AnonymousClass1 {
        static final /* synthetic */ int[] $SwitchMap$jadx$samples$TestEnum$Direction;

        static {
            int[] iArr = new int[Direction.values().length];
            $SwitchMap$jadx$samples$TestEnum$Direction = iArr;
            try {
                iArr[Direction.NORTH.ordinal()] = 1;
            } catch (NoSuchFieldError e) {
            }
            try {
                $SwitchMap$jadx$samples$TestEnum$Direction[Direction.SOUTH.ordinal()] = 2;
            } catch (NoSuchFieldError e2) {
            }
        }
    }
    public String testEnumSwitch(Direction direction) {
        int i = AnonymousClass1.$SwitchMap$jadx$samples$TestEnum$Direction[direction.ordinal()];
        return i != 1 ? i != 2 ? "<>" : "S" : "N";
    }
```

I am not sure if Jadx currently supports decompiling such dex code as the generated code fails to compile in the end (this is where the build process fails). 

@skylot One thing I noticed is that the latest version of Jadx in master seems to deobfuscate anonymous inner classes, may be because of their name like `$1`? I am not sure if this is a wanted behavior or a side effect of the latest patches.

skylot:
@jpstotz I already tried d8 as java converter but removed it later (cd8307f432947f8ac6d4b621134b41f294314bc5). The main reason was a lot of code optimizations and transformations that hard to revert. For example sometimes `switch` replaced with `if/else` pattern. This is not a big deal because semantically code is equivalent, but it is broke current tests and it is harder to compare with original code. I agree `d8` is better than `dx` because `dx` is unsupported for a long time so it is expected that it can have issues with the latest java versions.
To overcome these problems I made a common layer in dex reader, so it is possible now to write direct input plugin for java bytecode. Actually, I already start implementing this, but it is not finished yet.

> deobfuscate anonymous inner classes, may be because of their name like $1?

Yes, this is hardcoded here: https://github.com/skylot/jadx/blob/aa8fd3c861448102019d97e20e99d96fcd8d12f0/jadx-core/src/main/java/jadx/core/dex/visitors/ProcessAnonymous.java#L45

PR #1082, opened 2021-01-15
fix: elemSize=0 fill_array_data_payload insn obfuscation

AlienwareHe:
### fix case
the case is at the end of method code item,Insert some fill_array_data_payload instructions with elemSize of 0 at the end of the method.

such as:

image

The left is before optimization, the right is after optimization

and the smali is:
image


skylot (approved):

skylot:
@AlienwareHe thanks for the fix!

Issue #1052, opened 2020-12-15
[GUI] Can not select "Find reference" in context menu after I upgrade to macOS 11.1


shatyuka:
**Describe error**
As described in title, the GOTO and Find reference are both disabled, seems like I didn't select anything.

shatyuka:
Currently find a way out: use [universalJavaApplicationStub](https://github.com/tofi86/universalJavaApplicationStub) to pack the jar to an app.

shatyuka:
Seems that [`codeArea.getMousePosition()`](https://github.com/skylot/jadx/blob/7f8d03d1927b95565938a4e2089f4cbed3417979/jadx-gui/src/main/java/jadx/gui/ui/codearea/JNodeMenuAction.java#L39) failed to retrieve the position.
Now I have an alternative method. #1086

skylot:
Fixed in PR #1086

PR #1086, opened 2021-01-17
fix(gui): codearea popup menu always disabled in macos (#1052)

shatyuka:
### Description
Can not select "Find Usage" and "Go to declaration" in context menu after I upgrade to macOS 11.1
I also tried many versions of java, but none of them works.
Then I look into the source code. Seems that [`codeArea.getMousePosition()`](https://github.com/skylot/jadx/blob/7f8d03d1927b95565938a4e2089f4cbed3417979/jadx-gui/src/main/java/jadx/gui/ui/codearea/JNodeMenuAction.java#L39) failed to retrieve the position.

Related issue: #1052


jpstotz:
I performed a quick test of the changed code on Windows and it seems to have no negative impact. Therefore I approve this PR.

skylot (approved):

skylot:
Merged. @shatyuka thank you for the fix.

Issue #1032, opened 2020-11-25
[GUI]CommonSearchDialog::openSelectedItem closes search dialog once item selected.


4B5F5F4B:
Hi,

JADX is my favorite APK reverse engineering tool but one thing annoys me a lot. Whenever I selects code snippet in search dialog and jumps to the code, search dialog will close. So I have to open search dialog again if I want to continue my searching.

I notice there's a dispose function called in CommonSearchDialog::openSelectedItem

```
protected void openSelectedItem() 
{
		int selectedId = resultsTable.getSelectedRow();
		if (selectedId == -1) {
			return;
		}
		JNode node = (JNode) resultsModel.getValueAt(selectedId, 0);
		tabbedPane.codeJump(new JumpPosition(node.getRootClass(), node.getLine()));

		dispose();
}
```
I tries to patch the code like this

```
protected void openSelectedItem() 
{
		int selectedId = resultsTable.getSelectedRow();
		if (selectedId == -1) {
			return;
		}
		JNode node = (JNode) resultsModel.getValueAt(selectedId, 0);
		tabbedPane.codeJump(new JumpPosition(node.getRootClass(), node.getLine()));

		//dispose();
}
```
Then search dialog will not close any longer:)

jpstotz:
I also wanted to implement something similar but until now I have not managed to do so. 
I think the main problem is that just disabling is in my opinion not a good option. The search dialog is a "multi-purpose dialog used several times within Jadx and not always it is the best solution to keep it open. 

I had something in mind where there is a checkbox in the search dialog to "keep it open". What do you think of such a solution?

skylot:
Added in PR #1108

PR #1108, opened 2021-01-30
enable to search resource files like .xml/.html, etc...

LBJ-the-GOAT:
#1032 

Hi, I've always loved the search feature provided by Jadx It's so fast so helpful, if support to search resource files like .xml or .html it would be even more powerful and people don't have to leave Jadx to other IDEs for such feature, so I opened this PR.

Hopefully it could help.

search_resource_present2


skylot (approved):

skylot:
@LBJ-the-GOAT thanks! Great work!

Issue #1114, opened 2021-02-11
[feature] Bytes-Column for Smali Code


OdinVex:
I think it would be quite handy to show the bytes involved per instruction line of Smali code.

jpstotz:
Unfortunately I don't think this is currently possible. Jadx uses [dexlib2](https://github.com/JesusFreke/smali) for generating smali code and this library seem to allow only smali code generation at class level. Therefore we can not generate a mapping between instructions and the raw bytes that belong to the instruction.

OdinVex:
I've submitted a feature-request to dexlib2 for an option. Hopefully this can be incorporated.

jpstotz:
I just checked and to be precise it is the baksmali part of that library not dexlib2.

OdinVex:
I received a [response](https://github.com/skylot/jadx/issues/1114#issuecomment-777294174) from the author of baksmali.

jpstotz:
OdinVex forgot to mention that baksmali is used as a library and on library level there seems no way to generate the smali code in smaller parts that on class level for adding some sort of mapping between smali code and the offset in the dex file.

skylot:
Well, I checked smali dump command, looks like it dumps a whole dex file and not only one class. So we can't use this directly.
Another option: I can make my own smali printer, so we can add any feature into it. This is not hard but needs a lot of time, and I am not sure how many people need this. 
@OdinVex can you tell why you need this feature and how you will use it? I just wonder what kind of work require seeing raw bytes of instructions :smile:

OdinVex:
> Well, I checked smali dump command, looks like it dumps a whole dex file and not only one class. So we can't use this directly.
> Another option: I can make my own smali printer, so we can add any feature into it. This is not hard but needs a lot of time, and I am not sure how many people need this.
> @OdinVex can you tell why you need this feature and how you will use it? I just wonder what kind of work require seeing raw bytes of instructions smile

I've already begun local work on a git clone. There are a lot of Android ROMs out there with drivers and software with no public source code available for _whatever_ reason. Porting Linux to certain devices needing certain drivers goes a lot easier with a decent decompiler. Knowing offsets allows manipulation of these libs to better modify their behavior such as rewriting fixed-length arrays with dynamic ones so that more cores (which were statically limited) can (or in some cases _must_) be used.

skylot:
@OdinVex I see, so you are doing binary patching.
But I thought for dex it is easier editing smali and build it back rather than hardcore binary fixing.
[...]

PR #1126, opened 2021-03-02
improve smali printer to show bytecode

LBJ-the-GOAT:
#1114 
All the instruction decoding jobs had been done by @skylot since don't know when, I just put them together to print smali code.

added 3 columns to the output, 

- first column is a file offset within the dex file
- the second is bytecode
- the third one is the code offset within the method.

and added an addtional font setting it contains fixed-width fonts, for alignment of output. 

smali_bytecode_present


Surendrajat (commented):
Nice job @LBJ-the-GOAT. I've a few comments:
- I think `ByteCode` has a very different meaning in Java world. Here simply using `Bytes` or `OpCode Bytes` is more relevent.
- Does the last comment ie. field/method offset (eg. `method@0123`) represent absolute address in dex? Or related to something? I may be mistaken about something here :)
- How much work might it require for synchronizing smali and java methods? Like scrolling in some other method moves cursor in related java code too. I'm sure it won't work for innerclasses right now. Just a thought.

LBJ-the-GOAT:
> Nice job @LBJ-the-GOAT. I've a few comments:
> 
> * I think `ByteCode` has a very different meaning in Java world. Here simply using `Bytes` or `OpCode Bytes` is more relevent.
> * Does the last comment ie. field/method offset (eg. `method@0123`) represent absolute address in dex? Or related to something? I may be mistaken about something here :)
> * How much work might it require for synchronizing smali and java methods? Like scrolling in some other method moves cursor in related java code too. I'm sure it won't work for innerclasses right now. Just a thought.

> Okay, I think Btys is better.
> it's an offset to the start of the code item or something, I added it because the output of dexdump has it, more importantly it looks pretty pro, haha...

> It's a good idea, It might take a while, I just simply print the result now, it needs attach line info to do it.

skylot (approved):

Surendrajat:
> Okay, I think Btys is better.

Cool. This will avoid the confusion.
> more importantly it looks pretty pro, haha...

Yeah. Then it needs to go in :rofl:
> I just simply print the result now, it needs attach line info to do it.

I see. Cool. It'll be a nice feature TBH. Till now Smali pane is not so useful for me.

skylot (approved):

Surendrajat:
Seems like we're going to have `Bytecodes` for now.

skylot:
> Seems like we're going to have Bytecodes for now.

Oops! Did I merged too early? Sorry :disappointed:

Surendrajat:
Since @LBJ-the-GOAT agreed for `Bytes`, I was thinking so. But it can be done in a follow up PR too.

LBJ-the-GOAT:
> > Seems like we're going to have Bytecodes for now.
> 
> Oops! Did I merged too early? Sorry 😞

@Surendrajat  Haha.. I was typing and then I saw it merged, so just continued to catch up some TV shows

PR #1127, opened 2021-03-02
Code comments

### LINKED ISSUE #359: Comments feature
It could be useful the possibility to add inline comments to the decompiled code, to make easier the reversing process.

skylot:
Add comments feature to jadx-gui (fixes #359)

- All comments stored in project file. Make sure you save it before exit (or enable autosave in preferences).
- Multi-line comments supported
- Comments attached by raw instructions offset, so it is possible to see them in fallback mode.
- Search for comments also supported (global or in current file)


Surendrajat:
@skylot someone comments search in active file is always empty for me. Am I doing something weird? :)

skylot:
@Surendrajat thanks! Looks like issue reproduces for inner classes. With warning in log:
```
WARN : Failed to get ref node for comment: JadxCodeComment{...
```

Surendrajat:
@skylot seems like I panicked without realizing I was only testing inner-classes ;P

Surendrajat (approved):
Last commit fixed the remaining issue, so I'm happy with this feature. Thanks again!

LBJ-the-GOAT:
Get rid of the "offset" field in the CodePosition is so exciting, it makes rename without reloading classes and update index much more easier. 

I think it's better not use CodePosition as a map key, so we can just update its relevant position after rename no need to reindex anymore.

Since CodePosition is absolutely sorted, I think use a list and binary search is enough for efficiency.

skylot:
@LBJ-the-GOAT you are right, but only partially.
`CodePosition` now store duplicate information and can be replaced with "text offset from start" (soo many `offset` in the code :rofl:)

But I don't think it is a good idea to use inplace renaming, because it is not general enough, and as a result we will need to implement 2 ways of applying rename or comments or any other code modifications.
Also, now text index holds prepared lines objects and just show them in search results without creating over an over again. Not sure how can affect removing of such cache.

LBJ-the-GOAT:
@skylot  resource search doesn't has a cache it's not slow. We only want a match in a line and it's in-memory, so I think a full document search shouldn't be a big problem.
We can try to make the text and usage search always refer to the code text and positions of `ICodeInfo` and `CodePosition`, by doing so we can just take care this two classes and forget about the caches and indexes 🤗🤗🤗 Isn't it greate?

skylot:
@LBJ-the-GOAT sure, code index not really needed, but it slightly speeds up search results generation by caching them. So we need to make some PoC to check how such changes will affect performance and memory usage.
Anyway, I plan to refactor search to lazy pageable model, because now, in most cases, more than a hundred results not really needed, but search spent time and memory to prepare them.

And for this PR: I fixed some small search issues, and not found any new one, so I think it is ready for merge.

LBJ-the-GOAT:
@skylot Great plan, I once tried to change search featrue to multithreaded, but failed.
btw I am doing a smali debugger, by now it's almost 8k lines, would you accept a PR with such many lines but no test files for it? just for curiosity  :)

skylot:
> I am doing a smali debugger, by now it's almost 8k lines, would you accept a PR with such many lines but no test files for it?

Well, tests are a good thing, but now jadx-gui don't have them, so I think it is ok :rofl:

jpstotz:
In a quick test of the new code comment feature I wasn't able to use the key combinations proposed by the context menu. May be because there is no `;` key on a German keyboard layout (I have to press Shift + `,` to get a `;`). I assume this will be also true for other languages and will raise issues.

skylot:
@jpstotz hm, I thought, these keystroke bindings are kinda language independent, and now I have no idea how to fix this. I will try to search for a solution anyway. And thanks for share this issue!

PR #1140, opened 2021-03-31
Update Messages_ko_KR.properties

choiman1559:
Updated korean translation

PS: How can I connect to adb inside jadx?
I tried with the tcpip address and port of my device but it keeps failing.


LBJ-the-GOAT:
> PS: How can I connect to adb inside jadx?
> I tried with the tcpip address and port of my device but it keeps failing.

You can click the `Start ADB Server` button to start an adb server on local, and then click `Refresh`,
the adb of jadx works like this: ` Jadx ---> local adb ----> android device `.

Tcp ip is for if you have two computers you can connect to another one's adb server, but I haven't test this function yet. :-)

choiman1559:
> > PS: How can I connect to adb inside jadx?
> > I tried with the tcpip address and port of my device but it keeps failing.
> 
> You can click the `Start ADB Server` button to start an adb server on local, and then click `Refresh`,
> the adb of jadx works like this: `Jadx ---> local adb ----> android device`.
> 
> Tcp ip is for if you have two computers you can connect to another one's adb server, but I haven't test this function yet. :-)

Still, adb inside jadx does not run.

```
$ adb devices
List of devices attached
LGMX320SWKJJ9DFUJV	device
1f55bc1a	device
```

As you can see, even though I connected my devices via usb and tried ```Start ADB Server``` after killing the original adb's server, I only got an error message saying ```process hasn't exited```.

LBJ-the-GOAT:
@choiman1559  Cam you set the `ADB Addr` and `ADB Port` to localhost and 5037, and try again? Bacuae you said you had tried a tcp ip, so may be  you forgot to change it back. Otherwise it's so weird.😳

jadx-gui/src/main/resources/i18n/Messages_ko_KR.properties
```diff
@@ -221,51 +221,52 @@ apkSignature.warnings=경고
...
-#debugger.run=Run (F9)
-#debugger.stop=Stop debugger and kill app
-#debugger.pause=Pause
-#debugger.rerun=Rerun
-#debugger.cfm_dialog_title=Exit while debugging
-#debugger.cfm_dialog_msg=Are you sure to terminate debugger?
+debugger.process_selector=디버깅 할 프로세스 선택
+debugger.step_into=한 단계씩 코드 실행 (F7)
+debugger.step_over=프로시저 단위 실행 (F8)
+debugger.step_out=프로시저 나가기 (Shift + F8)
```
skylot, line 228:
```suggestion
```
Please don't insert empty lines or change lines order

skylot (changes requested):
Please don't insert empty lines.
We have automated tests to check and simplify translations editing, so all files must have same keys on same lines.

PR #1143, opened 2021-04-03
Small search dialog optimizations

jpstotz:
When opening the search dialog for the first time the layout was a bit strange, it seems the WrapLayout was reserving space for placing the two search option boxes stacked vertically, but later redecided and placed them horizontally but keeping the vertical space. 
This was may be related to the fact that the target window size was restored after creating all components and additionally an fixed intermediate window size was set. Seems that this confused the Java layouting engines.

Additionally I modified the search dialog to show a red background color in case an invalid regex was entered.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/SearchDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/SearchDialog.java
@@ -95,10 +95,10 @@ public class SearchDialog extends CommonSearchDialog {
 		this.options = buildOptions(preset);
 		this.options.addAll(additionalOptions);
 
+		loadWindowPos();
 		initUI();
 		searchFieldSubscribe();
 		registerInitOnOpen();
-		loadWindowPos();
 		registerActiveTabListener();
 	}
 
@@ -246,8 +246,6 @@ public class SearchDialog extends CommonSearchDialog {
 		});
 
 		setTitle(NLS.str("menu.text_search"));
-		pack();
-		setSize(800, 500);
 		setLocationRelativeTo(null);
 		setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
 		setModalityType(ModalityType.MODELESS);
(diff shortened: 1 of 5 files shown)


jadx-gui/src/main/java/jadx/gui/ui/SearchDialog.java
```diff
@@ -246,8 +248,6 @@ public void keyReleased(KeyEvent e) {
 		});
 
 		setTitle(NLS.str("menu.text_search"));
-		pack();
-		setSize(800, 500);
```
skylot, line 250: Hm, if you don't call `pack()` or `setSize` first run will show zero size window.
So I suggest to check if window pos not applied and call `pack()`: 
```java
if (!mainWindow.getSettings().loadWindowPos(this)) {
		pack();
}
```
Also, window pos is ignored for some cases: like moving window outside of screen bounds (check `JadxSettings#loadWindowPos()` code and PR #521).

skylot (changes requested):

jadx-gui/src/main/java/jadx/gui/ui/SearchDialog.java
jpstotz: You are correct, I did not consider the case when Jadx is started the first time. I will make some more tests. 

Unfortunately your proposed changes doesn't fix all problems. I assume that there is a bug in the calculation of WrapLayout. I will check that tomorrow.

jpstotz: It took a little bit longer to check the WrapLayout. I updated the code so the dialog size is now set if the dialog position can not be restored from settings.

jpstotz:
The major change is in WrapLayout the change of the comparison:

* old: `rowWidth + width > maxWidth`
* new: `rowWidth + width >= maxWidth`

Based on my tests the greater-euqal comparison is required to make the WrapLayout work correctly.

jpstotz:
I am sorry - sometimes I really don't know what the hell git respectively IntelliJ is doing. I never configured this repo as remote for that branch. 
Now the the push was done to the correct repo.

skylot (approved):

skylot:
@jpstotz thanks!

Issue #1151, opened 2021-04-15
[feature] Enable to exclude multiple packages at the same time


LBJ-the-GOAT:
Yesterday, I decompiled an APK, it was huge, very huge that Jadx took 5 minutes to do the indexing, and then I was thinking a wise man like me would exclude some packages that doesn't matter to the  project, so I excluded a package... Jadx immediately reopened the APK and took another 5 minutes to index ... 

I think that to support to exclude many packages at once instead of exclude them one by one is 
quite necessary especially when it would take a long time to reindex, so I opened this issue for a PR.:)

skylot:
> support to exclude many packages at once

@LBJ-the-GOAT you can do this in preferences using `Excluded packages` option. This input allow to enter several packages separated by space. Is this not enough?

LBJ-the-GOAT:
@skylot  It needs typing though and it's one line, I perfer to use a tree to select those packages like this in the picture. What do you think?

![16185082245321194241864328636709](https://user-images.githubusercontent.com/66319139/114914008-6a223400-9e54-11eb-9162-1b0e3a7b5107.jpg)

skylot:
Oh, I see. Looks great, sure it is much more usable than text input :+1:

LBJ-the-GOAT:
@skylot It needs typing though and it's one line, I perfer to use a tree to select those packages like this in the picture. What do you think?

![Screenshot_2021-04-16-01-43-13-20](https://user-images.githubusercontent.com/66319139/114914647-27149080-9e55-11eb-91f6-189eb15d3c5c.jpg)

skylot:
Merged

PR #1152, opened 2021-04-16
Improve exclude package feature

LBJ-the-GOAT:
#1151 

Add a tree to select packages for exclude them at the same time, this gives people an opportunity to just analyze the only package they want in an APK, because now we can exclude packages very easily. ✌

master_exclude_package_present


skylot (approved):

skylot:
@LBJ-the-GOAT thanks!

PR #1160, opened 2021-04-25
fix a minor bug in while loop

LBJ-the-GOAT:
Recently, I found a mistake that I had made in all my code.
I had always thought that the `continue` (in code below) will go to the `do` statement and start a new round, but nope, according to the official Java documentation
> The continue statement ... skips to the end of the innermost loop's body and evaluates the boolean expression that controls the loop.

So turns out the `continue` goes to the `while (false)`  instead of  `do`. 
```java
do {
 if (loop())
  continue;
} while (false);
```

I am very sorry for wrote such code to Jadx.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/ADBDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/ADBDialog.java
@@ -624,7 +624,8 @@ public class ADBDialog extends JDialog implements ADB.DeviceStateListener, ADB.J
 						}
 					}
 					resultDesc = rst.desc;
-				} while (false);
+					break;
+				} while (true);
 			} catch (IOException e) {
 				e.printStackTrace();
 			}


skylot (approved):

skylot:
So previously next port retry was broken and this patch fix that. Good :+1:

LBJ-the-GOAT:
> So previously next port retry was broken and this patch fix that. Good 👍

Yes 👍

skylot:
> fixed alerts:
   1 for Continue statement that does not continue

Looks like next time I will force to fix all LGTM alerts, because it is really useful :joy:

LBJ-the-GOAT:
> > fixed alerts:
> > 1 for Continue statement that does not continue
> 
> Looks like next time I will force to fix all LGTM alerts, because it is really useful 😂

Yeah, I agree with you, I feel that now, and I caused all that alerts. 🤣

PR #1162, opened 2021-04-25
fix LGTM alerts

LBJ-the-GOAT:
Try to fix LGTM alerts.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/device/debugger/BreakpointManager.java
+++ b/jadx-gui/src/main/java/jadx/gui/device/debugger/BreakpointManager.java
@@ -166,7 +166,7 @@ public class BreakpointManager {
 
 		@Override
 		public int hashCode() {
-			return 31 * (int) codeOffset + 31 * cls.hashCode() + 31 * mth.hashCode();
+			return Objects.hash(codeOffset, cls, mth);
 		}
 
 		@Override
--- a/jadx-gui/src/main/java/jadx/gui/device/debugger/BreakpointManager.java
+++ b/jadx-gui/src/main/java/jadx/gui/device/debugger/BreakpointManager.java
@@ -166,7 +166,7 @@ public class BreakpointManager {
 
 		@Override
 		public int hashCode() {
-			return (int) (31 * codeOffset + 31 * cls.hashCode() + 31 * mth.hashCode());
+			return 31 * (int) codeOffset + 31 * cls.hashCode() + 31 * mth.hashCode();
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


jadx-gui/src/main/java/jadx/gui/device/debugger/BreakpointManager.java
```diff
@@ -166,7 +166,7 @@ protected String getFullMthRawID() {
 
 		@Override
 		public int hashCode() {
-			return (int) (31 * codeOffset + 31 * cls.hashCode() + 31 * mth.hashCode());
+			return 31 * (int) codeOffset + 31 * cls.hashCode() + 31 * mth.hashCode();
```
skylot, line 169: This is quite bad implementation of `hashCode`. If you are not sure how to write it correctly just use buildin helper method:
```suggestion
			return Objects.hash(codeOffset, cls, mth);
```

jadx-gui/src/main/java/jadx/gui/device/debugger/smali/Smali.java
```diff
@@ -660,9 +660,13 @@ private void fmtCols(InsnData insn, LineInfo line) {
...
+		int bytecode = 0;
 		for (int i = 0; i < maxLen; i++) {
-			int temp = ((bytes[i++] & 0xff) << 8) | (bytes[i] & 0xff);
-			inHex.append(String.format("%04x ", temp));
+			bytecode = bytecode << 8 | bytes[i] & 0xff;
+			if (i % 2 == 1) {
+				inHex.append(String.format("%04x ", bytecode));
+				bytecode = 0;
+			}
 		}
```
skylot, line 670: I think this can be simplified with just changing step to 2:
```java
		for (int i = 0; i < maxLen - 1; i += 2) {
			int temp = ((bytes[i++] & 0xff) << 8) | (bytes[i] & 0xff);
			inHex.append(String.format("%04x ", temp));
		}
```
Anyway previous version must be output incorrect bytes :(
Please test before submit.

skylot (changes requested):

skylot:
@LBJ-the-GOAT also please don't close/open new PR for just update code. Commit fixing commit to same branch.

jadx-gui/src/main/java/jadx/gui/device/debugger/smali/Smali.java
LBJ-the-GOAT: I tested it, and it outputs the same result with the previous code.

LBJ-the-GOAT:
> @LBJ-the-GOAT also please don't close/open new PR for just update code. Commit fixing commit to same branch.

Aye.

jadx-gui/src/main/java/jadx/gui/device/debugger/smali/Smali.java
skylot: Haha, I also messed this time ))
No need to increment `i` inside loop:
```java
		for (int i = 0; i < maxLen - 1; i += 2) {
			int temp = ((bytes[i] & 0xff) << 8) | (bytes[i + 1] & 0xff);
			inHex.append(String.format("%04x ", temp));
		}
```

LBJ-the-GOAT: Okay.

skylot (approved):

Issue #1171, opened 2021-05-18
[feature] Show hexdecimal instead decimal in Java view


LGLTeam:
I'm dealing with resource ID error and wanted to view in jadx to see which ID caused the problem, but the thing is it shows as decimal, i must view as Smali to see the hexdecimal of ID. I know it's easy to see field names in smali but i like to read in Java better, i don't understand smali much.

![image](https://user-images.githubusercontent.com/61142621/118686245-1fedf300-b804-11eb-97b9-e9cede60d6f2.png)

Can we have an option to show hexdecimal instead decimal in Java view? or at least right-click and select Hexdecimal like in IDA Pro?

![image](https://user-images.githubusercontent.com/61142621/118686193-149ac780-b804-11eb-839b-4b9e86b14590.png)

skylot:
@LGLTeam I can make a right-click menu action like `Copy as Hex (0xFF)`, which also shows number in hex.
And yes, I can make a global option to render all numbers in hex format, not sure if it is really needed, but it is possible.

P.S.: As workaround you can export java code to Android Studio, next `Ctrl+Enter` on number and select `Convert number to ...`

skylot:
As @LBJ-the-GOAT suggested in PR #1173 I made a hex format for IDs in R class.
@LGLTeam if you still want other option let me know :D

LGLTeam:
The PR one is enough for me :)

psolyca:
Hey ^^
Could this be reopened ?
I'd like an option to globally have hex number format.
My project uses hex numbers and converting dec to hex or hex to dec is a pain.
Thanks

skylot:
> I'd like an option to globally have hex number format

@psolyca ok, I can do that

FabianPastor:
Sorry for comenting it here, not suer if it's worth to open a new "issue"...
I am now at 1.4.4 and upgraded from 1.1.0 and by using the cli jadx i noticed that R.java and probably other places it changed all from decimal to hexadecimal.

Just asking to have some kind of flag to go back to the old behavior. Just leaving it here. If there are no plans to add it i'll deal with it xD. 
Thank you :D

skylot:
Global option added by @Arsylk in PR #1869.
Auto - mode will use hex for resources values, dec - for others. Other modes will force selected type everywhere.

PR #1173, opened 2021-05-20
resource id in hex in R class

LBJ-the-GOAT:
#1171 

Use hex for the resource ids, I think It makes RE experience better.

Screenshot (16)


skylot:
I see some flaws in this PR:
1. Values changed only for fields generated by jadx, exist fields in R classes not affected
2. `EncodedValue` type supposed to be a raw decoded type (that is why it placed in `jadx.api.plugins` module), so changing it just for format output is not a good idea.

I think a straight approach is fine here:
```java
ClassNode parentClass = cls.getParentClass();
if (parentClass != null && parentClass.getShortName().equals("R")
		&& encodedValue.getType() == EncodedType.ENCODED_INT) {
	// force hex format for Android resources ids
	code.add(String.format("0x%X", ((Integer) encodedValue.getValue())));
} else {
	annotationGen.encodeValue(cls.root(), code, encodedValue);
}
```
instead line at https://github.com/skylot/jadx/blob/4e5fac4b88e2d4a0a2b1b0f7270b569dc4de651f/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java#L405

Other options are:
- Mark `R` classes with additianal flag (like `ANDROID_R_CLASS`)
- Mark every resource field with `FORMAT_AS_HEX` flag

Anyway, we will need some custom case for render format choice.
Also, I want to notice that this code will be changed as soon as `java-input` branch get merged, check here: https://github.com/skylot/jadx/blob/317a260ec9cea7ec324f01f7ddb4e8c49456c3fc/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java#L413-L421

LBJ-the-GOAT:
Hello, I had considered just patching to check if the class is R and was thinking maybe we would need to apply hex to other classes as well so I add a new field to `EncodedValue`. 

Looks like this `java-input` new branch got some exciting things going on, so I think should leave this hex formatting job to you.

Good day  :D

Issue #1119, opened 2021-02-18
[feature] Add Quark-Engine as an APK analyzer


pulorsok:
### Description

[Quark-Engine](https://github.com/quark-engine/quark-engine) is an APK malware analyzer which can tell what APK does. We believe that Quark as a static analyzer can cooperate very well with Jadx. The user first can take a quick look at the APK malware activities then trace down in source code using Jadx, I think that can greatly improve efficiency for malware analytics.

Here is the scenario :
1. User using Jadx to decompile APK
2. Quark analyze APK then present a potential malware activities report
3. User click on an activity that wants to trace
4. Search and highlight the corresponding smali or java source codes

So far, we have experience with two projects integration ([APKLab](https://github.com/APKLab/APKLab), [ghidraquark](https://github.com/quark-engine/ghidraquark))

If you agree to this idea, I will start working on the integration and submit PR when finished.

### Additional context
Below are features and reports released recently.

![](https://i.imgur.com/OCqqD3Y.png)
![](https://i.imgur.com/l8QTVXh.png)
![](https://i.imgur.com/CyyG7tR.png)

jpstotz:
Of course for some people that may be of interest. However I fear that for the common Jadx user there is one main problem: the quark-engine bases on Python and Jadx is a pure Java based solution. Therefore I would assume that a large number of Jadx users don't even have Python installed. This applies I would assume for at least 90% of all Windows users. @skylot Does GitHub provide information about the platforms of the users downloading Jadx or is there a different way to determine how the OS distribution is among the Jadx users? 

A second problem I see is that Jadx does not yet have some sort of plugin concept. Therefore an integration like ghidraquark which integrates Quark-Engine into Ghidra (which is also Java based like Jadx) is not that easy. Therefore all changes would have to be integrated into Jadx itself. I don't have any experience with Java-to-Python interfaces but from my perspective it would be recommendable that such changes would consist of pure Java code, so that building Jadx will stay as simple as it is at the moment.
Do you think this is feasible?

skylot:
@jpstotz 
> Does GitHub provide information about the platforms of the users downloading Jadx

No, I don't have such statistics, but I think you are right and most users are on Windows.

For integration, we can just use command line interface and assume that Quark already installed on user machine. So we just run shell command and parse output (Quark can output report in json format), next format report in nice page similar to current `APKSignature` node. Such integration is quite feasible.

[...]

PR #1199, opened 2021-06-25
Improvements of Quark integration

pulorsok:
This PR refers to [#1119](https://github.com/skylot/jadx/issues/1119).
### Describe the feature

Hi, sorry for taking so long to fix this.
We (me, @18z, @krnick) have implemented some features:
- **Quark installation in Jadx**:
    - A feature that allows users to install Quark-Engine in Jadx with `pip3` (also available on Windows)
- **Error/warning dialog**:
    - A feature that pops up the error/warning dialog when Quark was (or maybe will) not work correctly
- **Improvement of progress bar**:
    - Implement the progress bar of Quark to `mainWindow`

Changes in this PR:
- **Quark installation in Jadx**:
    - Add pip3 and Quark installed check
    - Add function to create virtualenv
    - Add quark install process
    - Add quark rules update process
- **Error/warning dialog**:
    - Add a feature that pops up an error dialog for each exception
    - Add a feature that pops up a warning dialog when the selected APK file too large
- **Improvement of progress bar**:
    - Change `QuarkTask` from `extend SwingWorker` to `implement IBackgroundTask`
    - Remove progress bar from Quark dialog

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/QuarkDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/QuarkDialog.java
@@ -12,7 +12,9 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import javax.swing.JButton;
 import javax.swing.JComboBox;
@@ -44,6 +46,7 @@ class QuarkDialog extends JDialog {
 	private static final String QUARK_INTERRUPT_MESSAGE = "Quark process interrupted: {}";
 	private static final String QUARK_FAILED_MESSAGE = "Failed to execute Quark.";
 	private static final String QUARK_CMD = "quark";
+	private static final int LARGE_APK_SIZE = 30;
 	private static final Path QUARK_DIR_PATH = Paths.get(System.getProperty("user.home"), ".quark-engine");
 
 	private Path venvPath = Paths.get(QUARK_DIR_PATH.toString(), "quark_venv");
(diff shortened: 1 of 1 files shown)


jpstotz:
I have seen in this PR that you are using `pip`. Wouldn't it be better to use `pip3` just to make sure that on systems where both Python versions 2 and 3 are installed version 3 is used?

skylot (approved):

skylot:
@pulorsok this PR is good. Thanks!

> Optimizing user experience (e.g., improve progress bar)

For background task progress and notification, you can implement `IBackgroundTask` and execute it from `BackgroundExecutor` instance stored in main window

@jpstotz 
> Wouldn't it be better to use pip3

I think just `pip` is ok, it may be symlinked to desired/default pip version. And forcing one version can be misleading.

Anyway, this PR is ready to merge, but I prefer to wait couple days, so other users may also check this PR and share some  suggestions or issues.

green9317:
I think this is a good change but still need to notify the user when something goes wrong.  Perhaps if you could have an error icon with a tooltip or something similar in addition to logging the error.

Experience / Feedback:
My default pip was still for python 2 as I rarely use Python and Quark is only supported with Python 3.7+.  This caused the installation to fail but within JadX I didn't get a notification.

After manually installing quark the first APK that I tested just spun without doing anything.  Checking the error log shows that quark failed to execute.  I attempted the exact same apk using the command line Quark and got a memory error.  

When I attempted to run Quark against a smaller apk it did work and I got the results as expected.

LBJ-the-GOAT:
@green9317  yeah,  I got confused once before but I didn't actually need it so I just ignored it.

green9317:
To be honest ill probably never use Quark within JadX unless the GUI is updated with a specific viewer to make it easier to find the referenced potions of code.  That said I know some people that get really excited with this sort of automated analysis. 

Just adding feedback about my difficulties so it can get documented / fixed before the next stable version of JadX gets released.  I dont think it needs to be included in this PR as the code does work and is am improvement on the existing implementation.

pulorsok:
Hi, @skylot, @jpstotz, @green9317, LBJ-the-GOAT. 
We appreciate all your suggestion and feedback.

> I think just pip is ok, it may be symlinked to desired/default pip version. And forcing one version can be misleading.

I agree with @jpstotz. Since Quark requires python version 3.7+ for execution, it is better to specify `pip3`.

> For background task progress and notification, you can implement IBackgroundTask and execute it from BackgroundExecutor instance stored in main window

Thanks for the suggestion. I will implement that in the next PR for the background progress.

---
@green9317, @LBJ-the-GOAT thanks for the feedback
> I think this is a good change but still need to notify the user when something goes wrong. Perhaps if you could have an error icon with a tooltip or something similar in addition to logging the error.

We will implement this feature in the next PR as soon as this one merged.

> I attempted the exact same apk using the command line Quark and got a memory error.

Not sure which error you are faced, but when Quark analyzing a large APK, it sometimes takes a long time to analyze it and consumes much memory.
We will also add a large size APK warning in the next PR.

skylot:
@pulorsok please apply all additional changes and suggestions into this PR.

pulorsok:
Hi, @skylot, @jpstotz
Thank you for the suggestion.  :D
I have applied all the changes in this PR. 

> Just read from the stream until you reach EOF, because otherwise you may loose the last lines when the process exists but not al lines have been read yet. The used InputStream will automatically close if the process ends.

And I have fixed the problem that loses the last lines when reading the stream of Quark process.

skylot (approved):

skylot:
@pulorsok thanks!
But looks like you misuse jobs in background task, these jobs executed in parallel if threads count set to more than 1 in preferences. So, install and analysis jobs can be executed at the same time. I will fix that myself because it may need some adjustments for tasks implementation.

Issue #1208, opened 2021-07-18
[core] Broken AndroidManifest.xml


AndnixSH:
**Describe error**
I came across an APK with comes with broken AndroidManifest.xml which looks like obfuscation. This is not the first time i have seen this, other apk did come out as this as well when decompiled using apktool.jar, and i had already reported this issue on apktool repo as well, now it seems they already fixed it

Can jadx be implemented the fix as well?

APK https://apkcombo.com/apk-downloader/?package=com.abiss.roem&device=default&arches=armeabi-v7a&sdkInt=default&format=apk&dpi=480&lang=en

![image](https://user-images.githubusercontent.com/40742924/126072526-209db35a-e260-4e22-8b5d-2e7db388ec8a.png)

jpstotz:
An interesting manifest. The XML strings pool contains obfuscated value, only the AndroidResMap contains the real value.

Currently Jadx prefers the value from xml strings pool and only uses the one from AndroidResMap if it is missing the the strings pool. Assuming all the entries in AndoridManifest.xml in the APK you found are valid (and not just garbage aiming in hiding the relevant entries) then Android seems to work the other way around, first check AndroidResMap and the the string pool.

I have prepared a branch with this change which works on the linked APK file. But I have not tested this [new branch](https://github.com/jpstotz/jadx/commits/manifest_attr) on other APK files.

Just the changes: https://github.com/jpstotz/jadx/commit/b71fac78ad594ffefef1f7bf8c2478afedd75b51

skylot:
Should be fixed by PR #1214.

PR #1214, opened 2021-07-24
reverse XML attribute name decoding priority

jpstotz:
Reversed XML attribute name decoding priority as discussed in #1208

I looked at different binary XML files and they look consistent, there it seems that reversing the XML attribute decoding priority does not have a negative effect.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java
@@ -335,19 +335,25 @@ public class BinaryXMLParser extends CommonBinaryParser {
 	}
 
 	private String getAttributeName(int id) {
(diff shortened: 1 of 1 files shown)


skylot (approved):

skylot:
@jpstotz thanks!
Do we need to wait for someone to also check this PR?

jpstotz:
@skylot It correctly decodes the attribute names in AndroidManifest.xml of the app mentioned in #1208 and I checked several other APKs, therefore I would think it is ready to be merged into the unstable version. 

Or did you had other people in mind to check this PR?

skylot:
Well, I thought at least @AndnixSH (author of #1208) can check it. But he doesn't reply after a week, so there is a little hope.
Also, we can ask @bagipro he often fixes issues in XML resources.

Anyway, maybe you are right, and I will just merge this PR and if new cases appear we will fix them :+1:

Issue #998, opened 2020-10-25
[core] Add parameter to disable all debug comments


sergey-wowwow:
Hey @skylot,
Can you implement a flag please that disables all comments such as:
```java
    /* JADX INFO: this call moved to the top of the method (can break code semantics) */
    public /* synthetic */ AudioPlayer(java.lang.String str, com.wickr.calling.media.audio.AudioFormat audioFormat2, com.wickr.calling.media.audio.AudioPlayer.Callback callback2, com.wickr.calling.media.audio.AECHelper aECHelper, int i, kotlin.jvm.internal.DefaultConstructorMarker defaultConstructorMarker) {
        this(str, audioFormat2, callback2, (i & 8) != 0 ? null : aECHelper);
    }
```
```java
    public final /* synthetic */ class WhenMappings {
```
```java
    /*  JADX ERROR: JadxRuntimeException in pass: InlineMethods
        jadx.core.utils.exceptions.JadxRuntimeException: Failed to process method for inline
        	at jadx.core.dex.visitors.InlineMethods.processInvokeInsn(InlineMethods.java:68)
        	at jadx.core.dex.visitors.InlineMethods.visit(InlineMethods.java:43)
        Caused by: jadx.core.utils.exceptions.JadxRuntimeException: Class not yet loaded at codegen stage
        	at jadx.core.dex.nodes.ClassNode.reloadAtCodegenStage(ClassNode.java:589)
        	at jadx.core.dex.visitors.InlineMethods.processInvokeInsn(InlineMethods.java:60)
        	... 1 more
        */
    @org.jetbrains.annotations.NotNull
    public static final java.lang.String getTAG(@org.jetbrains.annotations.NotNull java.lang.Object r3) {
        /*
            java.lang.String r0 = "$this$TAG"
            kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull(r3, r0)
            java.lang.String r3 = r3.toString()
            r0 = 0
            java.lang.String r1 = "."
            r2 = 2
            java.lang.String r3 = kotlin.text.StringsKt.substringAfterLast$default(r3, r1, r0, r2, r0)
            return r3
        */
        throw new UnsupportedOperationException("Method not decompiled: com.wickr.calling.Extensions.getTAG(java.lang.Object):java.lang.String");
    }
```

alissonlauffer:
[...]

PR #1219, opened 2021-08-01
Small improvements (tooltip, dex source file)

jpstotz:
1. If multiple files are loaded in Jadx their file paths are now shown as tooltip on the JRoot node
2. Decompiled classes get a comment from which dex files they have been loaded from. This is useful if the APK files contains dex files in non-standard locations (e.g. in assets or other places).

For 2. I also tried to include the APK source file name in the comment in case multiple files are loaded, but it seems that Jadx does not store this information at least I wasn't able to identify how to access this information.  I assume this would require some bigger changes to `DexFileLoader` to carry the original source file information through all the different loader methods to save it in the end in `DexClassData` .

### DIFF
--- a/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
+++ b/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
@@ -147,6 +147,7 @@ public class ClassGen {
 
 		annotationGen.addForClass(clsCode);
 		insertRenameInfo(clsCode, cls);
+		CodeGenUtils.addInputFileInfo(clsCode, cls);
 		clsCode.startLineWithNum(cls.getSourceLine()).add(af.makeString());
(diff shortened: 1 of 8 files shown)


skylot (approved):

skylot:
@jpstotz thanks!
Although, I am not sure that everyone will need these additional comments, but it is useful :+1:

jpstotz:
@skylot May be I should add an option to enable/disable this type of comments?

skylot:
> @skylot May be I should add an option to enable/disable this type of comments?

Not sure. I don't want to add an option to every small feature. And looks like we need some common approach to add additional info into code and customize it in options. Maybe similar to log level as discussed in #998. Also in jadx-gui we can add a class tooltip or some kind of popup with additional info like in all IDEs.

PR #1225, opened 2021-08-07
Improve jadx#2

### LINKED ISSUE #1182: [feature] Remember project settings
Right now jadx does not remember project settings. For example, lets say I launch two instances of jadx.
The first instance will decompile an obfuscated apk, so I turn on deobfuscation. Then, when I launch the second instance, deobfuscation will be turned on. However, I don't want to deobfuscate it since the second apk contains all required data, so I disable it.
Then I close both instances and open a new one. I open the .jadx project from the obfuscated apk and I have to re-enable deobfuscation because it was disabled last time.

This could be solved by saving settings by project, on the .jadx file. A new object could be added to the json, named, for example, "settings" which contains things like:

"settings": {
    enableDeobfuscation: true,
    excludedPackages: {
        "com.example.excluded"
    },
    autoStartBackgroundDecompilation: false,
    showInconsistentCode: true,
....
}

Then, when opening the project, the settings would be read and everything would be configured automatically.

### LINKED ISSUE #1210: Persistent sync with editor
It would be nice if there is a feature to have the sync with editor option stick so that when you are flipping through tabs it jumps to where a specific class is located without pressing Ctrl+T or by manually clicking the button.

### LINKED ISSUE #1212: Allow drag and drop reordering of tabs
It looks like jadx does not currently allow the reordering of tabs via drag and drop. If you have a lot of tabs open, it might be helpful to allow this so you are keeping all your interesting code close together rather than clicking through the different tabs to find which one you were working on.

MrIkso:
- fixed #1212. It can be difficult to grab hold of the tab title here, as the focus is lost. I haven't figured out what they do with focus yet, but it works if you grab the very bottom of the title tab (see video)
- fixed #1210 . To enable **View -> Always Select Opened File/Class** (see vieeo)
- fixed #1182 . Do not restore the saved GUI (themes, main window size, etc.). Perhaps it is not necessary, therefore, and did not even do with it)
- updated SearchBar in Code Viewer (see screenchot)

## Screenchot
image

## Video
https://user-images.githubusercontent.com/43380144/128615562-7d175a34-1f09-43da-81c0-88a500c5e03d.mp4


MrIkso:
I think it's worth clearing the logs in actions, about unsuccessful builds, otherwise I tried to figure out what was wrong and rolled back the commits. Perhaps I have incorrectly set up the studio for this project to format the code, which causes such problems.

skylot:
> I think it's worth clearing the logs in actions, about unsuccessful builds

Please run `gradle clean build` before pushing changes, it will save your time and reduce number of notifications for me and other watchers. 

>  Perhaps I have incorrectly set up the studio for this project to format the code

You just need to run `gradle spotlessApply` to apply formatting, so it doesn't matter which formatting options you have in your IDE.
Changes in import with star (`.*`) is ok, because I don't know how to enforce them :disappointed:, anyway I prefer to not use star imports.

> It can be difficult to grab hold of the tab 

This mean this feature are not ready yet, and this issue must be fixed before merging.

> Remember project settings. Do not restore the saved GUI

Same here, it is too early to merge this. Also, project settings must have some kind of UI editor and explicitly show which option saved in which pool (global or project). In your changes it is unclear how options are saved and in most cases this feels like an error.

Other issues I notice in your PR:
- **never** commit commented code.
- it is better to split changes for different issues/features in different PR. It will be easier for you to fix code in case of requested changes, also it will allow keeping nice history in repository.

As conclusion:
Changes for search bar and auto sync are good but I can't merge them because other changes are not ready yet.

jadx-gui/src/main/java/jadx/gui/settings/JadxProject.java
```diff
@@ -39,19 +40,23 @@
...
 	private transient MainWindow mainWindow;
-	private transient JadxSettings settings;
 
 	private transient String name = "New Project";
 	private transient Path projectPath;
 
 	private transient boolean initial = true;
 	private transient boolean saved;
 
+	@SerializedName("files")
```
skylot, line 50: I think these annotations not needed if name equals field name.

jadx-gui/src/main/java/jadx/gui/ui/DnDTabbedPane.java
```diff
@@ -0,0 +1,353 @@
...
+		// .filter(Action::isEnabled)
+		// .ifPresent(a -> a.actionPerformed(new ActionEvent(this, ActionEvent.ACTION_PERFORMED, null, 0,
+		// 0)));
+		// // ActionMap map = getActionMap();
+		// // if (Objects.nonNull(map)) {
+		// // Action action = map.get(actionKey);
+		// // if (Objects.nonNull(action) && action.isEnabled()) {
+		// // action.actionPerformed(new ActionEvent(this, ActionEvent.ACTION_PERFORMED, null, 0, 0));
+		// // }
+		// // }
```
skylot, line 65: Don't leave commented code.

skylot (changes requested):

jadx-gui/src/main/java/jadx/gui/ui/DnDTabbedPane.java
MrIkso: this is not my code, that's why I didn't clear it

jadx-gui/src/main/java/jadx/gui/settings/JadxProject.java
MrIkso: I know, but it is more correct to do it for json serialization.  In general, it should be one model class that describes the structure of the project/settings...

jadx-gui/src/main/java/jadx/gui/ui/DnDTabbedPane.java
skylot: Hm, it is even worse, because now we need to add copyright and verify license of this code.

MrIkso:
> This mean this feature are not ready yet, and this issue must be fixed before merging.

That's who will tell me how to do it right.  on another project everything is fine, but there are some peculiarities of changing focus. I don't like the way it is done...

jadx-gui/src/main/java/jadx/gui/ui/DnDTabbedPane.java
MrIkso: Code from [here](https://java-swing-tips.blogspot.com/2010/02/tabtransferhandler.html?m=1)

PR #1245, opened 2021-09-01
Improve jadx#3

### LINKED ISSUE #1210: Persistent sync with editor

MrIkso:
- updated SearchBar in Code Viewer
- fixed #1210 . To enable View -> Always Select Opened File/Class


skylot:
@MrIkso thanks!

PR #1256, opened 2021-10-01
Use param names in annotations

DanielFi:
Use the values in MethodParameters annotations as parameter names when available.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/codegen/MethodGen.java
+++ b/jadx-core/src/main/java/jadx/core/codegen/MethodGen.java
@@ -1,5 +1,6 @@
 package jadx.core.codegen;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


jadx-plugins/jadx-plugins-api/src/main/java/jadx/api/plugins/input/data/attributes/types/AnnotationsAttr.java
```diff
@@ -23,9 +22,7 @@ public static AnnotationsAttr pack(List<IAnnotation> annotationList) {
 		}
 		Map<String, IAnnotation> annMap = new HashMap<>(annotationList.size());
 		for (IAnnotation ann : annotationList) {
-			if (ann.getVisibility() != AnnotationVisibility.SYSTEM) {
```
skylot, line 26: Looks like this change break many tests :(

jadx-core/src/main/java/jadx/core/codegen/MethodGen.java
```diff
@@ -180,8 +183,25 @@ private void addOverrideAnnotation(ICodeWriter code, MethodNode mth) {
 		}
 	}
 
+	private List<String> extractParamNamesFromAnnotation() {
+		AnnotationsAttr annotationList = mth.get(JadxAttrType.ANNOTATION_LIST);
+		if (annotationList != null) {
+			JadxAnnotation annotation = (JadxAnnotation) annotationList.get("Ldalvik/annotation/MethodParameters;");
```
skylot, line 189: As soon as jadx now trying to decompile java bytecode and not only dex, dalvik specific annotations should be parsed in `jadx-dex-input` plugin and data class for common representation of such attribute should be placed in `jadx-plugins-api`.

skylot (changes requested):

skylot:
So, no reply :(

Issue #1274, opened 2021-11-14
[feature] output jobf file when decompiling to Java


yotamN:
Currently there is no way to get the renaming list if you decompile to Java through the CLI without activating deobfuscation.

The expected behavior would be to at least output it when I pass the `--deobf-cfg-file` flag

skylot:
@yotamN sure, thanks for suggestion!

skylot:
Fixed in PR #1275

PR #1275, opened 2021-11-14
Save jobf when decompiling to Java through the cli

yotamN:
As I suggested in #1274, the commit will output a jobf file when decompiling to Java.

I'm not sure if it should be on by default because there doesn't seem to be a way to disable it except of passing `/dev/null` as the deobfuscation name but I guess this is the same problem with the deobfuscation feature. What do you think?

### DIFF
--- a/jadx-core/src/main/java/jadx/core/deobf/DeobfPresets.java
+++ b/jadx-core/src/main/java/jadx/core/deobf/DeobfPresets.java
@@ -134,7 +134,7 @@ public class DeobfPresets {
 			list.add(String.format("m %s = %s", mthEntry.getKey(), mthEntry.getValue()));
 		}
 		Collections.sort(list);
-		if (list.size() == 0) {
+		if (list.isEmpty()) {
 			if (LOG.isDebugEnabled()) {
 				LOG.debug("Deobfuscation map is empty, not saving it");
 			}
--- a/jadx-core/src/main/java/jadx/core/deobf/DeobfPresets.java
+++ b/jadx-core/src/main/java/jadx/core/deobf/DeobfPresets.java
@@ -134,6 +134,12 @@ public class DeobfPresets {
 			list.add(String.format("m %s = %s", mthEntry.getKey(), mthEntry.getValue()));
 		}
 		Collections.sort(list);
+		if (list.size() == 0) {
+			if (LOG.isDebugEnabled()) {
+				LOG.debug("Deobfuscation map is empty, not saving it");
+			}
+			return;
+		}
 		Files.write(deobfMapFile, list, MAP_FILE_CHARSET,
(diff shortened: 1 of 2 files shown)


skylot:
> if it should be on by default because

I think this is ok, except it will also save an empty file, so it is better to skip file write for empty list at https://github.com/skylot/jadx/blob/dfdc14ea8618a7242d04bd6ea066d3e6b60a10b9/jadx-core/src/main/java/jadx/core/deobf/DeobfPresets.java#L137

yotamN:
Sure, I added a check and a debug log line

jadx-core/src/main/java/jadx/core/deobf/DeobfPresets.java
```diff
@@ -134,6 +134,12 @@ public void save() throws IOException {
 			list.add(String.format("m %s = %s", mthEntry.getKey(), mthEntry.getValue()));
 		}
 		Collections.sort(list);
+		if (list.size() == 0) {
```
skylot, line 137:
```suggestion
		if (list.isEmpty()) {
```

skylot (approved):

skylot:
@yotamN thanks!

CmP-lt:
This change causes jobf file to be saved when JADX is used in GUI mode and deobfuscation is not enabled. Was this intended?

skylot:
@CmP-lt jadx can rename identifiers even if deobfuscation is disabled, check `Rename identifiers` group in preferences.
It is possible to add another option to disable saving jobf file, but I don't think it is really bother anyone and I want to keep options count as little as possible, so just ignore that file if you don't need it :)

Samulia-Sanchez:
@skylot 
I actually came here looking for a way of disabling .jobf file output next to the apk file, as I see no value in it, while it causes me issues. I use jadx GUI to analyze apks in different versions of the same folder hierarchy.  It's been contaminating all folders I open apks from with these files over the past weeks, which in turn causes the folders to show up as different in a folder comparison while they originally weren't.
While I could delete .jobf files recursively, I'd rather not have to do that every time to make sure it doesn't falsify my comparison results.
I would appreciate if .jobf file creation were optional, or at least moved to an entirely different location than the .apk file I open.
It may make sense to store them in the jadx config folder by default and use a hash of the .apk file as their filename, so they can be quickly and reliably found no matter where the .apk file is moved.

skylot:
@Samulia-Sanchez and @CmP-lt 
Ok, I will add a new option to save jobf file, it will be disabled by default.
I open new issue #1351 for additional suggestions and discussion.

skylot:
@yotamN, @Samulia-Sanchez and @CmP-lt
I implement an option to control how to handle deobfuscation map file, and now it is set to disable jobf file saving by default.
Details can be found in https://github.com/skylot/jadx/issues/1351#issuecomment-1046085306

PR #1278, opened 2021-11-18
fix: configured resource indexing size limit is now correctly considered

jpstotz:
According to the preferences page the resource size index limit is specified in MB but in code it was size * 10KB instead.


skylot (approved):

skylot:
@jpstotz thanks!

PR #1279, opened 2021-11-20
chore: Updated German translation

jpstotz:
Usually I prefer Jadx in English but the last days a number of times my preferences were reset so I ended up with German as language and noticed that some translations were really hard to understand even for me as I know Jadx very good.


skylot (approved):

skylot:
@jpstotz great! Thank you!

skylot:
> number of times my preferences were reset

Do any errors appear in the logs? Maybe I break something in recent commits 😨

jpstotz:
@skylot I am not sure what caused the reset. I am also not sure if the whole preferences were reset or just the language. UI language is just the most obvious preference. 

In the end it is nothing I was able to reproduce. Also I can't name a commit where this started or stopped as I use Jadx on multiple computers in different stable/unstable versions and frequently change preference options.

PR #1105, opened 2021-01-27
Improve color schemes

Surendrajat:
### Description

This PR:
1. Removes hard-coded color value for highlighting symbol(default color looks better)
2. Adds a new color scheme: **druid**

Related but doesn't clo$e #1101

PR #1283, opened 2021-11-22
smali code highlighting is basically perfect

MartinKayJr:
smali highlight is here
#1105
Follow-up to make some improvements to the color matching


skylot:
@MartinKayJr great work!
Several questions:
- why you don't want to include original jflex file? It will be easier to change/fix token types later
- did you copy jflex file from somewhere or write it yourself? Just trying to check copyright :)
- can we use different types of instructions and type names, like use `identifier` for instructions? Because now whole line have same color and hard to read:

MartinKayJr:
@skylot hey bro
1.I don’t know where the jflex file should be placed. I put it in the resource directory of the gui in the first submission.
2.This jflex took me 11 consecutive hours to complete and was completed by myself.
3.Regarding the color, I think the RSTA reference manual and the preset Token are not suitable for the smali language, so I am going to optimize the color matching in the follow-up

MartinKayJr:
In my recent submission, I transferred the jflex file to `config/jflex` and provided a simple shell script for reference to generate `SmaliTokenMarker.java`.

I tried to use `Token.identifier` for the instructions , but there was no color, so I changed the fully qualified class name to `DATA_TYPE`.

skylot (approved):

skylot:
@MartinKayJr thanks! Looks good.
Although, I think, I will rewrite your shell script using gradle and jflex plugin to make it environment independent.

PR #1286, opened 2021-11-24
Minor fixes XML decoding and APK signature description

jpstotz:
I noticed to minor problems:

1. Hexadecimal integers e.g. in AndrdoirManifest.xml were correctly decoded as hex values but the `0x`prefix was missing.
2. The warning about unprotected entries in APK signature only applies to v1 signatures. EN + DE text updated accordingly. Not sure how to handle this for KR and CN languages.


skylot (approved):

skylot:
> Not sure how to handle this for KR and CN languages.

@jpstotz I am also not sure, because change is not so big for reverting to english version and there are no good way to mark these lines as outdated. May be comment will help, but it may not pass current checks.

PR #1290, opened 2021-11-28
Updated German translation

Henry2o1o:

### Description
Please describe your pull request.
Reference issue it fixes.


jadx-gui/src/main/resources/i18n/Messages_de_DE.properties
```diff
@@ -121,14 +121,14 @@ preferences.decompile=Dekompilierung
...
-preferences.lineNumbersMode=Modus zur Anzeige der Zeilennummern im Editor
+preferences.lineNumbersMode=Editor Zeilennummern-Modus
 preferences.check_for_updates=Nach Updates beim Start suchen
 preferences.fallback=Zwischencode ausgeben (einfacher Speicherauszug)
 preferences.showInconsistentCode=Inkonsistenten Code anzeigen
 preferences.escapeUnicode=Unicodezeichen escapen
 preferences.replaceConsts=Konstanten ersetzen
 preferences.respectBytecodeAccessModifiers=Modifikatoren für Bytecode-Zugriff beachten
-preferences.useImports=Generiere Import statements
+preferences.useImports=Importanweisungen anwenden
```
jpstotz, line 131: This translation is in my opinion not better than before. Based on your description it is unclear what it means.

jadx-gui/src/main/resources/i18n/Messages_de_DE.properties
```diff
@@ -229,7 +229,7 @@ certificate.serialSHA1=SHA-1 Fingerabdruck
 certificate.serialSHA256=SHA-256 Fingerabdruck
 certificate.serialPubKeyY=Y
 
-apkSignature.signer=Unterzeichner
+apkSignature.signer=Signieren
```
jpstotz, line 232: In the context it used we need substantive, not a verb. It is used for displaying the signatures of the APK, therefore "Unterzeichner" ist the better translation.

jadx-gui/src/main/resources/i18n/Messages_de_DE.properties
```diff
@@ -242,7 +242,7 @@ apkSignature.unprotectedEntry=Dateien, die nicht durch eine APK-Signatur v1 gesc
 issues_panel.label=Probleme:
 issues_panel.errors=%d Fehler
 issues_panel.warnings=%d Warnungen
-issues_panel.tooltip=Im der Logansicht anzeigen
+issues_panel.tooltip=Im Protokoll-Anzeiger öffnen
```
jpstotz, line 245: I don't think "Anzeiger" is the correct word here see https://www.duden.de/rechtschreibung/Anzeiger

jadx-gui/src/main/resources/i18n/Messages_de_DE.properties
```diff
@@ -128,7 +128,7 @@ preferences.showInconsistentCode=Inkonsistenten Code anzeigen
 preferences.escapeUnicode=Unicodezeichen escapen
 preferences.replaceConsts=Konstanten ersetzen
 preferences.respectBytecodeAccessModifiers=Modifikatoren für Bytecode-Zugriff beachten
-preferences.useImports=Importanweisungen anwenden
+preferences.useImports=Importierte Anweisungen anwenden
```
jpstotz, line 131: I still think that people don't understand this preference option if you name it `Importierte Anweisungen anwenden`. Actually it decides if Jadx always uses fully qualified class names in the source code (option disabled) or if `import packagea.ClassNameY` statements are generated at the beginning at each Java source file. 

In my opinion that can be understood by the current translation `Generiere Import statements`  but is more difficult to understand by your translations. Please keep in mind that just translating items is not what makes a good translation. A good translation should name the option it belongs to in a way people using the language understand it. Therefore for a translation it is best to check where it is used, then understand what it actually means and then create the translation.

jpstotz (approved):

skylot:
@Henry2o1o and @jpstotz thank you for your work!

PR #1293, opened 2021-11-29
XML decoding speed enhancement

jpstotz:
On a large APK I had the problem that AndroidManifest.xml generation took ~1 minute or more because the method to search for class references loops over all classes every time a potential reference is found in the XML. 

Instead I added a cache (on-demand initialization) so that we only have to loop one time over all classes now. To reflect changes the cache is deleted after each parse call. 

Cache handling is not thread safe as in my tests it seems like resource decoding is not done multi-threaded.


skylot (approved):

skylot:
@jpstotz nice improvement, thanks!
I also added a small change to skip this whole method for `jadx-cli` (attached annotations not needed).

Issue #1312, opened 2021-12-20
[feature] Use relative paths in project file


kingfisher63:
Paths in the _files_ section of the project file (.jadx) are currently absolute paths. As a result paths become invalid when the project folder is moved to a new location. Using relative paths (if a file is in the same folder as the project file or a subfolder thereof) would resolve this.

jpstotz:
A very good idea. I prepared the necessary changes.

skylot:
Fixed in PR #1313

PR #1313, opened 2021-12-20
chore: use relative file paths in .jadx project file (#1312)

jpstotz:
Additionally the save as dialog now suggests a file name for the jadx project file if only one file is loaded.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/settings/JadxProject.java
+++ b/jadx-gui/src/main/java/jadx/gui/settings/JadxProject.java
@@ -31,22 +31,13 @@ import jadx.gui.settings.data.ProjectData;
 import jadx.gui.settings.data.TabViewState;
 import jadx.gui.ui.MainWindow;
 import jadx.gui.ui.codearea.EditorViewState;
-import jadx.gui.utils.PathTypeAdapter;
+import jadx.gui.utils.RelativePathTypeAdapter;
 
 public class JadxProject {
 	private static final Logger LOG = LoggerFactory.getLogger(JadxProject.class);
 
 	private static final int CURRENT_PROJECT_VERSION = 1;
 	public static final String PROJECT_EXTENSION = "jadx";
-
-	private static final Gson GSON = new GsonBuilder()
-			.registerTypeHierarchyAdapter(Path.class, PathTypeAdapter.singleton())
-			.registerTypeAdapter(ICodeComment.class, GsonUtils.interfaceReplace(JadxCodeComment.class))
-			.registerTypeAdapter(ICodeRename.class, GsonUtils.interfaceReplace(JadxCodeRename.class))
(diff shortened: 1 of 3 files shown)


jadx-gui/src/main/java/jadx/gui/settings/JadxProject.java
```diff
@@ -179,9 +170,10 @@ public void saveAs(Path path) {
 	}
 
 	public void save() {
-		if (getProjectPath() != null) {
-			try (Writer writer = Files.newBufferedWriter(getProjectPath(), StandardCharsets.UTF_8)) {
-				GSON.toJson(data, writer);
+		Path savePath = getProjectPath();
+		if (savePath != null) {
+			try (Writer writer = Files.newBufferedWriter(savePath, StandardCharsets.UTF_8)) {
+				buildGson(savePath.getParent()).toJson(data, writer);
```
skylot, line 176:
```suggestion
				buildGson(savePath.toAbsolutePath().getParent()).toJson(data, writer);
```
@jpstotz you should be extra carefull while using `Path.getParent()` method because it can return `null` even for non-root files. For example `jadx-gui test.jadx` will fail with NPE. So it is better to expand path using `toAbsolutePath()` before calling `.getParent()`

jadx-gui/src/main/java/jadx/gui/settings/JadxProject.java
```diff
@@ -192,7 +184,7 @@ public void save() {
 	public static JadxProject from(Path path) {
 		try (Reader reader = Files.newBufferedReader(path, StandardCharsets.UTF_8)) {
 			JadxProject project = new JadxProject();
-			project.data = GSON.fromJson(reader, ProjectData.class);
+			project.data = buildGson(path.getParent()).fromJson(reader, ProjectData.class);
```
skylot, line 187:
```suggestion
			project.data = buildGson(path.toAbsolutePath().getParent()).fromJson(reader, ProjectData.class);
```
same here

jadx-gui/src/main/java/jadx/gui/utils/RelativePathTypeAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+
+public class RelativePathTypeAdapter extends TypeAdapter<Path> {
+	private final Path basePath;
+
+	public RelativePathTypeAdapter(Path basePath) {
+		this.basePath = basePath;
```
skylot, line 16:
```suggestion
		this.basePath = Objects.requireNonNull(basePath);
```
Additional check to fast fail if null somehow leaked here

skylot (changes requested):

skylot:
@jpstotz thanks!

Issue #1349, opened 2022-01-22
[feature]find usage should find the same method in super and interfaces


cyqw:
Some class have override method from super or interfaces.
The caller function may use super or interface to call this method.
jadx.gui.ui.dialog.UsageDialog#collectUsageData should get UseIn of self and same method of super and interfaces.

cyqw:
I can get MethodNode of super and interfaces,how to get JavaMethod?

skylot:
@cyqw it is indeed a bug, I totally forged about that, thanks!
An easier way is to use `JavaMethod.getOverrideRelatedMethods()` to get a list of all overridden methods from class hierarchy, and next merge all usages (`JavaMethod.getUseIn()`) from each method.
Do you willing to prepare PR? If yes, please assign this task for yourself.

cyqw:
I will try to prepare PR.

skylot:
Fixed in PR #1352 by @cyqw

PR #1352, opened 2022-01-23
[feature]find usage should find the same method in super and interfaces #1349

cyqw:


skylot (approved):

skylot:
@cyqw thanks, looks good!
Although, it is not efficient to do a search if we already know which method used in which class. But I am not sure how to better implement this right now. So, I think, I can merge this PR.

Issue #1350, opened 2022-01-23
Anonymous classes are not shown in jobf file


yotamN:
Classes that were renamed to `AnonymousClassX` are not shown in the jobf file and it seems there is no way to know they were renamed. Is it the intended behavior or a bug?

It seems that at `RenameVisitor.checkClassName` change the name without updating the `Deobfuscator` map file. How should I go about adding it? I guess I should not just simply put the the changed class because there is no such method in the `Deobfuscator` class

skylot:
@yotamN it is a bug.
Deobfuscator quite messy right now, and looks like many renames are missing, because deobfuscator only used to generate new names and if we use another way to make a new name it will be missing or incorrect. This is a case for user renames and disabled deobfuscator.
I am thinking about more general approach: before saving, collect renames from `*Info`  classes (`ClassInfo`, `MethodInfo`, `FieldInfo`) directly and merge/overwrite already exists map in deobfuscator.

yotamN:
I've tried to move the `fixClsShortName` into the deobfuscator now but I'm not sure I understand the logic in here. Why does it return after using the deobfuscator in line 90 and why does it change the name to anonymous class when the class starts with a digit?

Adding the rename from line 93 to the deobfuscator seems to break the state, it works but throws an exception

skylot:
> why does it change the name to anonymous class when the class starts with a digit?

Java identifiers can't start with a digit, also it is a way java compiler name anonymous classes (just numbers)

> Why does it return after using the deobfuscator in line 90

Looks like, `newShortName` variable is null, but next code in method using it, so we just quit.

> I've tried to move the `fixClsShortName` into the deobfuscator

I think this is a bad idea because this method trying to fix invalid names and can work without enabled deobfuscator.
As I suggested before, it is better to process already applied renames at the end of `Deobfuscator.fillDeobfPresets()` method like this:
```java
		for (ClassNode cls : root.getClasses()) {
			ClassInfo clsInfo = cls.getClassInfo();
			if (clsInfo.hasAlias()) {
				deobfPresets.getClsPresetMap().put(clsInfo.makeRawFullName(), clsInfo.getAliasShortName());
			}
			// TODO: similar for fields and methods
		}
```
This way, we don't need to worry about missing renames in jobf files.

yotamN:
> it is a way java compiler name anonymous classes (just numbers)

Do you mean *just* numbers or also classes like `1bc`?

[...]

PR #1354, opened 2022-01-23
Only mark classes as anonymous if they are a number

yotamN:
A fix for the problem that you mentioned in my other issue #1350

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/rename/RenameVisitor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/rename/RenameVisitor.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


jadx-core/src/main/java/jadx/core/dex/visitors/rename/RenameVisitor.java
```diff
@@ -130,11 +130,12 @@ private static void checkPackage(Deobfuscator deobfuscator, ClassNode cls, Class
 	private static String fixClsShortName(JadxArgs args, String clsName) {
 		boolean renameValid = args.isRenameValid();
 		if (renameValid) {
-			char firstChar = clsName.charAt(0);
-			if (Character.isDigit(firstChar)) {
+			if (clsName.matches("^\\d+$")) {
```
skylot, line 133: Please split compile and matches parts for better performance:
```java
// additional field
private static final Pattern ANONYMOUS_CLASS_PATTERN = Pattern.compile("^\\d+$");

// replace condition with
if (ANONYMOUS_CLASS_PATTERN.matcher(clsName).matches()) {
```

skylot (changes requested):

jadx-core/src/main/java/jadx/core/dex/visitors/rename/RenameVisitor.java
yotamN: Fixed it

skylot (approved):

skylot:
@yotamN great! Thanks :+1:

Issue #1355, opened 2022-01-24
[feature] Menu Action to copy as frida code snippets  


eybisi:
In action menu bar when user right clicks to methods/classes there will be option to `copy as frida code snippet`. It will get class name from methodnode, get argument names and create frida code snippet.

```java
package wocwvy.czyxoxmbauu.slsa;

/* loaded from: classes.dex */
public class clgqtzqdh extends Service {


    public int onStartCommand(Intent intent, int i, int i2) {
        String i3 = this.a.i(this);
        this.b.getClass();
        if (!"Android Security".equals("")) {
            this.b.getClass();
            i3 = "Android Security";
        }
        Toast makeText = Toast.makeText(this, this.a.e(this, "StringAccessibility") + " '" + i3 + "'", 1);
        makeText.setGravity(16, 0, 0);
        makeText.show();
        return i;
    }
}
```
Example scenario will be : right clicking on onStartCommand creates and saves created text to clipboard : 
```js
let clgqtzqdh = Java.use("wocwvy.czyxoxmbauu.slsa.clgqtzqdh")
clgqtzqdh.onStartCommand.implementation = function(intent, i, i2){
	console.log('onStartCommand is called')
	let ret = this.onStartCommand(intent, i, i2);
	console.log('onStartCommand return value : ' + ret)
	return ret
}
```
PR is ready to be fired when we agree on this feature : https://github.com/eybisi/jadx/tree/frida

skylot:
@eybisi sure, submit a PR!

skylot:
Implemented in PR #1356 by @eybisi

only52607:
@eybisi  The copied code snippet has a problem if the class member name is a unprintable character.

For example, the following code will have parsing problems when "onStartCommand" is a unprintable character.

```
let clgqtzqdh = Java.use("wocwvy.czyxoxmbauu.slsa.clgqtzqdh")
clgqtzqdh.onStartCommand.implementation = function(intent, i, i2){
	console.log('onStartCommand is called')
	let ret = this.onStartCommand(intent, i, i2);
	console.log('onStartCommand return value : ' + ret)
	return ret
}
```

A better way would be something like the following

```
let clgqtzqdh = Java.use("wocwvy.czyxoxmbauu.slsa.clgqtzqdh")
clgqtzqdh["onStartCommand"].implementation = function(intent, i, i2){
	console.log('onStartCommand is called')
	let ret = this.onStartCommand(intent, i, i2);
[...]

PR #1356, opened 2022-01-24
add frida action to copy methods/classes as frida snippets

eybisi:
PR for #1355.
Adds `Copy as frida snippet` option to code area popup.
- Adds correct overload signature to function snippets.
- Adds arguments (with names) as function parameter.
- Tracks which class definition already given to user. ( no duplicate definitions )
- Inserts kinda dummy information logging to function body.
- Inserts $ to inner class names so frida will be happy.


skylot (approved):

skylot:
@eybisi thanks!

jpstotz:
This PR has one major mistake: The created frida snippet uses the deobfuscated class and method names which only exist in Jadx, instead of the real class names.

skylot:
@jpstotz oh, I see. Funny that class copy uses correct name.
Anyway, it is better to use `ClassInfo.getRawName()` to get class name used in bytecode.

PR #1359, opened 2022-01-31
chore: improved decoding of flag attributes in binary XML files

### LINKED ISSUE #1156: [core] AndroidManifest protectionLevel flags decoding issue
Hi.

I think there is an issue when decoding the `android:protectionLevel` attribute of permission nodes from AndroidManifest.xml file.

For example, if the value of the attribute is `0x11`, jadx will display `normal|dangerous|system`. A normal permission that is also dangerous doesn't make sense.

May be I am wrong, but I think the flags that are mentioned to be *Base permission type* [here](https://developer.android.com/reference/android/R.attr#protectionLevel) should be mutually exclusive.

See: [ManifestAttributes.java#L179](https://github.com/skylot/jadx/blob/012f7665aaddaf0eea023cfb8a61e4890d60a161/jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java#L179)

jpstotz:
Hopefully this is the final PR to close #1156.

Based on my proposal in https://github.com/skylot/jadx/issues/1156#issuecomment-825660907 I improved decoding of flag attributes.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
@@ -1,9 +1,13 @@
 package jadx.core.xmlgen;
 
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


skylot (approved):

skylot:
@jpstotz thanks!

PR #1362, opened 2022-02-02
Add log messages in initialization phase

yotamN:
I've had to resort to some print debugging because of a bug that occurred only outside Intellij, I've added a few logs that I think could be useful to others as well.

One thing to note: I don't think I saw the loaded input files message even in debug mode, not sure if it's a bug with the verbose flag

### DIFF
--- a/jadx-cli/src/main/java/jadx/cli/LogHelper.java
+++ b/jadx-cli/src/main/java/jadx/cli/LogHelper.java
@@ -54,7 +54,7 @@ public class LogHelper {
 		Logger rootLogger = (Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
 		rootLogger.setLevel(logLevel.getLevel());
 
-		if (logLevel != LogLevelEnum.QUIET) {
+		if (logLevel == LogLevelEnum.PROGRESS) {
 			// show progress for all levels except quiet
 			setLevelForClass(JadxCLI.class, Level.INFO);
 			setLevelForClass(JadxDecompiler.class, Level.INFO);
--- a/jadx-core/src/main/java/jadx/api/JadxDecompiler.java
+++ b/jadx-core/src/main/java/jadx/api/JadxDecompiler.java
@@ -128,6 +128,7 @@ public final class JadxDecompiler implements Closeable {
 				loadedInputs.add(loadResult);
 			}
 		}
+		LOG.debug("Loaded using {} inputs plugin", loadedInputs.size());
 	}
 
 	private void reset() {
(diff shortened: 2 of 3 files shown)


skylot:
> I don't think I saw the loaded input files message even in debug mode, not sure if it's a bug with the verbose flag

Oh, this is actually a bug, looks like I messed up at https://github.com/skylot/jadx/blob/dd29d37154d34c0961dbe604cee8b28a72996029/jadx-cli/src/main/java/jadx/cli/LogHelper.java#L57
here should be 
```java
if (logLevel == LogLevelEnum.PROGRESS) {
```

And, after that change, we can see logs from JadxDecompiler class in verbose:
```
INFO  - loading ...
DEBUG - Loading plugin: java-convert
DEBUG - Loading plugin: smali-input
DEBUG - Loading plugin: dex-input
DEBUG - Loading plugin: java-input
INFO  - Resolved 3 plugins
DEBUG - Resolved plugins: [dex-input, java-convert, smali-input]
DEBUG - Loading dex: classes2.dex
DEBUG - Loading dex: classes3.dex
DEBUG - Loading dex: classes.dex
DEBUG - Loaded input files with dex-input
INFO  - Loaded 1 inputs
```
Here are my thoughts:
- `Loading plugin:` - is fine 
- `Resolved 3 plugins` - can be omitted because next line already show that info
- `Loaded input files with dex-input` - not really needed because `dex-input` already printed what was done
- `Loaded 1 inputs` - a little confusing, maybe `Loaded from 1 inputs plugin` will be better (counter for non-empty load results)
- `debug` level should be used for all such logs

@yotamN please apply my fix and recheck your changes.
I am fine with more verbose output, but it should be clear and don't contain duplicates :)

> I've had to resort to some print debugging because of a bug

What kind of bug? Are you trying to develop your own plugin? I am just curious 😄

yotamN:
I fixed everything you said and added a fix to show the logs from `JadxDecompiler`.  Although I still think that the message `Loaded input files with` might be useful because it doesn't assume that the plugin itself have a log message about itself.

> What kind of bug? Are you trying to develop your own plugin? I am just curious smile

I'm using Jadx nightly as a library to decompile some APKs so I built it myself and when I worked on it from Intellij it worked great but when I built it myself as an uberjar and run it inside Flatpak it found 0 classes because all the plugins except one were missing. I'm still not sure why the `ServiceLoader` failed to find it but for now I manually load the plugins.

If I could manage to extract a reproducible example I will open an issue.

skylot:
> because all the plugins except one were missing

This can happen if plugin services files were overridden (same file with same path in all plugins).
Shadow jar gradle plugin should be configured like this:
```groovy
shadowJar {
	mergeServiceFiles()
}
```

> Although I still think that the message `Loaded input files with` might be useful

Well, you can add it, but in your case this will not help, because it will not appear.

yotamN:
> This can happen if plugin services files were overridden (same file with same path in all plugins).

I will try it, thanks!

> Well, you can add it, but in your case this will not help, because it will not appear.

Why will it not appear?

skylot:
> Why will it not appear?

Incorrect plugin will return empty `ILoadResult` so nothing will be added to `loadedInputs`

yotamN:
Fair, so I think you can merge it as is if it's okay by you

skylot (approved):

Issue #1104, opened 2021-01-27
[feature] local variable cross-references in random order


Surendrajat:
*Describe your idea:*
After #1103, Jadx can show cross-refs(usage) for local variables. However, those references show up in random order in the usage dialog(see below) which is not a big deal in itself, but it may be improved by correcting the refs order.

https://user-images.githubusercontent.com/11215000/106024508-dc089e00-60ed-11eb-9af7-e8e885b46986.mp4

PR #1363, opened 2022-02-05
fix(gui): The result in usage search should sort by name

cyqw:
The result in usage search is sort by CodeNode::COMPARATOR, it's not suitable for human reading.


skylot:
@cyqw I see what kind of issues you are trying to fix, but it is better just to fix that comparator. And only name is not enough here because, for example, variable usage needs to be sorted by usage place not name (check #1104).
So, I suggest fixing [CodeNode comparator](https://github.com/skylot/jadx/blob/f37005958fed88bde7b5e362333104940b64467b/jadx-gui/src/main/java/jadx/gui/treemodel/CodeNode.java#L112-L114) to sort by `long name` which used in first column renderer ([here](https://github.com/skylot/jadx/blob/6e8baef9b2f9e3f8697ca07df2e3ca7549513325/jadx-gui/src/main/java/jadx/gui/ui/dialog/CommonSearchDialog.java#L507)) to:
```java
	private static final Comparator<CodeNode> COMPARATOR = Comparator
			.comparing(CodeNode::makeLongString)
			.thenComparingInt(CodeNode::getPos);
```
Please try to apply this change and check if it fixed your issue. If still not, share a screenshot of case which you want to improve.

skylot (changes requested):
Check my comment on this PR

cyqw:
Fixed CodeNode comparater.

skylot (approved):

Issue #1365, opened 2022-02-05
[feature] Use the raw names of classes when creating Frida snippets


oriori1703:
Use the raw name from the apk instead of the renamed names of classes when creating a Frida snippet, so the hook would actually work.

skylot:
Fixed in PR #1366 by @oriori1703

PR #1366, opened 2022-02-05
Fix/frida raw class names

oriori1703:
As described in #1365 I changed the Frida snippet to use the raw name of the class instead of the renamed name.
I also cleaned the code a bit, and added the ability to generate a snippet for accessing fields.

BTW, I noticed a very strange behavior when trying to get the top parent of an inlined class.
I noticed that when I do `javaMethod.getTopParentClass()` I get the parent of the class after it was inlined.
However when I tried to `methodNode().getTopParentClass()` for the same method I got the inner class.
Is this the intended behavior? If not I will open an issue with an example.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
@@ -17,7 +17,6 @@ import org.slf4j.LoggerFactory;
 import jadx.api.data.annotations.VarDeclareRef;
 import jadx.core.dex.info.MethodInfo;
 import jadx.core.dex.instructions.args.ArgType;
-import jadx.core.dex.nodes.ClassNode;
 import jadx.core.dex.nodes.MethodNode;
 import jadx.gui.treemodel.JClass;
 import jadx.gui.treemodel.JMethod;
@@ -30,7 +29,6 @@ public final class FridaAction extends JNodeMenuAction<JNode> {
 	private static final Logger LOG = LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
 	private static final long serialVersionUID = 4692546569977976384L;
 	private final Map<String, Boolean> isInitial = new HashMap<>();
-	private String methodName;
 
 	public FridaAction(CodeArea codeArea) {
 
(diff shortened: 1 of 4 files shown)


jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
```diff
@@ -32,157 +32,134 @@
...
+
+	private String generateMethodSnippet(JMethod jMth) {
+		JavaMethod javaMethod = jMth.getJavaMethod();
+		MethodInfo methodInfo = javaMethod.getMethodNode().getMethodInfo();
+		String methodName = methodInfo.getName();
+		if (methodName.equals("<init>") || methodName.equals("onCreate")) {
+			methodName = "$init";
+		}
+		String rawClassName = javaMethod.getDeclaringClass().getRawName();
+		String shortClassName = javaMethod.getDeclaringClass().getName();
```
skylot, line 81:
```suggestion
		String shortClassName = javaMethod.getDeclaringClass().getClassNode().getClassInfo().getShortName();
```
Is here should be raw short name?
If yes, maybe it is better to add method `getShortRawName()` to `JavaClass` class.

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
```diff
@@ -32,157 +32,134 @@
...
+
+		String functionUntilImplementation;
+		if (!javaMethod.getMethodNode().getOverloads().isEmpty()) {
```
skylot, line 84:
```suggestion
		if (javaMethod.getMethodNode().isArgsOverloaded()) {
```
Will this method work here? If not and you still need yours `getOverloads()` method, move it to this class, because java definition of overloaded methods is slightly different (i.e. you don't check args count and inherited methods).

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
```diff
@@ -32,157 +32,134 @@
 public final class FridaAction extends JNodeMenuAction<JNode> {
 	private static final Logger LOG = LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
```
skylot, line 33:
```suggestion
	private static final Logger LOG = LoggerFactory.getLogger(FridaAction.class);
```

skylot (changes requested):

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
oriori1703: I think it's better to keep `javaMethod.getDeclaringClass().getName()` because if the class has an alias it returns it instead of the short name.
This way we can use any renaming the user did for the variable name that holds the class, while still referring to the actual class in the arguments to the Frida functions.
For example if I renamed a class, and then generate a snippet I would get something like this: `let renameClass = Java.use("com.example.rawClass")`

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
oriori1703: I was under the impression that overloads are any method in the same class that has the same name, no matter how much parameters it has, or their types.
So this would be considered an example of overloading:
```java
class Addition {
    public int add(int a, int b)
    {
        int sum = a + b;
        return sum;
    }

    public int add(int a, int b, int c)
    {
        int sum = a + b + c;
        return sum;
    }
}
```
I'm pretty sure that Frida uses this definition when it requires specifying the overload (to know which method, from all of the method with the same name, to call).
I will move the method to this class if you still think it doesn't belong in  `MethodNode`.

I'm not sure what Frida does when there is also an inherited method with the same name. I think it's fine to leave the logic like this and add more sophisticated logic for overloading later, if it's needed.

skylot: Well, I tried to check [java specs](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.4.9)... and it is very cryptic, so I am not sure about args count match, but examples show inheritance. So it is better to check that on sample code.
Anyway, in jadx I use overload with a same args count, and it will be confusing to keep different implementations for same thing 😕

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
skylot: Ok, that's fine

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
oriori1703: I checked, and Frida only requires `.overload()` when there are methods with the same name in the same class.
It doesn't care about inheritance, but it does care about other methods with different types of arguments.

I moved the overload check to this class to avoid the confusion with the other overload method 😄

skylot (approved):

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
skylot: Great! Thanks!

skylot:
@oriori1703 thank you for you work!

BTW
>  I noticed a very strange behavior when trying to get the top parent of an inlined class ...

Sure, please share details and sample in new issue 👍🏽

PR #1369, opened 2022-02-06
chore: do not deobfuscate known top  level domains

jpstotz:
Do not deobfuscate known top  level domains with 2 or 3 characters:

I noticed that common packages like `ch.qos.logback`, `de.greenrobot`, `io.jsonwebtoken`... are always renamed because of length of their first level package name. As they are known top level domains I added a check to exclude all current tlds that have 2 or  3 characters.


skylot:
@jpstotz great idea!
One thing: I think it will be better to init that list only once, so please move that parsing to a separate class and to `private static final` field with a static method for check. Also add a small doc to that class, because `tld` is a very short name 😄

jpstotz:
@skylot As deobfuscation is optional and only used in a short period while loading I thought it is not necessary to keep the tld list in memory all the time. But as the list is very short it does not really make a difference, so I have adapted my PR.

skylot:
@jpstotz thanks!
Actually, class will not be loaded until first use, also in jadx-gui full reload happen on every settings change, so improve performance a little is always nice.

Issue #1367, opened 2022-02-06
[core] Gradle export fails if targetSdkVersion is not set


nitram84:
If targetSdkVersion is not set in Manifest.xml a grade export fails with:

```
ERROR` - jadx error: For input string: ""
java.lang.NumberFormatException: For input string: ""
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Integer.parseInt(Integer.java:592)
        at java.lang.Integer.valueOf(Integer.java:766)
        at jadx.core.export.ExportGradleProject.getApplicationParams(ExportGradleProject.java:113)
        at jadx.core.export.ExportGradleProject.<init>(ExportGradleProject.java:49)
        at jadx.api.JadxDecompiler.getSaveTasks(JadxDecompiler.java:266)
        at jadx.api.JadxDecompiler.getSaveExecutor(JadxDecompiler.java:238)
        at jadx.api.JadxDecompiler.getSaveExecutor(JadxDecompiler.java:226)
        at jadx.api.JadxDecompiler.save(JadxDecompiler.java:191)
        at jadx.cli.JadxCLI.processAndSave(JadxCLI.java:48)
        at jadx.cli.JadxCLI.execute(JadxCLI.java:35)
        at jadx.cli.JadxCLI.main(JadxCLI.java:19)
```

A sample APK can be found here https://mahjong-solitaire-free2.apk.gold/ (version 1.0.0.9).

TargetSdkVersion is optional. If not set, the default value equals that given to minSdkVersion (https://developer.android.com/guide/topics/manifest/uses-sdk-element#target).

The fix is a one-liner and can be found in my branch: https://github.com/nitram84/jadx/tree/export_optional_targetSdkVersion
with a first unit test for gradle exports. I can create a PR if you want.

skylot:
@nitram84 your commit looks great! Please submit a PR! 👍🏽

skylot:
Fixed in PR #1370 by @nitram84

PR #1370, opened 2022-02-06
Fix gradle export if targetSdkVersion is missing (#1367)

nitram84:
Here is my proposed fix for issue #1367.

Unfortunately JadxDecompiler was not directly mockable to load raw xml files, so I needed some additional code, to mock it without modifying it.
I hope the testing code for gradle exports can be improved and extended when we have more cases.


skylot (approved):

skylot:
@nitram84 thanks! Great work!

PR #1414, opened 2022-03-20
Smali debugger/Adb io read problems, proper socket closing, ... 

jpstotz:
There were some more IO read problems, also the socket handling is better with ry-with-resource.

I also started to refactor the large classes `ADB` and `SmaliDebugger` by extracting some inner classes.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/device/protocol/ADB.java
+++ b/jadx-gui/src/main/java/jadx/gui/device/protocol/ADB.java
@@ -80,7 +80,7 @@ public class ADB {
 			return IOUtils.readNBytes(stream, len);
 		} catch (IOException e) {
 			LOG.error("Failed to read readServiceProtocol: {}", e.toString());
-			return new byte[0];
+			return null;
 		}
 	}
 
--- a/jadx-gui/src/main/java/jadx/gui/device/protocol/ADB.java
+++ b/jadx-gui/src/main/java/jadx/gui/device/protocol/ADB.java
@@ -1,5 +1,6 @@
 package jadx.gui.device.protocol;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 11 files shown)


jadx-gui/src/main/java/jadx/gui/device/protocol/ADB.java
```diff
@@ -148,17 +135,18 @@ public static boolean startServer(String adbPath, int port) throws IOException {
...
-		is.read(bytes, 0, size);
-		return new String(bytes).contains(tcpPort);
+		ByteArrayOutputStream out = new ByteArrayOutputStream();
+		try (InputStream in = proc.getInputStream()) {
+			int read;
+			byte[] buf = new byte[1024];
+			while ((read = in.read(buf)) >= 0) {
+			}
+		}
+		return new String(out.toByteArray()).contains(tcpPort);
```
skylot, line 145: Write to `out` variable is missing here

skylot (changes requested):

jadx-gui/src/main/java/jadx/gui/device/protocol/ADB.java
jpstotz: sorry stupid bug. Seems like I am getting laze as usually I use `InputStream.getAllBytes()`

jadx-gui/src/main/java/jadx/gui/device/protocol/ADB.java
```diff
@@ -73,29 +70,21 @@ private static boolean execCommandAsync(OutputStream outputStream,
...
+			byte[] buf = IOUtils.readNBytes(stream, 4);
+			int len = unhex(buf);
+			if (len == 0) {
+				return new byte[0];
 			}
-		} catch (IOException ignore) {
+			return IOUtils.readNBytes(stream, len);
+		} catch (IOException e) {
+			LOG.error("Failed to read readServiceProtocol: {}", e.toString());
+			return new byte[0];
```
skylot, line 83:
```suggestion
			return null;
```
This fix endless `Failed to read readServiceProtocol: java.net.SocketException: Socket closed` error

skylot (approved):

PR #1426, opened 2022-03-27
add appdata.xml

Bleuzen:
adds appdata.xml with some metadata about this project, needed for flathub

### DIFF
--- a/appdata.xml
+++ b/appdata.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+    <id>com.github.skylot.jadx</id>
+    <metadata_license>CC0-1.0</metadata_license>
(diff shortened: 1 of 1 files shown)


Bleuzen:
note: this file also has to be included in the next binary/release zip
(I currently can't build from source/git repo because internet access during build time is not allowed (fails downloading dependencies))

skylot (approved):

Bleuzen:
damn
was just about to write that there is a problem and it shouldn't be merged yet but just in that minute you merged :D

From build log:
```
Error loading AppData file: failed to parse /app/share/appdata/com.github.skylot.jadx.appdata.xml: Error on line 14 char 87: <li> already set 'decode' and tried to replace with ' and other resources from '
```

maybe we have to remove the <code> tags

skylot:
@Bleuzen You can open a new PR 🤣

Issue #1427, opened 2022-03-27
distribute as flatpak on flathub.org


Bleuzen:
Hi, I currently package jadx as flatpak to be available on flathub.org.
You can see the current progress in this pr: https://github.com/flathub/flathub/pull/2942

1) Are you fine with distributing jadx on flathub?
2) If so, this still needs some metadata. The appstream.xml should be included in this repo, see: https://github.com/skylot/jadx/pull/1426

skylot:
> Are you fine with distributing jadx on flathub?

Sure, that will be great!

skylot:
@Bleuzen I have a small question: you mention that I will need to include `appdata.xml` in release zip, does anything else needs to done on release? Maybe you can link some docs about new release process?

Bleuzen:
@skylot 

> Maybe you can link some docs about new release process?

Here is the spec:
https://www.freedesktop.org/software/appstream/docs/

> does anything else needs to be done on release?

Yes. In the metadata file there is a releases node, where I already added version 1.3.4.
See: 
https://github.com/skylot/jadx/blob/e5fa818b5c370145b28626918d64c154a67bf633/appdata.xml#L39

Before a new version is released, you should add the new release there.

You _could_ even add a changelog / release description there (in `<description>` tag within a `<release>`).
As an example look at the file here:
https://github.com/flathub/com.github.hopsan.Hopsan/blob/2d9c0c698a5cb8109e351a75a5948a18c4d2c3b3/com.github.hopsan.Hopsan.appdata.xml

Also there are other attributes available, see:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases

However this is optional.
Minimum is that at least one release with version and date is provided.
See: https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html

> The application metainfo should at least provide one <releases/> tag, which has one or more <release/> children to define the version and release date of this application.

So if you don't care about version history and changelog to show up in software stores, you can always only have one `<release>` tag in there and just use a text replace tool like `sed` to put in the newest version and date of its release.

How much you want to provide is up to you, as long as at least the newest release with version and date is in there.

PR #1430, opened 2022-03-31
Update appdata.xml

Bleuzen:
Don't count my PRs for this, I know exactly what I'm doing.
learned: appstream is a little picky princess

apparently she didn't like the taste of `<strong>` nodes

`<ul>`'s had to be moved outside of `<p>` or would just silently not render

Also removed the "Main features:" and "jadx-gui features:" texts before the lists and merged them, because "Main features:" is too short. You can't have short texts above lists, this is a very important reason to fail, world is going down otherwise.

```
builddir/files/share/appdata/com.github.skylot.jadx.appdata.xml: FAILED:
• style-invalid         : Content before <ul> is too short [18], at least 20 characters required
```

btw. for the future: you can check those appdata.xml files with `appstream-util validate <xml-file>`.
But you need to have luck that it does output all errors and not just say it is fine while rendering "(null)" in software stores.

whatever. You can see the current xml in action here:
https://flathub.org/apps/details/com.github.skylot.jadx

Should be fine now. If there are wishes from your side, tell me, I can test them on flathub before next release is made.
Also be careful if you want to edit the description in the future.

---

Fixes https://github.com/skylot/jadx/issues/1427#issuecomment-1083144197

### DIFF
--- a/appdata.xml
+++ b/appdata.xml
@@ -7,24 +7,16 @@
     <summary>Dex to Java decompiler</summary>
     <description>
         <p>Command line and GUI tools for producing Java source code from Android Dex and Apk files</p>
-        <p>
-            <strong>Main features:</strong>
-            <ul>
-                <li>decompile Dalvik bytecode to java classes from APK, dex, aar, aab and zip files</li>
-                <li>decode AndroidManifest.xml and other resources from resources.arsc</li>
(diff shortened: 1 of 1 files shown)


skylot:
@Bleuzen thanks!

Issue #1434, opened 2022-04-04
Impossible to set the option "--decompilation-mode"


boberle:
Throws an exception telling that the values `auto`, `simple`, etc. (the ones in the help and readme file) are not valid. Instead, it tells me that possible values are those of another option (`--rename-flags`).

```
ERROR - Process error:
java.lang.IllegalArgumentException: 'auto' is unknown for parameter --decompilation-mode, possible values are case, valid, printable
        at jadx.cli.JadxCLIArgs$RenameConverter.convert(JadxCLIArgs.java:470)
        at jadx.cli.JadxCLIArgs$RenameConverter.convert(JadxCLIArgs.java:448)
        at com.beust.jcommander.JCommander.convertValue(JCommander.java:1340)
        at com.beust.jcommander.ParameterDescription.addValue(ParameterDescription.java:249)
        at com.beust.jcommander.JCommander.processFixedArity(JCommander.java:920)
        at com.beust.jcommander.JCommander.processFixedArity(JCommander.java:901)
        at com.beust.jcommander.JCommander.parseValues(JCommander.java:731)
        at com.beust.jcommander.JCommander.parse(JCommander.java:363)
        at com.beust.jcommander.JCommander.parse(JCommander.java:342)
        at jadx.cli.JCommanderWrapper.parse(JCommanderWrapper.java:35)
        at jadx.cli.JadxCLIArgs.processArgs(JadxCLIArgs.java:200)
        at jadx.cli.JadxCLI.execute(JadxCLI.java:35)
        at jadx.cli.JadxCLI.main(JadxCLI.java:20)
```

Same for options `-m` and `decompilation-mode`.

Appears since commit d8306cb1c09698886f21b1406c6c751f368dc5f1.

boberle:
Need to change the file `jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java`, line 69, replace:

```
converter = RenameConverter.class
```

by:

```
converter = DecompilationModeConverter.class
```

PR #1435, opened 2022-04-04
fix: bug impossible to set the option "--decompilation-mode"

boberle:
Correction of issue #1434, where it was not possible to set the option `--decompilation-mode`.


skylot (approved):

skylot:
@boberle thanks! That was a really stupid copy/paste mistake :disappointed:

Issue #751, opened 2019-09-11
Error decode: resources.arsc (Facebook messenger)


jpstotz:
Jadx has totally fails to decode the resources in `resources.arsc` of [Facebook messenger 231.0.0.25.121 (170889108)](https://apkpure.com/de/facebook-messenger/com.facebook.orca/download/170889108-APK?from=variants%2Fversion):

```
Caused by: java.io.IOException: value size, expected: 0x8, actual: 0xa, offset: 0x13ef0e
	at jadx.core.xmlgen.ParserStream.throwException(ParserStream.java:118)
	at jadx.core.xmlgen.ParserStream.checkInt16(ParserStream.java:110)
	at jadx.core.xmlgen.ResTableParser.parseValue(ResTableParser.java:265)
	at jadx.core.xmlgen.ResTableParser.parseEntry(ResTableParser.java:254)
	at jadx.core.xmlgen.ResTableParser.parseTypeChunk(ResTableParser.java:216)
	at jadx.core.xmlgen.ResTableParser.parsePackage(ResTableParser.java:169)
	at jadx.core.xmlgen.ResTableParser.decodeTableChunk(ResTableParser.java:118)
	at jadx.core.xmlgen.ResTableParser.decode(ResTableParser.java:66)

```

According to Jadx the problematic entry belongs to resource  `0x7f090012 (2131296274)[DEFAULT] = id.browser_le_progress_bar`. I tries to compare those values to the output of `aapt.exe dump --values resources facebookMessenger.apk` but I was not able to identify anything unusual regarding this resource in the aapt dump.

root-intruder:
@jpstotz I just ran into the exact same error and wanted to open a ticket...problem still exists with newest jadx version and latest facebook messenger from apkpure as of today.

jpstotz:
Some times leaving a bug some time alone is helpful. This time, checking the resources decoding code again I discovered the cause:

It seems like Facebook messenger uses an obfuscation technique that causes this problem. But the actual problem is not that there is a new unknown type we don't know how to parse, instead the obfuscator manipulates the resource chunk in a way that it defines more entries that fit into the chunk. It seems like aapt and other resource parsing tools just stop when they reach the end of the chunk and ignore the remaining entries. 

In Jadx we ignored the chunk size and read the next data block and tried to interpret it the wrong way, which of course totally fails. That is the exception we see in Jadx.

I will prepare a PR after some more tests.

root-intruder:
[...]

PR #1436, opened 2022-04-04
Fix resource chunk loading (#751)

jpstotz:
Ignore resource chunk entries that are located after the resource chunk end (fix for #751).

I tested loading of a few dozen apps into Jadx and the new code part was never triggered except for the Facebook messenger apps. So with a good chance there should be no new problems introduced by this PR.


skylot (approved):

skylot:
Thank you, @jpstotz :+1:

PR #1467, opened 2022-04-25
feat: Add option to include/exclude dependencies when saving

### LINKED ISSUE #1466: [feature] Add "SkipDependencies" argument to not save dependencies to disk
I'm doing research on static analysis of decompiled code, and I'd like to only save the source files that exactly match the class filter.
Right now, when saving, using a filter for "com.foo" for example, the output folder will include the matching files plus dependencies ("com.google", etc)

Something like "setSkipDependencies" and "isSkipDependencies" would do the trick.
I already implemented this change, so if this is useful to others, I can make a PR.

The tool itself has been quite helpful so far. Thanks :)

NASGregorio:
By default, when a class filter is applied, only the matching files will be saved.
Alternatively, "setIncludeDependencies(true)" can be used to also have them saved.

Fix #1466

### DIFF
--- a/jadx-core/src/main/java/jadx/api/JadxArgs.java
+++ b/jadx-core/src/main/java/jadx/api/JadxArgs.java
@@ -49,6 +49,7 @@ public class JadxArgs {
 
 	private boolean skipResources = false;
 	private boolean skipSources = false;
+	private boolean includeDependencies = false;
 
 	/**
 	 * Predicate that allows to filter the classes to be process based on their full name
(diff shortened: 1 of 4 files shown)


skylot:
@NASGregorio you fix is working, but you just disable all work of `DecompilerScheduler` :smiling_face_with_tear:
So I add a commit to not save excluded classes by `classFilter` while processing dependencies as needed.
Please check these changes for your test case.

NASGregorio:
@skylot Sorry for the trouble.
I just tested it and it has the expected behaviour 👍

Issue #1456, opened 2022-04-21
java.lang.IndexOutOfBoundsException: Index: -2, Tab count: 29


betteray:
Please describe what you did before the error occurred.
**IMPORTANT!** If the error occurs with a specific APK file please attach or provide link to apk file!
 * Jadx version: 1.3.4
* Java version: 18
* Java VM: Homebrew OpenJDK 64-Bit Server VM
* Platform: Mac OS X (10.15.6 x86_64)
* Max heap size: 8192 MB
* Program args: -Xms128M -Xmx8g -Dawt.useSystemAAFontSettings=lcd -Dswing.aatext=true -XX:+UseG1GC

```
java.lang.IndexOutOfBoundsException: Index: -2, Tab count: 29
	at java.desktop/javax.swing.JTabbedPane.checkIndex(JTabbedPane.java:1794)
	at java.desktop/javax.swing.JTabbedPane.setSelectedIndex(JTabbedPane.java:622)
	at jadx.gui.ui.TabbedPane.lambda$new$0(TabbedPane.java:69)
	at java.desktop/java.awt.AWTEventMulticaster.mouseWheelMoved(AWTEventMulticaster.java:539)
	at java.desktop/java.awt.Component.processMouseWheelEvent(Component.java:6704)
	at java.desktop/java.awt.Component.processEvent(Component.java:6388)
	at java.desktop/java.awt.Container.processEvent(Container.java:2266)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:4991)
	at java.desktop/java.awt.Container.dispatchEventToSelf(Container.java:2348)
	at java.desktop/java.awt.Component.dispatchMouseWheelToAncestor(Component.java:5136)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:4888)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2324)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4823)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4948)
	at java.desktop/java.awt.LightweightDispatcher.processMouseEvent(Container.java:4604)
	at java.desktop/java.awt.LightweightDispatcher.dispatchEvent(Container.java:4516)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2310)
	at java.desktop/java.awt.Window.dispatchEventImpl(Window.java:2780)
[...]

PR #1469, opened 2022-04-26
fix(gui): fix IndexOutOfBoundsException

jpstotz:
fix IndexOutOfBoundsException when switching between tabs via mouse wheel
Fix for #1456


skylot (approved):

skylot:
@jpstotz great! Thank you :+1:

PR #1505, opened 2022-05-30
Support for exporting to proper deobfuscation mapping file formats

### LINKED ISSUE #1491: Proper deobfuscation mapping file format
As far as I could see, you currently have two options for exporting user-made mappings (renames): the `.jadx` project file and `.jobf` mapping files. The issue is, `.jadx` contains much other information that's not needed/wanted when I just want to get the mappings. And `.jobf` can neither handle Javadoc/comments nor is it easy to get them generated.
I suggest/request looking into adding `.enigma` or `.tiny` mapping format support, which are both currently used by FabricMC in their Minecraft deobfuscation toolchain. Enigma files look like this, for example: https://github.com/FabricMC/yarn/blob/1.19-pre2/mappings/net/minecraft/block/Block.mapping

It should be relatively easy to implement, too, via the [`mapping-io`](https://github.com/FabricMC/mapping-io) library. Maybe support for JADX's mappings can be PR'd to there too, so conversion of legacy mapping formats would be possible?

NebelNidas:
Adds support for exporting manually renamed classes/methods/fields etc. in the Tiny v2 format, which is the most versatile out there and can handle a lot more features than JADX's default `.jobf` format. For more details, see its [spec](https://github.com/FabricMC/tiny-remapper/issues/9).

Depends on FabricMC's [`mapping-io`](https://github.com/FabricMC/mapping-io) library, which theoretically allows JADX to support many more formats in the future.
This PR doesn't add Tiny v2 import support yet, as I don't need it personally atm, but it can be easily added later when needed.

Fixes #1491.

Currently working:
- [x] Exporting classes
- [x] Exporting fields
- [x] Exporting methods
- [x] Exporting method args
- [x] Exporting method vars
- [x] Exporting comments


NebelNidas:
@skylot It's almost done now, but I need some help with the method args. How can I access them? I looked around the code, but couldn't find a solution. I mean, there's `methodInfo.getArgsCount()` and such, but I don't just need the arg count/types, but their old and renamed names, too!
I've tried adding something like this in between [here](https://github.com/skylot/jadx/blob/d14c59a962749e6f5405d14658fc09fa02d7e120/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java#L213-L214), but it obviously doesn't work (What even are `RegisterArg`s?):
```java
boolean wasLoaded = mth.isLoaded();
try {
    mth.load();
    for (RegisterArg arg : mth.getArgRegs()) {
        if (mappedMethodArgsAndVars.contains(methodInfo.getDeclClass() + methodInfo.getShortId())) {
            mappingTree.visitClass(classPath);
            mappingTree.visitMethod(methodName, methodDesc);
            mappingTree.visitMethodArg(arg.getRegNum(), 0, arg.getName());
            mappingTree.visitDstName(MappedElementKind.METHOD_ARG, 0, arg.getName());
        }
    }
} catch (DecodeException e) {
    LOG.error("Error while decompiling method " + methodInfo.getShortId());
}
if (!wasLoaded) {
    mth.unload();
}
```

NebelNidas:
Thank you! I've already gone down another route, which also works well so far, only issue is that
```java
for (RegisterArg arg : mth.getArgRegs()) {
    arg.getSVar()
}
```
always returns `null`. Is there another way to access the arg's `SSAVar`? I need it to calculate the index for a new `JadxCodeRef` so I can compare it with the existing saved one (see my code here: [JadxCodeRef](https://github.com/NebelNidas/jadx/commit/fe104a1037ed8b184fa441c1516672f61aacb527))

NebelNidas:
Thank you for the help, method args work now :) One question though: What's the reason for using `getTopParentClass()` instead of `getParentClass()` here?
https://github.com/skylot/jadx/blob/fa41874e303f3570de389c118f59444a1139e948/jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java#L118

NebelNidas:
How would I do the same for method vars? I guess I would have to replace the `> lineEndPos` check here with `> methodEndPos`, but I don't know how I can get that information (where the method ends / how big the method is):
https://github.com/skylot/jadx/blob/fa41874e303f3570de389c118f59444a1139e948/jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java#L123-L124

skylot:
I think you can stop if `mth` in `VarNode` became different, i.e. `!varNode.getMth().equals(methodNode)`.
In ` collectMethodArgs` method I use that check just to be sure, but it not really needed.

NebelNidas:
Good idea! I still need to get a few properties from the resulting `VarNode`s, and I'm not sure how. For one, I need the var's `offset` to be able to use `JadxCodeRef.forInsn(varOffset)` on it. And mapping-io requires that I provide the method var's `lv-index` and `lv-start-offset` aka `start-op-code`:
```
<lv-index> refers to the local variable array index of the frames having the variable, see "index" in JVMS SE 8 §4.7.13
<lv-start-offset> is at most the start of the range in which the variable has a value, but doesn't overlap with another variable with the same <lv-index>, see "start_pc" in JVMS SE 8 §4.7.13. The start offset/range for tiny is measured in instructions with a valid opcode, not bytes.
```
For the local variable index already have code (which I think works correctly), see here:
https://github.com/skylot/jadx/blob/235991fb9c35136f0f8489125821a5d6c8e2999b/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java#L254
https://github.com/skylot/jadx/blob/235991fb9c35136f0f8489125821a5d6c8e2999b/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java#L265
And I'm not sure if there's an easier way provided by JADX. I have no idea yet how I can get the `lv-start-offset`

skylot:
> I need the var's offset to be able to use JadxCodeRef.forInsn(varOffset) on it

Offset stored in metadata annotation with type `OFFSET` and attached to line start, so offset on line before variable declaration is what you need.

> method var's lv-index and lv-start-offset

Variables array can be acquired using `MethodNode.getDebugInfo().getLocalVars()` next use `getStartOffset` for lv-start-offset and `getRegNum() - argsStartReg` for lv-index (you can add getter for argsStartReg field in MethodNode). But be aware that `getDebugInfo()` can return null if debug info is not available.
Also, if you want to check the code, you can start at reader of vars attributes [here](https://github.com/skylot/jadx/blob/1efdcd7b10fe2b3984db743a4deec0b472e0e2ce/jadx-plugins/jadx-java-input/src/main/java/jadx/plugins/input/java/data/attributes/debuginfo/LocalVarsAttr.java#L27-L37).

NebelNidas:
Can I use the `localVariable.getStartOffset()` directly instead of looping through the class's code metadata looking for the `OFFSET` annotation?

skylot:
> Can I use the localVariable.getStartOffset() directly

You can, but I am not sure how you will match it with `VarNode`, because there are can be several variables with same `reg num`. Anyway, looks like your format can't handle such cases, so we can't do much here.

NebelNidas:
> use [...] `getRegNum() - argsStartReg` for lv-index

Are you sure? I have this static method:
image

And since it's static, the local variable index should start at zero, meaning that the lv-index for `key` would be 0 and for `value` 1. Using your suggested code:
```java
int lvIndex = varNode.getReg() - methodNode.getArgsStartReg();
```
I get this:
image

Which I don't think is correct, is it? 🤔 

---
> there are can be several variables with same `reg num`

In which case?

skylot:
Well, I can be miss something, that is why I give you a link to a code jadx using to read these values :slightly_smiling_face:
And you can check bytecode and docs for api you are using.
Also, do you know any other tools to work with this format, can you compare or apply/check your output?

> In which case?

I don't know, but this can happen.

NebelNidas:
[Enigma](https://github.com/FabricMC/Enigma) ([download](https://maven.fabricmc.net/cuchaz/enigma-swing/2.1.0/)) can do everything except method vars. Using it on the same app as JADX (after running it through dex2jar) and doing the same renames, I get the expected results:
image

NebelNidas:
> > In which case?

> I don't know, but this can happen.

Tiny v2 allows specifying an optional `lvt-index`:
```
<lvt-index> is the index into the LocalVariableTable attribute's local_variable_table array, see "local_variable_table" in JVMS SE 8 §4.7.13, not to be confused with "index" referred by <lv-index>
```
Would this help in such edge cases?

skylot:
> I get the expected results

Wait, so you are trying to use format bound to java bytecode specific internals (check your lv-index doc) for dalvik bytecode? I think conversion can break that part.

> Tiny v2 allows specifying an optional lvt-index:

I don't think this will help. Jadx can generate different from original code and use more variables, so it is expected that not all of them can be traced back to original.

NebelNidas:
> Wait, so you are trying to use format bound to java bytecode specific internals (check your lv-index doc) for dalvik bytecode?

I know Java and Dalvik bytecode have a fair share of differences, but doesn't stuff like this carry over? 🤔 Guess I'll have to read the docs more thoroughly (or do you already know for sure there are differences?)

> I don't think this will help. Jadx can generate different from original code and use more variables

Wouldn't this be a problem for JADX itself too? When I have old save files and the Java decompiler changes with new versions

skylot:
> or do you already know for sure there are differences?

Stack vs registers? Jadx actually convert stack to registers, that is why there is a minus part in `getRegNum() - argsStartReg` to undo that conversion (check [this](https://github.com/skylot/jadx/blob/1ce3fc972a31e06f82e4928006abe98cb9e46d3d/jadx-plugins/jadx-java-input/src/main/java/jadx/plugins/input/java/data/code/JavaCodeReader.java#L153) line)

> Wouldn't this be a problem for JADX itself too? 

It can, but ssa versions tied to CFG, and it needs a big change in jadx to break this.

> [Enigma](https://github.com/FabricMC/Enigma) can do everything

Thanks! I just tried Enigma, and it is very nice, why you ever need a jadx? :rofl:

NebelNidas:
> Stack vs registers? Jadx actually convert stack to registers, that is why there is a minus part in getRegNum() - argsStartReg to undo that conversion (check [this](https://github.com/skylot/jadx/blob/1ce3fc972a31e06f82e4928006abe98cb9e46d3d/jadx-plugins/jadx-java-input/src/main/java/jadx/plugins/input/java/data/code/JavaCodeReader.java#L153) line)

Sorry if I'm a bit dumb here; I'm still no bytecode expert by any means. I didn't really get what's the difference between Dalvik and Java bytecode in this specific case, since we're just counting relative positions. In both Java and Dalvik, longs and doubles take up two registers, otherwise one:

https://github.com/skylot/jadx/blob/0c3afcc24cdc911667f373f3753223796cd46539/jadx-core/src/main/java/jadx/core/dex/instructions/args/ArgType.java#L775-L783

I fortunately stumbled upon this wiki page which helped me understand the situation: https://github.com/JesusFreke/smali/wiki/Registers

So basically, for method arguments, Java bytecode starts counting at zero (for static methods) or one, and then counts up. Dalvik on the other hand starts at `availableRegisters - argCount` and always counts up all the way to `availableRegisters`.  This means conversion is still easy, and we could easily use the normal Tiny v2 format (according to Java conventions), so we 1) comply with the spec and 2) have the file more readable (every sane person would start counting at 0 or 1, not some arbitrary number). JADX can convert it back really fast when importing.
Method vars on the other hand are, as you said, quite suboptimal to implement, and frankly, I guess it's not worth the effort (at least for now).


>  just tried Enigma, and it is very nice, why you ever need a jadx? 🤣

Mostly because Enigma can't handle Dalvik bytecode, and I thought it's easier to add proper mapping support to JADX than to rebuild Enigma for work with APKs 😅 Also, JADX can do Smali debugging, which is needed for reverse engineering the YouTube app (over at @ReVanced)

NebelNidas:
> Offset stored in metadata annotation with type OFFSET and attached to line start, so offset on line before variable declaration is what you need.

If I understand it correctly, this code should work, right?
```java
private List<SimpleEntry<VarNode, Integer>> collectMethodVars(MethodNode methodNode) {
	ICodeInfo codeInfo = methodNode.getTopParentClass().getCode();
	int mthDefPos = methodNode.getDefPosition();

	List<SimpleEntry<VarNode, Integer>> vars = new ArrayList<>();
	AtomicInteger lastOffset = new AtomicInteger(-1);
	codeInfo.getCodeMetadata().searchDown(mthDefPos, (pos, ann) -> {
		LOG.info("Annotation: " + ann);
		if (ann instanceof InsnCodeOffset) {
			LOG.info("Annotation is CodeOffsetAnnotation");
			lastOffset.set(((InsnCodeOffset) ann).getOffset());
		}
		if (ann instanceof NodeDeclareRef) {
			ICodeNodeRef declRef = ((NodeDeclareRef) ann).getNode();
			if (declRef instanceof VarNode) {
				VarNode varNode = (VarNode) declRef;
				if (!varNode.getMth().equals(methodNode)) {
					// Stop if we've gone too far and have entered a different method
					return Boolean.TRUE;
				}
				if (lastOffset.get() == -1) {
					LOG.error("Couldn't determine the offset for a local variable, skipping");
					return Boolean.TRUE;
				}
				vars.add(new SimpleEntry<VarNode, Integer>(varNode, lastOffset.get()));
			}
		}
		return null;
	});
	return vars;
}
```
However, it always errors out with "Couldn't determine the offset for a local variable, skipping"

skylot:
> do you know a tool/library for remapping an entire APK?

Sorry, I don't know. Maybe only [Threadtear](https://github.com/GraxCode/threadtear) (description mention that android support is coming soon, but still not). 

> I've also noticed that renamed elements do not carry over to the Smali code, would this be possible to implement eventually?

You mean in jadx? Current "plain" smali is shown using smali library from original files, so it doesn't have renames. But there is also a `Show dalvik bytecode` version (checkbox in popup menu) which is build on top of jadx IR, so it can show a renames.

NebelNidas:
Okay, we're approaching the finish line! Only one question and one issue left.
First, I noticed a discrepancy in this part:
https://github.com/skylot/jadx/blob/14fd88b2f8e3a18fe2bfb7a9cdece3451341a029/jadx-core/src/main/java/jadx/core/dex/info/ClassInfo.java#L183-L189 Shouldn't `makeAliasRawFullName` also take `getAliasPkg()` and `getAliasShortName()` as the first two parameters? If not, then I need to create such a method nevertheless, how should I call it?

And my issue: Somehow, the `offset` I get from here:
https://github.com/skylot/jadx/blob/9ad0f93a53f6644b6164e5821fe722eb0e2d0f96/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java#L165-L189
is different from the offset that's saved in JADX's project data (`JadxCodeRef`), computed here:
https://github.com/skylot/jadx/blob/14fd88b2f8e3a18fe2bfb7a9cdece3451341a029/jadx-gui/src/main/java/jadx/gui/ui/codearea/CommentAction.java#L94-L103

Which prevents method var comments from being exported:
https://github.com/skylot/jadx/blob/9ad0f93a53f6644b6164e5821fe722eb0e2d0f96/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java#L318-L320

skylot:
> Shouldn't makeAliasRawFullName also take getAliasPkg() and getAliasShortName() as the first two parameters? If not, then I need to create such a method nevertheless, how should I call it?

Yeah, here should be alias pkg and name. Intention was to make aliased name while not using inner classes, i.e. replace back '.' with '$', but looks like this was not used anywhere. And to be honest, I have no idea why you may need this method.

> And my issue: Somehow, the offset I get from here:
is different from the offset that's saved in JADX's project data (JadxCodeRef), computed here:

It is hard to tell. Maybe variable declaration without code don't have offset (like `int a;`) so you get offset from previous instruction. Jadx don't allow attaching comments for such variables for now, you can open a new issue if this is your case.

NebelNidas:
I'm an absolute idiot... The returned offset was correct the entire time; it was just my test comment that was placed on the wrong line! 🤦
Anyways, after the last two days being completely wasted, this PR is finally done! Definetely not the cleanest code I've ever written, but it does its job. 😅 I'll clean it up in the following PRs when I add mapping-io import support and completely replace the JOBF format with Tiny v2.

skylot:
@NebelNidas please move all your changes to a new class into a jadx-gui module and `jadx.gui.plugins.mapping` package (or any other name instead `mapping`). There are several reasons to do this:
1. Looks like you don't use any methods of deobfuscator class, so don't mix these features
2. Unfortunately `mapping-io` don't use maven central repository (and on top of that you just add jar file), so I don't want users of "jadx-lib" also add addition repo.
3. I am trying to strip all additional features from `jadx-core` and keep only decompilation in it.
4. Later I will try to move your changes to a new plugin and use it also in jadx-cli.

I understand that this will be make an import a little harder to implement (may need to add a new API to insert custom jadx passes), but result will be more modular (and less coupled).

jadx-core/src/main/java/jadx/api/data/IJavaNodeRef.java
```diff
@@ -2,7 +2,7 @@
 
 public interface IJavaNodeRef extends Comparable<IJavaNodeRef> {
 
-	enum RefType {
+	public enum RefType {
```
skylot, line 5: Not really needed, because all entries inside interface have `pulblic` modifier by default.

jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java
```diff
@@ -107,6 +132,237 @@ public void savePresets() {
...
+				String classPath = classInfo.makeRawFullName().replace('.', '/');
+				String rawClassName = classInfo.getRawName();
+
+				if (classInfo.hasAlias()
+						&& !classInfo.getAliasShortName().equals(classInfo.getShortName())
+						&& mappedClasses.contains(rawClassName)) {
+					mappingTree.visitClass(classPath);
+					String alias = classInfo.makeAliasRawFullName().replace('.', '/');
+
+					if (alias.length() >= Consts.DEFAULT_PACKAGE_NAME.length()
```
skylot, line 260: Idea suggest to use `alias.startsWith(Consts.DEFAULT_PACKAGE_NAME)` instead this :)

jadx-core/src/main/java/jadx/api/data/IJavaNodeRef.java
NebelNidas: Removed in f53735209191d912664181e9d5752efcdfbd9c6e!

jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java
NebelNidas: That would be easier, indeed xD

NebelNidas: Resolved by 782a56c482a0a382b1d0f2c9c6bf4b38f4593145

NebelNidas:
> Looks like you don't use any methods of deobfuscator class, so don't mix these features

I purposefully left it in there, because I want to replace the JOBF format with Tiny v2 in the next PR (and JOBF is also partly handled in the `Deobfuscator` class). But since that will be a separate PR, I can move it into your suggested package, np :)

skylot (approved):

NebelNidas:
Can you wait a few minutes before merging please? I'm making the exporting async

skylot:
> Can you wait a few minutes before merging please?

Sure. You can mark PR as draft if it is not ready for merge :)

> because I want to replace the JOBF format with Tiny v2 in the next PR

You don't need to replace jobf we're just adding support for other formats.
I plan to extract support for mapping format import/export into a separate optional plugin. Deobfuscator will also be moved to a separate plugin eventually.

> I purposefully left it in there

Please, don't mix mapping, deobfuscation and renames these are different things, yes they are related, but yet these are different features. Also, you don't need deobfuscator because node aliases can be accessed (and also updated) from `RootNode`.

NebelNidas:
Took a bit longer than expected, but it's ready to be merged now :)

PR #1511, opened 2022-06-04
Fix restoration of windows saved state

CmP-lt:
Suggested code changes alter current behavior of window saved state restoration to address the following issues:
  - window saved bounds not being restored when they extend past screen bounds in any direction;
  - divider location of main window's split pane is restored incorrectly when main window state is restored to maximized.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
+++ b/jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
@@ -1,9 +1,6 @@
 package jadx.gui.settings;
 
 import java.awt.Font;
-import java.awt.GraphicsDevice;
-import java.awt.GraphicsEnvironment;
-import java.awt.Rectangle;
 import java.awt.Window;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


skylot:
Changes with extended state is fine, but you can't just remove `isContainedInAnyScreen` method because it was added with a purpose (from PR #521):
>This change accepts the stored windows positions only if they are completely inside one of the visible screens, otherwise show it by the default location in the default screen.
The reason of this, if a user changes from two screens to one screen, then he won't be able to see the windows at all.
Ideally, we should check if a part of the window is inside the screen, or if the window is split between two screens, but I don't think this is necessary at this time.

And looks like it is a time to somehow merge these requirements. Maybe also save screens config and if it changes ignore saved position.

CmP-lt:
> Ideally, we should check if a part of the window is inside the screen
It does seem that such check is enough to handle all cases in which window saved bounds aren't included in any screen.

> Maybe also save screens config and if it changes ignore saved position.
Why would this be needed, assuming the check above is sufficient for handling problematic cases properly? Is it not?

skylot (approved):

skylot:
@CmP-lt looks good now, thanks!

CmP-lt:
451cc3d adds handling for cases when window saved bounds are outside of all screens by implementing the check described above. `isAccessibleInAnyScreen` is a new (suitable, I suppose) name for `isContainedInAnyScreen` method in which `Rectangle`'s `contains` method is replaced with `intersects`.

Issue #1548, opened 2022-06-21
Quark analysis failed


Alireza-Razavi:
I think there is a bug in Quark Engine.
StackTrace shown by jadx:
```
Quark analysis failed!
Error: null
java.lang.NullPointerException
	at jadx.gui.plugins.quark.QuarkReportPanel.prepareData(QuarkReportPanel.java:73)
	at jadx.gui.plugins.quark.QuarkReportPanel.<init>(QuarkReportPanel.java:67)
	at jadx.gui.plugins.quark.QuarkReportNode.getContentPanel(QuarkReportNode.java:63)
	at jadx.gui.ui.TabbedPane.getContentPanel(TabbedPane.java:319)
	at jadx.gui.ui.TabbedPane.showNode(TabbedPane.java:208)
	at jadx.gui.plugins.quark.QuarkManager.loadReport(QuarkManager.java:114)
	at jadx.gui.plugins.quark.QuarkManager.lambda$null$0(QuarkManager.java:54)
	at jadx.gui.jobs.SimpleTask.onFinish(SimpleTask.java:39)
	at jadx.gui.jobs.BackgroundExecutor$TaskWorker.lambda$doInBackground$1(BackgroundExecutor.java:148)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:301)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
```

[...]

PR #1556, opened 2022-06-24
QuarkReport: data validation added and other minor improvements

jpstotz:
Issues like #1548 are difficult to debug as they are cause by the data returned by Quark. 

I added data validation regarding the returned confidence value. In case this value is not as expected the whole record is printed, which allows a better understanding what went wrong in Quark.


skylot (approved):

skylot:
@jpstotz thanks!

Issue #1585, opened 2022-07-25
java.lang.NullPointerException


sdandroid:
Please describe what you did before the error occurred.
**IMPORTANT!** If the error occurs with a specific APK file please attach or provide link to apk file!
 * Jadx version: 1.4.3
* Java version: 11.0.13
* Java VM: Oracle Corporation Java HotSpot(TM) 64-Bit Server VM
* Platform: Linux (5.18.0-2-amd64 amd64)
* Max heap size: 10240 MB
* Program args: -Xmx10G -Xms128M -XX:MaxRAMPercentage=70.0 -XX:+UseG1GC -Dawt.useSystemAAFontSettings=lcd -Dswing.aatext=true

```
java.lang.NullPointerException
	at jadx.gui.device.protocol.ADBDevice.updateDeviceInfo(ADBDevice.java:42)
	at jadx.gui.ui.dialog.ADBDialog.onDeviceStatusChange(ADBDialog.java:282)
	at jadx.gui.device.protocol.ADB.lambda$listenForDeviceState$0(ADB.java:213)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)

```

PR #1586, opened 2022-07-25
fix: NullPointerException in adb device viewer (#1585)

jpstotz:
Fixed a minor problem when a device disconnects and reconnects: in that case the device data can be incomplete and a device with `null` serial was created.

Furthermore I improved `ADBDeviceInfo` class a bit (getter and constructor).


skylot (approved):

skylot:
Looks good! Thanks :slightly_smiling_face:

PR #1588, opened 2022-07-26
Fix mixed-up quotation marks in lang files

NebelNidas:
I know the guidelines tell you to open an issue first before creating PRs, but I reckoned pointing out all the questionable locations would take longer than simply providing the fix directly


skylot (approved):
Thanks :+1:

Issue #1583, opened 2022-07-24
Array zero 


TimScriptov:
Decompile array error

Orig:
```
<array name=name">
<item>0</item>
</array>
```

Decompile:
```
<array name=name">
<item>@null</item>
</array>
```

It code = NulPointerException

jpstotz:
@TimScriptov Can you provide a sample APK that causes this error on decompile? That would be really helpful to identify the cause.

TimScriptov:
Apk File:

https://www.dxsoft.com/hfpagerng.apk

Website:
https://hfpager.com/hfpager/hf-pager

jpstotz:
Thanks, I was able to reproduce this issue, e.g. in `res/values/arrays.xml`:

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <array name="a_audio_modes">
        <item>MODE_NORMAL</item>
        <item>MODE_IN_COMMUNICATION</item>
        <item>MODE_IN_CALL</item>
    </array>
    <array name="a_rx_freqs1">
        <item>@null</item>
        <item>1</item>
        <item>2</item>
        <item>3</item>
        <item>4</item>
        <item>5</item>
        <item>6</item>
        <item>7</item>
        <item>8</item>
        <item>9</item>
    </array>
```

The same apk file decompiled with apktool creates a correct entry:

```
   <string-array name="a_rx_freqs1">
        <item>0</item>
        <item>1</item>
```

It seems like Jadx intentiously replaces `0` by `@null`, but I don't know why:
https://github.com/skylot/jadx/blob/master/jadx-core/src/main/java/jadx/core/xmlgen/ResXmlGen.java#L173

skylot:
@jpstotz do be honest, I also don't know why it's done this way, it was added in very old PR #201.
So if anyone knows how to correctly fix this, PR is welcome 👍🏽

jpstotz:
Looking into the code of apktool I see three references of `"@null"`: 

In `ResReferenceValue.java` `ResFlagsAttr.java` and `ResEnumAttr.java`.

So my guess at the moment that we should consider at this point `value.getRawValue().getDataType()`.
In my tests for `@null` replacements that seem to be reasonable I usually saw `TYPE_REFERENCE` as dataType value. But the array case this issue is about it was `TYPE_STRING`. 

I will try to load more APKs to better understand when a replacement is currently applied and if it makes sense.

MrIkso:
[...]

PR #1594, opened 2022-07-31
XML "@null" decoding

jpstotz:
Fix for #1583. `@null` is now restricted to values that have a data type of `TYPE_REFERENCE`.

Furthermore I restricted the attribute name replacement to `TYPE_INT_DEC` values, in my tests only attributes with that type had a successful replacement, all the other just caused a `NumberFormatException`.

I compared the resources of several apps decompiled by Jadx and by apktool regarding `@null` replacement. Most apps had the same reference counter of `@null` in both decompiled apps. Some apktool decompiled apps had a few references more but I wasn't able to detect where the exact difference was located as the xml files are often named differently and the attributes are created in a different order which makes comparison difficult.

Last but not least I noticed that deobfuscator had a rule for abstract classes but not for interfaces.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java
+++ b/jadx-core/src/main/java/jadx/core/deobf/Deobfuscator.java
@@ -428,7 +428,9 @@ public class Deobfuscator {
 			return "Enum";
 		}
 		String result = "";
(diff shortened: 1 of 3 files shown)


skylot (approved):
Thanks!

PR #1605, opened 2022-08-06
fix: when hooking a constructor with Frida, call `$new` instead of `$…

Areizen:
### Description
Fix the javascript generated by Frida action that, in case of a constructor, generate a call to `$init` instead of `$new` to instantiate the object.


skylot:
@Areizen thanks!
Do I understand correctly that `$init` is also work, but `$new` is just more correct (i.e. create object instead of hooking constructor)?

Areizen:
In fact `$init` cannot be called, it result in an error. It is only used as a keyword to hook the constructor. The method that can be used to instantiate an object is another keyword `$new` which can be called (but not hooked).

skylot (approved):

Issue #1628, opened 2022-08-11
java.lang.IllegalArgumentException: Comparison method violates its general contract!


w296488320:
Please describe what you did before the error occurred.
**IMPORTANT!** If the error occurs with a specific APK file please attach or provide link to apk file!
 * Jadx version: 1.4.3
* Java version: 11.0.12
* Java VM: JetBrains s.r.o OpenJDK 64-Bit Server VM
* Platform: Windows 10 (10.0 amd64)
* Max heap size: 30688 MB
* Program args: -Xms128M -XX:MaxRAMPercentage=70.0 -XX:+UseG1GC -Dawt.useSystemAAFontSettings=lcd -Dswing.aatext=true

```
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.base/java.util.ComparableTimSort.mergeHi(ComparableTimSort.java:870)
	at java.base/java.util.ComparableTimSort.mergeAt(ComparableTimSort.java:487)
	at java.base/java.util.ComparableTimSort.mergeCollapse(ComparableTimSort.java:413)
	at java.base/java.util.ComparableTimSort.sort(ComparableTimSort.java:213)
	at java.base/java.util.Arrays.sort(Arrays.java:1249)
	at java.desktop/sun.awt.shell.Win32ShellFolderManager2.get(Win32ShellFolderManager2.java:313)
	at java.desktop/sun.awt.shell.ShellFolder.get(ShellFolder.java:259)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI$DirectoryComboBoxModel.addItem(MetalFileChooserUI.java:1029)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI.doDirectoryChanged(MetalFileChooserUI.java:717)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI$5.propertyChange(MetalFileChooserUI.java:806)
	at java.desktop/java.beans.PropertyChangeSupport.fire(PropertyChangeSupport.java:341)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:333)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:266)
	at java.desktop/java.awt.Component.firePropertyChange(Component.java:8754)
	at java.desktop/javax.swing.JFileChooser.setCurrentDirectory(JFileChooser.java:608)
[...]

Issue #1606, opened 2022-08-07
[jadx-gui] An error occurred while opening the file


xxr0ss:
* Jadx version: dev
* Java version: 11.0.15
* Java VM: Oracle Corporation OpenJDK 64-Bit Server VM
* Platform: Windows 11 (10.0 amd64)
* Max heap size: 12288 MB
* Program args: -Xms128M -Xmx12G -XX:MaxRAMPercentage=70.0 -XX:+UseG1GC -Dawt.useSystemAAFontSettings=lcd -Dswing.aatext=true

```
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.base/java.util.ComparableTimSort.mergeHi(ComparableTimSort.java:870)
	at java.base/java.util.ComparableTimSort.mergeAt(ComparableTimSort.java:487)
	at java.base/java.util.ComparableTimSort.mergeForceCollapse(ComparableTimSort.java:426)
	at java.base/java.util.ComparableTimSort.sort(ComparableTimSort.java:222)
	at java.base/java.util.Arrays.sort(Arrays.java:1249)
	at java.desktop/sun.awt.shell.Win32ShellFolderManager2.get(Win32ShellFolderManager2.java:313)
	at java.desktop/sun.awt.shell.ShellFolder.get(ShellFolder.java:259)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI$DirectoryComboBoxModel.addItem(MetalFileChooserUI.java:1029)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI.doDirectoryChanged(MetalFileChooserUI.java:717)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI$5.propertyChange(MetalFileChooserUI.java:806)
	at java.desktop/java.beans.PropertyChangeSupport.fire(PropertyChangeSupport.java:341)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:333)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:266)
	at java.desktop/java.awt.Component.firePropertyChange(Component.java:8731)
	at java.desktop/javax.swing.JFileChooser.setCurrentDirectory(JFileChooser.java:608)
	at java.desktop/javax.swing.JFileChooser.<init>(JFileChooser.java:362)
[...]

Issue #1574, opened 2022-07-10
java.lang.IllegalArgumentException: Comparison method violates its general contract!


c0m94d3:
Please describe what you did before the error occurred.
Just tried to launch the exe. I tried reinstalling java, but didn't work.

**IMPORTANT!** If the error occurs with a specific APK file please attach or provide link to apk file!
Nope.

 * Jadx version: 1.4.2
* Java version: 18.0.1.1
* Java VM: Oracle Corporation OpenJDK 64-Bit Server VM
* Platform: Windows 11 (10.0 amd64)
* Max heap size: 4096 MB

```
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.base/java.util.ComparableTimSort.mergeHi(Unknown Source)
	at java.base/java.util.ComparableTimSort.mergeAt(Unknown Source)
	at java.base/java.util.ComparableTimSort.mergeForceCollapse(Unknown Source)
	at java.base/java.util.ComparableTimSort.sort(Unknown Source)
	at java.base/java.util.Arrays.sort(Unknown Source)
	at java.desktop/sun.awt.shell.Win32ShellFolderManager2.get(Unknown Source)
	at java.desktop/sun.awt.shell.ShellFolder.get(Unknown Source)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI$DirectoryComboBoxModel.addItem(Unknown Source)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI.doDirectoryChanged(Unknown Source)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI$5.propertyChange(Unknown Source)
	at java.desktop/java.beans.PropertyChangeSupport.fire(Unknown Source)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(Unknown Source)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(Unknown Source)
	at java.desktop/java.awt.Component.firePropertyChange(Unknown Source)
	at java.desktop/javax.swing.JFileChooser.setCurrentDirectory(Unknown Source)
	at java.desktop/javax.swing.JFileChooser.<init>(Unknown Source)
	at java.desktop/javax.swing.JFileChooser.<init>(Unknown Source)
	at jadx.gui.ui.dialog.FileDialog$FileChooser.<init>(FileDialog.java:155)
[...]

Issue #1552, opened 2022-06-22
java.lang.IllegalArgumentException: Comparison method violates its general contract!


flaltangyi:
Please describe what you did before the error occurred.
**IMPORTANT!** If the error occurs with a specific APK file please attach or provide link to apk file!
 * Jadx version: 1.4.2
* Java version: 18.0.1.1
* Java VM: Oracle Corporation OpenJDK 64-Bit Server VM
* Platform: Windows 11 (10.0 amd64)
* Max heap size: 4096 MB

```
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.base/java.util.ComparableTimSort.mergeHi(Unknown Source)
	at java.base/java.util.ComparableTimSort.mergeAt(Unknown Source)
	at java.base/java.util.ComparableTimSort.mergeForceCollapse(Unknown Source)
	at java.base/java.util.ComparableTimSort.sort(Unknown Source)
	at java.base/java.util.Arrays.sort(Unknown Source)
	at java.desktop/sun.awt.shell.Win32ShellFolderManager2.get(Unknown Source)
	at java.desktop/sun.awt.shell.ShellFolder.get(Unknown Source)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI$DirectoryComboBoxModel.addItem(Unknown Source)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI.doDirectoryChanged(Unknown Source)
	at java.desktop/javax.swing.plaf.metal.MetalFileChooserUI$5.propertyChange(Unknown Source)
	at java.desktop/java.beans.PropertyChangeSupport.fire(Unknown Source)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(Unknown Source)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(Unknown Source)
	at java.desktop/java.awt.Component.firePropertyChange(Unknown Source)
	at java.desktop/javax.swing.JFileChooser.setCurrentDirectory(Unknown Source)
	at java.desktop/javax.swing.JFileChooser.<init>(Unknown Source)
	at java.desktop/javax.swing.JFileChooser.<init>(Unknown Source)
	at jadx.gui.ui.dialog.FileDialog$FileChooser.<init>(FileDialog.java:155)
	at jadx.gui.ui.dialog.FileDialog$FileChooser.<init>(FileDialog.java:155)
[...]

PR #1630, opened 2022-08-11
Fix broken FileDialog

xxr0ss:
### Description

Related issue
#1628 #1606 #1213 #1574 #1552


skylot:
That is a very interesting approach, thanks! :+1:

skylot (approved):

PR #1655, opened 2022-08-25
fix(gui): update pt-BR translation

iShi0n:
Add missing translations and correct one word.


skylot:
@iShi0n thank you for your contribution! :+1:

Issue #1411, opened 2022-03-18
[feature] Add Logcat output into debugger


TheCobraChicken:
I would like the Jadx debugger to include the logcat output within the Jadx GUI.  Ive been building the feature and have it working for my purposes but not quite ready to include into production.  Creating this issue to get others opinion.

Currently can be found at https://github.com/green9317/jadx/tree/logcatAddition  

![jadx_logcat_example](https://user-images.githubusercontent.com/38409554/158965118-81710fd9-7efb-4bc3-bfbb-e777e8f975f2.gif)

Before I submit I still want to add 
1. ability to filter to debugged process 
2. automatically update process list
3. button to clear logcat
4. button to pause logcat (for performance sensitive users)
5. Change how logcat output is displayed to make it easier to read
6. clean code

green9317:
Ill do that.  I wasn't sure if not having the latest commit would cause an issue when I do a pull request.

PR #1666, opened 2022-09-08
Add logcat output to JadX Debugger

TheCobraChicken:
Includes Logcat output directly into the JadX debugger.  

Currently works by running a timer that will call the logcat command every second and parse / store the output in an ArrayList for follow on filtering.  I have run into some lagging issues when attempting to display a large number of log messages at once particularly when attempting to display logs for all processes and logcat -c has not been run in a while. I did add the ability to pause the logcat timer and run logcat clear from within JadX to help if there is a huge amount of log messages being generated.  Let me know if you think this method is to resource intensive and I can tweak it.

logcat_demo

Finishes issue #1411


skylot (approved):

skylot:
@TheCobraChicken looks good. Thank you for your work :+1:

PR #1700, opened 2022-10-10

zhongqingsong:
### Description
Add new text about logcat, translate to Chinese.


skylot:
Thanks! :+1:

Issue #1714, opened 2022-11-02
Frida snippet don't work for constructors anymore


daMatz:
When generating a Frida snippet via jadx for a constructor, the code looks like this:

```
let MyClass = Java.use("com.MyClass");
MyClass["$init"].implementation = function () {
    console.log('$init is called');
    let ret = this.$new();
    console.log('$init ret value is ' + ret);
    return ret;
};
```

This generates the following error in the latest frida versions (haven't tested how far back that dates):

```
Error: Implementation for MyClass expected return value compatible with void
```

When using `let ret = this.$init();` instead, it works for me.

```
let MyClass = Java.use("com.MyClass");
MyClass["$init"].implementation = function () {
    console.log('$init is called');
    let ret = this.$init();
    console.log('$init ret value is ' + ret);
    return ret;
};
```

jpstotz:
According to https://github.com/skylot/jadx/pull/1605 `$new` should the the correct way to call an constructor. 

What version of Frida and frida-server do you use?

skylot:
Hm. I am not a Frida expert, but I believe that a `$new` is a correct call here. Unfortunately, error message is not very helpful, so it is hard to understand what is the main cause of the issue. I will try to reproduce it, so I will keep this issue open for some time.

As a workaround, I can move template for this snippet to the preferences, so it will be possible to edit it, but this solution is not flexible enough, because a lot of code used to form a strings in snippet. So maybe script for that will be better (I will try to make it).

daMatz:
@skylot I created a controlled environment to test this.

I created new Android Studio Project with a basic Activity and a basic class with constructor.
Here is the working example with `$init`:

![image](https://user-images.githubusercontent.com/9104870/199670659-a94ba295-29a1-4c32-a437-1c51215bf17f.png)

Here is the failing example with `$new`:

![image](https://user-images.githubusercontent.com/9104870/199670881-aed9c90c-e283-4264-a0a3-40f760741ba4.png)

I am happy to provide the code or condunct further tests, but I think this is pretty clear :)

xxr0ss:
Hi @skylot @daMatz ,

See https://github.com/frida/frida/issues/2302#issuecomment-1303073610 and `$init`'s document:

![image](https://user-images.githubusercontent.com/31790712/199919619-6fb95ef9-006b-4575-ac72-fc98214a38ef.png)


So I think even `$new` should not be called in the `$init`, because the instance is already created.

daMatz:
[...]

PR #1715, opened 2022-11-02
Change callMethodName of constructors in Frida action

daMatz:
Don't use `$new` as `callMethodName` for constructors in frida snippets, but `$init`.
Fixes: https://github.com/skylot/jadx/issues/1714


skylot:
@daMatz I merged this PR, so you can get some credit, but I will commit my fix to apply additional changes for templates as discussed in #1714

Issue #1724, opened 2022-11-12
BUG: ./gradlew dist error


rickyrsyah:
**To Reproduce**
```sh
git clone --depth 1 -b master https://github.com/skylot/jadx.git
cd jadx
./gradlew dist
```

**Debug Log**
```
FAILURE: Build failed with an exception.

* Where:
Build file '/media/data/Downloaded/jadx/jadx-gui/build.gradle' line: 62

* What went wrong:
A problem occurred evaluating project ':jadx-gui'.
> Could not resolve all files for configuration ':jadx-gui:runtimeClasspath'.
   > Could not find apksig-7.3.1.jar (com.android.tools.build:apksig:7.3.1).
     Searched in the following locations:
         https://dl.google.com/dl/android/maven2/com/android/tools/build/apksig/7.3.1/apksig-7.3.1.jar
   > Could not find aapt2-proto-7.3.1-8691043.jar (com.android.tools.build:aapt2-proto:7.3.1-8691043).
     Searched in the following locations:
         https://dl.google.com/dl/android/maven2/com/android/tools/build/aapt2-proto/7.3.1-8691043/aapt2-proto-7.3.1-8691043.jar
   > Could not find r8-3.3.75.jar (com.android.tools:r8:3.3.75).
     Searched in the following locations:
         https://dl.google.com/dl/android/maven2/com/android/tools/r8/3.3.75/r8-3.3.75.jar

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 16s
```

skylot:
@rickyrsyah looks like a network issue, because these urls works fine for me.
Also, I don't think you PR is needed because gradle shortcut is correct and `maven.google.com` repository also points to urls at `dl.google.com` (check `jar` artifact link at https://maven.google.com/web/index.html#com.android.tools.build:apksig:7.3.1).
So issue may be in name resolution (DNS) or Google mess up with redirections :rofl:

rickyrsyah:
> @rickyrsyah looks like a network issue, because these urls works fine for me. Also, I don't think you PR is needed because gradle shortcut is correct and `maven.google.com` repository also points to urls at `dl.google.com` (check `jar` artifact link at https://maven.google.com/web/index.html#com.android.tools.build:apksig:7.3.1). So issue may be in name resolution (DNS) or Google mess up with redirections :rofl:

[...]

PR #1725, opened 2022-11-12
Change Google Maven repository

rickyrsyah:
### Description
Fix issue #1724  error when downloading dependencies with result not found


skylot:
I think this change is not needed, check my comment at https://github.com/skylot/jadx/issues/1724#issuecomment-1312515761

PR #1746, opened 2022-12-12
Java.perform added for frida-snippet

Alireza-Razavi:
Currently the generated frida-snippet can not be executed by frida because it needs to be inside the ```Java.perform```.
I have added this possibility at this commit.


Alireza-Razavi:
@jpstotz 
Refer to my experiences with frida i decided to add this code.
I think without the ```Java.perform``` the generated snippet is unusable and every time Reverse-Engineer needs to add it manually.

skylot:
> If the snippet would contain Java.perform then this would require when pasting the n+1st snippet to remove the unnecessary Java.perform block.

I agree with @jpstotz here, so this PR is not needed.

@Alireza-Razavi if you really need to customize this snippet, you can use `jadx-next` branch with [`gui_custom_frida`](https://github.com/skylot/jadx/blob/jadx-next/jadx-plugins/jadx-script/examples/scripts/gui_custom_frida.jadx.kts) script.

PR #1747, opened 2022-12-13
Fix encoding special characters in AAB files

bagipro:
### Description
Special characters in AAB files are not currently encoded. This PR fixes it


skylot (approved):

skylot:
@bagipro thanks for fix :+1:

Issue #1772, opened 2023-01-27
[feature] frida snippet log should print the renamed function name


Ran-Naor:
When copying a frida snippet of a function the console.log contains the original name, even if you rename the function.
Lets say I rename the function "a" in class "SomeManager" to "startWork".
current frida snippet -
```js
let SomeManager = Java.use("com.example.SomeManager");
SomeManager["a"].overload('android.content.Context').implementation = function (context) {
    console.log('SomeManager.a is called' + ', ' + 'context: ' + context);
    let ret = this["a"](context);
    console.log('SomeManager.a return: ' + ret);
    return ret;
};
```
new frida snippet would look like this -
```js
let SomeManager = Java.use("com.example.SomeManager");
SomeManager["a"].overload('android.content.Context').implementation = function (context) {
    console.log('SomeManager.startWork is called' + ', ' + 'context: ' + context);
    let ret = this["a"](context);
    console.log('SomeManager.startWork return: ' + ret);
    return ret;
};
```
I will add the code for this change in a PR :).

skylot:
PR merged. Thanks! :slightly_smiling_face:

PR #1773, opened 2023-01-27
print the renamed function name in a frida snippet

Ran-Naor:
The implementation for this feature #1772

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
@@ -16,8 +16,6 @@ import jadx.api.ICodeInfo;
 import jadx.api.JavaClass;
 import jadx.api.JavaField;
 import jadx.api.JavaMethod;
-import jadx.api.data.ICodeRename;
-import jadx.api.data.impl.JadxNodeRef;
 import jadx.api.metadata.ICodeNodeRef;
 import jadx.api.metadata.annotations.NodeDeclareRef;
 import jadx.api.metadata.annotations.VarNode;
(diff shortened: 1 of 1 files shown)


jpstotz:
@Ran-Naor  IMHO it would be nicer if the log message would be generated similar as frida-trace does, which means not by string concat but this way:

    console.log(`SomeManager.a is called: context=${context}`);

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
```diff
@@ -78,10 +80,18 @@ private String generateMethodSnippet(JMethod jMth) {
 		MethodNode mth = jMth.getJavaMethod().getMethodNode();
 		MethodInfo methodInfo = mth.getMethodInfo();
 		String methodName;
+		String newMethodName;
 		if (methodInfo.isConstructor()) {
 			methodName = "$init";
+			newMethodName = methodName;
 		} else {
 			methodName = StringEscapeUtils.escapeEcmaScript(methodInfo.getName());
+			newMethodName = methodName;
```
skylot, line 89: Just use `methodInfo.getAlias()`
```suggestion
			newMethodName = StringEscapeUtils.escapeEcmaScript(methodInfo.getAlias());
```

skylot (changes requested):

jadx-gui/src/main/java/jadx/gui/ui/codearea/FridaAction.java
Ran-Naor: Oh ye you are correct, I missied this alias proprety so I retrived it directly from the renames list. I will change it and commit :)

skylot (approved):

jpstotz (approved):

PR #1792, opened 2023-03-05
Upgrade Spotless and wire Cleanthat

blacelle:
This proposes an upgrade of Spotless, and the integration of Cleanthat (to improve automatically the codebase).

The default configuration integrate [SafeAndConsensual](https://github.com/solven-eu/cleanthat/blob/master/java/src/main/java/eu/solven/cleanthat/engine/java/refactorer/mutators/composite/SafeAndConsensualMutators.java) mutators.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 plugins {
 	id 'com.github.ben-manes.versions' version '0.45.0'
-	id 'com.diffplug.spotless' version '6.13.0'
+	id 'com.diffplug.spotless' version '6.16.0'
 }
 
 ext.jadxVersion = System.getenv('JADX_VERSION') ?: "dev"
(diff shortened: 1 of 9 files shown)


blacelle:
>  I don't want integration tests to be changed by cleanthat (but formatting with spotless is fine) and for now I am not sure how to archive this.

I updated the PR to achieve this.

> Unfortunately, spotless stop supporting java 8 in version [6.14.0](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md#6140---2023-01-26) and jadx still uses java 8 for build. I plan to move to java 11 in future versions, but not decided in which one yet.

Fine. Cleanthat is also JDk11+.

> So, I can't accept this PR right now, but once above issues are resolved I will try to add cleanthat again because it looks cool! 🙂

Thanks for the positive feedback.

skylot:
> I updated the PR to achieve this.

> exclude 'jadx-core/src/test/java/jadx/tests/integration/**/*.java'

I mean, I want these files to be formatted by spotless but not changed by cleanthat, because code in these files IS a test, so it shouldn't be changed. :smile:

PR #1804, opened 2023-03-18
chore(core): handle RES_TABLE_TYPE_OVERLAY

### LINKED ISSUE #1748: [core]APP parsing error, and the overlay cannot parse
```
INFO  - loading ...
ERROR - Failed to parse '.arsc' file
jadx.core.utils.exceptions.JadxException: Error decode: resources.arsc
        at jadx.api.ResourcesLoader.decodeStream(ResourcesLoader.java:78)
        at jadx.core.dex.nodes.RootNode.loadResources(RootNode.java:173)
        at jadx.api.JadxDecompiler.load(JadxDecompiler.java:122)
        at jadx.cli.JadxCLI.processAndSave(JadxCLI.java:48)
        at jadx.cli.JadxCLI.execute(JadxCLI.java:36)
        at jadx.cli.JadxCLI.main(JadxCLI.java:20)
Caused by: java.io.IOException: Encountered unsupported chunk type TYPE_OVERLAY
at offset 0x5e0260
        at jadx.core.xmlgen.ResTableParser.parsePackage(ResTableParser.java:172)
        at jadx.core.xmlgen.ResTableParser.decodeTableChunk(ResTableParser.java:110)
        at jadx.core.xmlgen.ResTableParser.decode(ResTableParser.java:86)
        at jadx.core.xmlgen.ResDecoder.decode(ResDecoder.java:28)
        at jadx.core.dex.nodes.RootNode.lambda$loadResources$7(RootNode.java:173)
        at jadx.api.ResourcesLoader.decodeStream(ResourcesLoader.java:73)
        ... 5 common frames omitted
        
        
        and  Overlayable cannot be parsed
```

jpstotz:
The implementation to parse OVERLAY chunks in resource table. 

Fixes #1748

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ParserConstants.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ParserConstants.java
@@ -33,7 +33,8 @@ public class ParserConstants {
 	protected static final int RES_TABLE_TYPE_SPEC_TYPE = 0x0202; // 514
 	protected static final int RES_TABLE_TYPE_LIBRARY = 0x0203; // 515
 	protected static final int RES_TABLE_TYPE_OVERLAY = 0x0204; // 516
-	protected static final int RES_TABLE_TYPE_STAGED_ALIAS = 0x0206; // 517
+	protected static final int RES_TABLE_TYPE_OVERLAY_POLICY = 0x0205; // 517
+	protected static final int RES_TABLE_TYPE_STAGED_ALIAS = 0x0206; // 518
 
 	/**
 	 * Type constants
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ResTableParser.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ResTableParser.java
@@ -168,8 +168,11 @@ public class ResTableParser extends CommonBinaryParser implements IResParser {
 					parseLibraryTypeChunk(chunkStart);
 					break;
 				case RES_TABLE_TYPE_OVERLAY: // 0x0204
+					parseOverlayTypeChunk(chunkStart);
+					break;
+				case RES_TABLE_TYPE_OVERLAY_POLICY: // 0x0205
 					throw new IOException(
-							String.format("Encountered unsupported chunk type TYPE_OVERLAY at offset 0x%x ", chunkStart));
+							String.format("Encountered unsupported chunk type RES_TABLE_TYPE_OVERLAY_POLICY at offset 0x%x ", chunkStart));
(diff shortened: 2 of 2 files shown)


skylot (approved):

skylot:
@jpstotz thanks!
BTW, looks like we can make chunk parsing more general by moving calculating chunk size and skipping to the end in outer loop. This will allow fixing similar issues for other types and to skip not yet supported chunks without errors :slightly_smiling_face:

jpstotz:
@skylot Yes that would be possible but I intentionally implemented it this way to make sure people create an issue and provide sample files. Otherwise I don't see a chance in finding matching APK files for evolving Jadx resource parsing. 

It is possible to learn a lot by looking at the apktool source but often the apktool source code also doesn't work so it is IMHO better to raise an error, get the sample file and thus be able to implement the necessary code base don a real-world example.

skylot:
>  to make sure people create an issue and provide sample files

@jpstotz agree, I like this strategy :+1::rofl:

jpstotz:
@skylot may be we can make it a little bit more user friendly: parse unknown resource table chunks, show an error dialog but continue loading the other chunks.  

I implemented this approach as a test: 
https://github.com/jpstotz/jadx/commit/6aa46ad7394f532b505a5e700354213afe085ae3 

The shown error dialog contains the complete chunk data base64 encoded. For common chunks that are smaller than 4KB this should be possible. Also this simplifies getting sample data in case the APK can not be shared.

skylot:
@jpstotz well, `UncaughtExceptionHandler` usage look very hacky and I think it can kill jadx-cli, also it can be unexpected for users of jadx as a library. In `jadx-next` branch I started a gui api for plugins and jadx script ([JadxGuiContext](https://github.com/skylot/jadx/blob/b7368af34eda80aaef259b8dfc3be9682ba50c19/jadx-core/src/main/java/jadx/api/plugins/gui/JadxGuiContext.java#L12)) I think it is better to add a method there to show such errors, code will be like this:
```java
JadxGuiContext guiContext = root.getDecompiler().getPluginsContext().getGuiContext();
if (guiContext != null) {
	// running in jadx-gui, show error dialog
	guiContext.showExceptionDialog(...)
}
```

jpstotz:
@skylot Yes that is a bit hacky, but I have tested that it does not cause problems for jadx-cli. 

Of course the way [JadxGuiContext](https://github.com/skylot/jadx/blob/b7368af34eda80aaef259b8dfc3be9682ba50c19/jadx-core/src/main/java/jadx/api/plugins/gui/JadxGuiContext.java#L12) is designed makes it much cleaner. I will take a look onto jadx-next...

Issue #1073, opened 2020-12-30
[feature] Replace raw values in resource's xml with existing enum/flag whenever possible


ANR2ME:
There is this APK https://www.apkmonk.com/app/rti.business/ which seems to use: (based on it's AndroidManifest.xml)
```
compileSdkVersion="29"
platformBuildVersionCode="29"
minSdkVersion="17"
targetSdkVersion="29"
```

When i setup my `build.gradle` to use the same SDK & buildTools and tried to build the APK
I'm getting a lot of these kind of error:
```
\src\main\res\values-v21\styles.xml:486:5-492:13: AAPT: error: expected dimension enum but got (raw string) -1.
```
because of there are raw value (ie. `-1`) instead of enum identifier (ie. `auto`)
```
<style name="res_2131755463_widget_design_bottomsheet_modal" parent="@android:style/Widget">
        <item name="android:background">?android:attr/colorBackground</item>
        <item name="android:elevation">@dimen/design_bottom_sheet_modal_elevation</item>
        <item name="behavior_hideable">true</item>
        <item name="behavior_peekHeight">-1</item>
        <item name="behavior_skipCollapsed">false</item>
</style>
```
And fixed it by changing that `-1` to `auto` as declared in `\src\main\res\values\attrs.xml`:
```
<attr name="behavior_peekHeight" format="dimension">
        <enum name="auto" value="-1" />
</attr>
```
_*There can be more than 1 enum on other places with similar issue_

Sometimes i also saw similar issue with `flag` where the flag were declared within android SDK (ie. `C:\Android\android-sdk\platforms\android-29\data\res\values\attrs.xml`)
```
<!-- Default text typeface style. -->
    <attr name="textStyle">
        <flag name="normal" value="0" />
        <flag name="bold" value="1" />
        <flag name="italic" value="2" />
    </attr>
```

I'm wondering if it's possible to replace those raw values with their respective `enum`/`flag` identifier whenever available, in order to minimize the number of error related to raw values? 
So user only need to fix the remaining error (ie. missing enum identifier or combined flags)

Or may be there are better/easier way, like forcing AAPT to accept raw values through configuration (ie. build.gradle may be, not sure if it's possible)?

**Edit:**
[...]

Issue #1682, opened 2022-09-26
0x10  in styles.xml  <item name="android:gravity">0x10</item> <item name="android:textStyle">0x1</item>


tsatatwer:
Hi , 
after decompilation I found  in styles.xml file these :
`<item name="android:gravity">0x10</item> <item name="android:textStyle">0x1</item>`
Any idea??

jpstotz:
Yes, attributes values like `gravity` are not yet decoded to a constant value. We already have an issue open on that problem: https://github.com/skylot/jadx/issues/1631

skylot:
Fixed in PR #1812 by @nitram84.

PR #1812, opened 2023-04-07
Fix resource issues

nitram84:
I tested recompiling with some random simple apps, e.g.

https://apkpure.com/typescript-tutorial/com.cosmiclearn.typescripttutorial
https://apkpure.com/gradle-dependencies-generator/com.androidgradle.app

and just fixed the first resource or gradle issue in each iteration. The first app now fails with a java issue. I used gradle 7.4.2 to take another bug (also reproducible with gradle 7.6). 

For review: Each commit fixes one bug. I used apktool as a reference for the exported resources.

The following issues can now be closed: https://github.com/skylot/jadx/issues/1073 and https://github.com/skylot/jadx/issues/1682

For the next issues a data structure to aggregate infos for gradle build file generation would help, e.g. for a flag to track usage of "android:pathData-Attributes" + minSdkVersion < 21 to set "vectorDrawables.useSupportLibrary = true" or a legacy Apache http client detection....

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/nodes/RootNode.java
+++ b/jadx-core/src/main/java/jadx/core/dex/nodes/RootNode.java
@@ -43,6 +43,7 @@ import jadx.core.utils.Utils;
 import jadx.core.utils.android.AndroidResourcesUtils;
 import jadx.core.utils.exceptions.JadxRuntimeException;
 import jadx.core.xmlgen.IResParser;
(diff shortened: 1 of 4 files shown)


skylot (approved):

skylot:
@nitram84 thanks for fixes!
These sample apps are really very simple, so I think it is possible to fix java code compilation issues :+1:

PR #1813, opened 2023-04-07
fix: filter invalid chars for gradle wrapper

nitram84:
Some special chars in the app name can cause conflicts when you try to add a gradle wrapper. In this pull request I filter all invalid chars before generating settings.gradle

Here is an example app containing invalid chars: https://xn--who-becomes-rich--84ns4b8gwbu6jza3u.apk.gold/

Steps to reproduce:
- save as gradle project
- run 'gradle wrapper'

```
FAILURE: Build failed with an exception.

* Where:
Settings file 'settings.gradle' line: 2

* What went wrong:
A problem occurred evaluating settings.
> The project name 'Who Becomes Rich?' must not contain any of the following characters: [/, \, :, <, >, ", ?, *, |]. Set the 'rootProject.name' or adjust the 'include' statement (see https://docs.gradle.org/7.6/dsl/org.gradle.api.initialization.Settings.html#org.gradle.api.initialization.Settings:include(java.lang.String[]) for more details).
```

This issue can also easily be reproduced with new Android Studio project. Android Studio does not allow any special chars in the application name when the app is generated, but AndroidManifest.xml can be edited later like in my unit test.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/export/ExportGradleProject.java
+++ b/jadx-core/src/main/java/jadx/core/export/ExportGradleProject.java
@@ -6,6 +6,7 @@ import java.io.StringReader;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import javax.xml.parsers.DocumentBuilder;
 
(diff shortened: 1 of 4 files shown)


skylot (approved):

skylot:
@nitram84 thanks! Nice fix.

PR #1815, opened 2023-04-09
fix(xml): fix regression - use parent attibute only for styles

nitram84:
I'm sorry, my previous commit 378c747 introduced a regression. 378c747 added an empty parent attribute to arrays, plurals and attrs. This commit adds a check for type "style" to restore the previous behavior.


skylot (approved):

skylot:
@nitram84 thanks :+1:

PR #1857, opened 2023-05-06
Fix xml escaping for string arrays

nitram84:
I noticed that single quotes in xml string arrays were not escaped correctly (`&apos;` instead of \\'). 

Steps to reproduce:
Decompile this apk and run a gradle build: https://github.com/ds10git/tvbrowserandroid/releases/download/release_0.7.0.8/TV-Browser-v0.7.0.8.apk

The first build error is caused by a `&apos;` in res/values/arrays.xml.

Reference file is:
https://github.com/ds10git/tvbrowserandroid/blob/master/TV-Browser/res/values/strings.xml 
-> `/resources/string-arrray[@name='pref_category_reminder_names_second']`

This PR will fix this issue.

But first I prepared some unit tests for ResXmlGen to prevent regressions. ResXmlGen had a poor coverage before.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ResXmlGen.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ResXmlGen.java
@@ -243,7 +243,7 @@ public class ResXmlGen {
 			cw.add(" />");
 		} else {
 			cw.add('>');
-			if (itemTag.equals("string")) {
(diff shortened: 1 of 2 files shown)


skylot (approved):

skylot:
@nitram84 thanks! Tests looks great :+1:

PR #1860, opened 2023-05-08
Feat/kotlin metadata

Arsylk:
### Description

- Added support for `kotlin.Metadata` parsing & automatic renames
- Added support for data class flag parsing, and some hacky implementation of toString method for field and method renaming

If you care to implement any of those features, I would love to help. I have no way of contacting you about what changes I should make to get this merged, but I have some free time, and am willing to help. "
Either email me at iwaniuk.krzysztof@outlook.com or hit me up on discord Arsylk#8993


skylot:
Looks good :+1:
I can merge your changes as is, but I think at this point it will be nice to move such functionality into a separate plugin module. Some suggestions:
- as a base, you can use [`rename-mappings` plugin](https://github.com/skylot/jadx/blob/master/jadx-plugins/jadx-rename-mappings/src/main/java/jadx/plugins/mappings/RenameMappingsPlugin.java)
- remove `parseKotlinMetadata` option from `JadxArgs` and create similar one in your plugin.
- simple integration test will be nice to have (use [JadxScriptPluginTest](https://github.com/skylot/jadx/blob/06ec12ba17b02601145c47348930c81c232eb616/jadx-plugins/jadx-script/jadx-script-plugin/src/test/kotlin/JadxScriptPluginTest.kt#L15) as example). Or at least provide some sample apk, so I can check your changes.

Arsylk:
I'll try to rework it into the plugin system, seems more modular and probably will make it easier to maintain later on.

NebelNidas:
In the future, it would be nice if JADX could expose a Name Proposal interface where features like these would be channeled through (with corresponding highlighting in the GUI); see Enigma's [`NamePropsalService` interface](https://github.com/FabricMC/Enigma/blob/26ba5a59976a8b4a5e9d71d52f5a13a2aa2882de/enigma/src/main/java/cuchaz/enigma/api/service/NameProposalService.java) (and the built-in [Enum name proposer](https://github.com/FabricMC/Enigma/blob/26ba5a59976a8b4a5e9d71d52f5a13a2aa2882de/enigma/src/main/java/cuchaz/enigma/analysis/BuiltinPlugin.java#L49-L54) for reference)

skylot:
@NebelNidas this is nice suggestion, thanks! I create a new issue for further discussion (#1862). But I am not sure how to better implement this in UI, and it will be nice if you can post (at that issue) a couple screenshots from Enigma for a reference :slightly_smiling_face:

Arsylk:
Should be merge ready, reworked it to use plugins, added some basic tests and made toggle options for all the features inside.

jadx-core/src/main/java/jadx/core/plugins/JadxPluginManager.java
```diff
@@ -55,13 +55,13 @@ public void load() {
 	public void register(JadxPlugin plugin) {
 		Objects.requireNonNull(plugin);
 		PluginContext addedPlugin = addPlugin(plugin);
-		LOG.debug("Register plugin: {}", addedPlugin.getPluginId());
+		LOG.warn("Register plugin: {}", addedPlugin.getPluginId());
```
skylot, line 58: I don't think these changes to log level are needed

jadx-core/src/main/java/jadx/core/utils/kotlin/KotlinMetadataConsts.kt
```diff
@@ -0,0 +1,83 @@
...
+		?.takeIf { it.type == EncodedType.ENCODED_ARRAY && it.value is List<*> }
+	return encodedValue?.value?.let { it as List<EncodedValue> }
+}
+
+private fun IAnnotation.getParamAsStringArray(paramName: String): Array<String>? {
+	return getParamsAsList(paramName)
+		?.map<EncodedValue, Any?>(EncodedValue::getValue)
+		?.onEach { if (it != null && it !is String) /* TODO is this valid ? */ return@onEach }
+		?.map { "$it" }
+		?.toTypedArray()
```
skylot, line 58: Well, if `EncodedValue::getType` return `ENCODED_STRING` so returned value should be a `String`.
And you don't need to create a string just use cast.

skylot (changes requested):
Why you keep a lot of code in `jadx-core` module? All new classes and dependencies should be moved to your new module. Test also better to move, but it is slightly harder (I can do it myself)
Also run `./gradlew spotlessApply` to fix code formatting (as mention [here](https://github.com/skylot/jadx/wiki/Code-Formatting)).

P.S. I am not a Kotlin expert, and it is hard for me to check your code for correctness, so if test pass it is fine :rofl:

jadx-core/src/main/java/jadx/core/plugins/JadxPluginManager.java
Arsylk: Ah true, it was some leftover stuff from me trying to get the plugin to load, will revert and check for other logs

jadx-core/src/main/java/jadx/core/utils/kotlin/KotlinMetadataConsts.kt
Arsylk: The d2 field in kotlin.Metadata is Array<String> type, so I expected every element in it to be of String type, but the EncodedValue parser sometimes converted it to other types.

Can't filter those as array size needs to stay the same, otherwise i get `ArrayIndexOutOfBoundsException` from the `toKmClass()` method. I'm not sure if thats the correct way to convert that EncodedValue to the original string it was decoded from, i did not notice anything being broken because of it.

Arsylk:
I moved all the new stuff into the plugin module and did spotless apply

skylot:
Ok, thanks. I will move test.

skylot (approved):

skylot:
Merged. @Arsylk thank you for your work :+1:

PR #1869, opened 2023-05-15
option for hexadecimal integer format globally

Arsylk:
#1171 

Added global toggle for decimal or hexadecimal format

### DIFF
--- a/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
+++ b/jadx-cli/src/main/java/jadx/cli/JadxCLIArgs.java
@@ -24,6 +24,7 @@ import jadx.api.JadxArgs.RenameEnum;
 import jadx.api.JadxArgs.UseKotlinMethodsForVarNames;
 import jadx.api.JadxDecompiler;
 import jadx.api.args.GeneratedRenamesMappingFileMode;
+import jadx.api.args.IntegerFormat;
 import jadx.api.args.ResourceNameSource;
 import jadx.api.args.UserRenamesMappingsMode;
 import jadx.core.utils.exceptions.JadxException;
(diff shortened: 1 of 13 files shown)


skylot (approved):

psolyca:
Thanks ^^

Arsylk:
If there are any more features that aren't super complicated, I could probably finish some of them.
I'm playing round with  #1689 seems to be related to using line hints  before doing other transformations, seems to help, but I'm not really sure I should be messing with core decompiler parts like `IfRegionVisitor`

Other  than that I might try to implement inlining of `WhenMappings`.

skylot:
> If there are any more features that aren't super complicated

Not much actually, because I am also trying to fix easy ones at first :rofl:
Core decompiler mostly have bugs, which are hard to debug and fix without affecting other tests. So it leaves mostly UI features/issues, but they also may be complicated due to multithreaded issues :sob:

> I'm playing round with https://github.com/skylot/jadx/issues/1689 seems to be related to using line hints

Line numbers are very unreliable, because often missing or just incorrect, but sure, can help in some cases. Overall code style fixes are tricky because it is hard to distinguish and measure "what is better" and also can easily break other tests.

> Other than that I might try to implement inlining of WhenMappings.

Not sure what you mean here, do we have an issue for that?

Anyway, you are free to fix/implement what you want :slightly_smiling_face:

DedkovArtem:
when will this feature be released? really looking forward to it

psolyca:
You can grab an unstable release to test it, look into Actions tab.
Do not take the last one, it's buggy.

skylot:
@Arsylk I plan to make release at end of this month.
Until then, you can try [latest unstable build](https://nightly.link/skylot/jadx/workflows/build-artifacts/master).

Issue #1847, opened 2023-04-25
[core] BuildConfig files lost during exporting code as gradle project


Fi5t:
`jadx.core.export.ExportGradleProject` class contains a method `skipGeneratedClasses`, which prevents exporting `BuildConfig` classes. From my point of view this is a bug. I tried removing that call from the `init()` method and found no problem. @skylot, is this a bug or a feature? =)

skylot:
> That constants comes from the [buildConfigField](https://developer.android.com/build/gradle-tips#share-custom-fields-and-resource-values-with-your-app-code) options in the build.gradle files.

Oh, I see, so it will be better just keep these classes as is. I will remove filtering code :+1:

nitram84:
What about skipping this class, but analyzing it during gradle export and generate all non-standard buildConfigFields in build.gradle? Wouldn't this be the best option? Is there a sample app to test with?

Issue #1459, opened 2022-04-22
[feature] Add "provide java file" for partially deobfuscated files.


kschoos:
I'm working on an obfuscated app and it would be awesome if I could provide the actual java source file for certain obfuscated decompiled files.

For example, kotlin/jvm/internal/Intrinsics.java is deobfuscated. If I could just right click and provide it and it would fix the obfuscation based on function signatures for starters and maybe function logic in the long run, I think that would be useful.

interested? not interested?

skylot:
@kschoos your request is very similar to #66, and it is hard to match methods with everything renamed and also modified by java compiler and d8/r8 tool. Anyway, I may implement that in future because it will be very useful and also have high demand from users.

kschoos:
Thanks for checking in - in what form would you need something that you could pop in here easily? What kind of interface should a feature like this have so that you can add it without too much of a hassle?

skylot:
@kschoos you want to implement that feature? Great! :)
Unfortunately, there is no ready to use plugin api, so options are:
1. Fill renames list from [JadxCodeData](https://github.com/skylot/jadx/blob/dfdc14ea8618a7242d04bd6ea066d3e6b60a10b9/jadx-core/src/main/java/jadx/api/data/impl/JadxCodeData.java#L12=) same as [RenameDialog](https://github.com/skylot/jadx/blob/83decc24731dd5883cc8cd34e50e078c9244de61/jadx-gui/src/main/java/jadx/gui/ui/dialog/RenameDialog.java#L64) do (jadx-gui only approach)
1. Add to ICodeData a list with input files and process/apply this list in [RenameVisitor](https://github.com/skylot/jadx/blob/f674a29a6486c2f4814fed00a7dcc33135745ba0/jadx-core/src/main/java/jadx/core/dex/visitors/rename/RenameVisitor.java#L49). This will allow using this feature both in jadx-cli and jadx-gui.

I think first options will be much easier for you to implement, and later I can easily change code to move to second option.

Issue #66, opened 2015-05-31
Deobfuscation search engine


Sputuks3:
Congrats in implementing the deob feature, but I have an idea that I like to propose to JADX first. Some sort of java search engine. or signature recognition to identify obfuscated opensource libraries online on either Github or Google code. For example,

![deobfssearch](https://cloud.githubusercontent.com/assets/1269084/7900373/7a50259c-0785-11e5-8352-a18efc97d304.jpg)

skylot:
This is very interesting feature and I saw something similar in JEB decompiler ([Sign and Match Third-Party Library Code](https://www.pnfsoftware.com/blog/2013/07/27/jeb-1-3-how-to-sign-and-match-third-party-library-code/)) and maybe I will implement it in future, but for now there are a lot of questions how to implement it.

Sputuks3:
Perhaps in the de-obfuscation toggle button, rename all detected signatures with a comment

/*renamed from com.myapp.a.a*/

xseregax:
Sometimes there are names such as IlIIIIIlllIIIIlIllIIIIlllII, IlIIIIIlllIIIIlIllIIllllIII (all printable), can you add regex field in the settings that i could manually add my own patterns of bad names, or dump all the names in the file jobf for possible external processing aliases  or some other option if someone is needed. until i editing source code on each case.

nitram84:
One of the open questions of this issue is, how can we find the corresponding sources. If jadx knows the dependencies and their exact versions, this would be much easier to implement.

There are different approaches to achieve this e.g.
* https://github.com/LibChecker/LibChecker (rule based)
* https://github.com/reddr/LibScout (static analysis)

None of these tools can be used directly in jadx and the results are not good enough.

But during decompilation jadx gains a lot of data, that be used to detect dependencies.

My proposition is a public api in jadx to aggregate dependency data with version ranges, so plugins and script can read, provide detected dependencies or narrow versions.

Some examples
* apps with multiple dex files and minsdk<21 probably use 'androidx.multidex:multidex'. A script or a jadx-dependency-plugin may implement a multidex rule, read 'androidmultidexsupportversion.txt' and get the version
* if kotlin metadata is decoded, there is probably a kotlin stdlib dependency
* it should be possible to write a regex that detects the exact gson version - version strings can be found in the decompiled source code
* many obfuscated apps are leaking dependency data with version info in property files

[...]

PR #1879, opened 2023-05-24
fix(gradle): add legacy support for vector drawables

nitram84:
At the moment missing legacy support for vector drawables (https://developer.android.com/studio/write/vector-asset-studio) is one of the top reasons why recompilation fails. In most cases there wouldn't be a problem if the original dependencies were used (e.g. original material design dependency). Since we don't have any dependency info (yet) in jadx, I would like to propose this PR to add "vectorDrawables.useSupportLibrary = true" when it's needed to continue with the build. I identified two cases for api level 21 and  24.

Example 1 (api level 21):
https://github.com/niranjan94/show-java/releases/download/v3.0.5/ShowJava_v3_0_5.apk ( let Jadx decompile itself :-) )

`File was preprocessed as vector drawable support was added in Android 5.0 (API level 21)`

Example 2 (api level 24): https://github.com/corona-warn-app/cwa-app-android / https://m.apkpure.com/de/corona-warn-app/de.rki.coronawarnapp

`File was preprocessed as vector drawable android:filltype support was added in Android 7.0 (API level 24)`

To detect vector drawables resources must be processed before gradle export starts. So I had to create a task to run gradle export as final step. (I my opinion there are at least 2 more use cases why gradle export has to be the final task: https://github.com/skylot/jadx/issues/1847 and apache http legacy client detection).

Off topic: Some thoughts about an export process in the future:
In my opinion a batch execution model with fork join patterns might fit better for future requirements. I'm thinking of a workflow like this:

- step 1: initialization
- step 2: analyze all resources (not saving) in parallel​
- step 3: analyze all sources (not saving) in parallel
- step 4: run additional plugins and execute scripts (sequential execution, allow plugins and scripts to filter resources and sources)
- step 5/6: rerun step 2 and 3 (e.g. resolving types or deobfuscation (https://github.com/skylot/jadx/issues/66, https://github.com/skylot/jadx/issues/1459) based on dependencies detected by plugins and scripts), save sources and resources
- step 7: export gradle files
- step 8: run post processing plugins and scripts (e.g. security analysis, privacy leakage detection, ...)

### DIFF
--- a/jadx-core/src/main/java/jadx/api/JadxDecompiler.java
+++ b/jadx-core/src/main/java/jadx/api/JadxDecompiler.java
@@ -42,7 +42,7 @@ import jadx.core.dex.nodes.MethodNode;
 import jadx.core.dex.nodes.PackageNode;
 import jadx.core.dex.nodes.RootNode;
 import jadx.core.dex.visitors.SaveCode;
-import jadx.core.export.ExportGradleProject;
+import jadx.core.export.ExportGradleTask;
 import jadx.core.plugins.JadxPluginManager;
 import jadx.core.utils.DecompilerScheduler;
 import jadx.core.utils.Utils;
@@ -50,7 +50,6 @@ import jadx.core.utils.exceptions.JadxRuntimeException;
 import jadx.core.utils.files.FileUtils;
 import jadx.core.xmlgen.BinaryXMLParser;
 import jadx.core.xmlgen.ProtoXMLParser;
-import jadx.core.xmlgen.ResContainer;
 import jadx.core.xmlgen.ResourcesSaver;
 
 /**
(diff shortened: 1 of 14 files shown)


skylot (approved):

skylot:
@nitram84 thank you for your work! :+1:
Although, I will try to refactor code and get rid of task counting, because it is very easy to made incorrect count and as result the last task will never execute. I think using sub executor and awaiting all task finish can be a better option here, but it is not critical and just optional.

> Off topic: Some thoughts about an export process in the future:

Well, firstly, here a current code process stages:
- Loading (partial parsing to build class/methods lists)
- "Pre-decompilation" passes, sequential execution, run for all classes:
  1. Fast scan bytecode to build usage info
  2. Renames and deobfuscation
- Decompilation, parallel, can be optional because now jadx-gui trying not to force full decompilation. Class filtering done by adding `DONT_GENERATE` flag.
- No common post-processing stage, because not classes generated in jadx-gui.
But if needed can be done by requesting decompilation of every class (class code can be fetched from code cache), this done in code search for example.

So for me this is very close to your steps, but now it is only for code and without post-process step because of mentioned reasons.

P.S. Looks like you are really waiting for #66, maybe I will try to approach it one more time :rofl:

nitram84:
@skylot Thank you very much for your answer.

I was aware that task counting would be an intermediate solution, but it was a simple way to show why we may need a different execution model and it enables me to implement more features.

My primary goal is not directly #66. It's still recompilation with dependency detection in order to rebuild apps without trackers, ads and spyware :smile: . Automatic deobfuscation using existing sources would solve some important issues, but I think it's too early to start with it. I'll try to explain my idea in #66.

PR #1893, opened 2023-06-03
Fix mapping export

NebelNidas:
Not sure how I haven't caught this earlier, the current implementation overwrites new renames with the old renames from the imported mapping tree. The fix is to copy the old mappings first, and apply the new renames afterwards.

### DIFF
--- a/jadx-plugins/jadx-rename-mappings/src/main/java/jadx/plugins/mappings/save/MappingExporter.java
+++ b/jadx-plugins/jadx-rename-mappings/src/main/java/jadx/plugins/mappings/save/MappingExporter.java
@@ -144,10 +144,12 @@ public class MappingExporter {
 
 			String srcNamespace = MappingUtil.NS_SOURCE_FALLBACK;
 			String dstNamespace = MappingUtil.NS_TARGET_FALLBACK;
+
+			// Copy mappings from potentially imported mappings file
 			if (loadedMappingTree != null && loadedMappingTree.getDstNamespaces() != null) {
-				srcNamespace = loadedMappingTree.getSrcNamespace();
-				dstNamespace = loadedMappingTree.getDstNamespaces().get(0);
+				loadedMappingTree.accept(mappingTree);
 			}
+
 			mappingTree.visitHeader();
 			mappingTree.visitNamespaces(srcNamespace, Collections.singletonList(dstNamespace));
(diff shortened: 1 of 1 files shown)


jadx-plugins/jadx-rename-mappings/src/main/java/jadx/plugins/mappings/save/MappingExporter.java
```diff
@@ -144,10 +144,12 @@ public void exportMappings(Path path, JadxCodeData codeData, MappingFormat mappi
 
 			String srcNamespace = MappingUtil.NS_SOURCE_FALLBACK;
 			String dstNamespace = MappingUtil.NS_TARGET_FALLBACK;
+
+			// Copy mappings from potentially imported mappings file
 			if (loadedMappingTree != null && loadedMappingTree.getDstNamespaces() != null) {
-				srcNamespace = loadedMappingTree.getSrcNamespace();
-				dstNamespace = loadedMappingTree.getDstNamespaces().get(0);
```
skylot, line 149: @NebelNidas are you sure these namespaces assigns should be removed?

NebelNidas:
`accept` visits the namespaces automatically: https://github.com/FabricMC/mapping-io/blob/50e0877d700729dc29828713629f9e086a0f18c2/src/main/java/net/fabricmc/mappingio/tree/MemoryMappingTree.java#L216-L219

But there is more at play than I initially suspected, I actually had a reason as to why I implemented it the way I initially did. Your refactor into a plugin got rid of the rename forwarding to the mapping tree, and thus broke some of the existing assumptions in `MappingExporter`. This PR here will do for the time being, but it needs a proper fix in the long run. I'll create a separate issue with some further details

jadx-plugins/jadx-rename-mappings/src/main/java/jadx/plugins/mappings/save/MappingExporter.java
NebelNidas: `accept` visits the namespaces automatically, so it should be fine: https://github.com/FabricMC/mapping-io/blob/50e0877d700729dc29828713629f9e086a0f18c2/src/main/java/net/fabricmc/mappingio/tree/MemoryMappingTree.java#L216-L219

skylot (approved):

Issue #1832, opened 2023-04-21
[feature] Jadx script: jump to symbol and rename listener


yoavst:
**Background:**
I'm the author of  [Graffiti](https://github.com/yoavst/Graffiti), a tool for creating graphs interactively from the IDE.
In order to support JADX, those features in jadx-script are required:

- [x] Keybinding/menu for code area - Using `Gui.addPopupMenuAction`.
- [x] Get the method name and class from `ICodeNodeRef `
- [x] Detect renames so I could update the graph
- [ ] Given an "address" (full method signature), show it on JADX.

So those features are missing for support:

**Rename listener**
It would be nice to able to listen to the events of user renaming a symbol (class, method, field, package, package for class)

JEB for example, uses the following api:
```py
dex = prj.findUnit(IDexUnit)
dex.addListener(RenameListener())

class DexListener(IEventListener):
    def onEvent(self, e):
        if isinstance(e, JebEvent) and e.type == J.UnitChange and e.data is not None:
            if e.data.type == UnitChangeEventData.NameUpdate:
                target = e.data.target
```

**Jump to symbol**
Given a symbol (For example `ICodeNode`), it would be really helpful if we could be able to open its code in the view programmatically. Same as [The implementation](https://github.com/skylot/jadx/blob/master/jadx-gui/src/main/java/jadx/gui/ui/dialog/CommonSearchDialog.java#L149) in `CommonSearchDialog` probably.

JEB for example, uses:
```py
dex = prj.findUnit(IDexUnit)
ctx.navigate(dex, FULL_DEX_ADDRESS)
```
on IDA, it is:
```py
ida_kernwin.jumpto(addr)
```

skylot:
@yoavst recently I make a [jadx example plugin](https://github.com/jadx-decompiler/jadx-example-plugin) and also implement a way to install/manage plugins from cli and gui.
So, if you want, you can create a plugin instead script :slightly_smiling_face:

yoavst:
I'll try this, thanks!
Closing the issue as it is fixed.

PR #1903, opened 2023-06-09
Access node under caret/mouse and jump

yoavst:
### Description
This fixes the last bullet of #1832 by providing the following two features:
* jump to a `ICodeNodeRef` by exposing `jadx.gui.open`.   
The implementation is a simple wrapper over `mainWindow.tabbedPane.codeJump`. It uses `JNodeCache` to cache the `JNode`s created.
* Expose access to the enclosing node for actions, by adding `jadx.gui.(enclosing)?NodeUnder(Caret|mouse)`.   
It is useful for scripts that operates on methods (like Graffiti), as it allows the user to activate the action anywhere inside the function, not only on the function name.

In addition, I've added two sample scripts:
* `gui/caret_mouse.jadx.kts` - Shows the return value of `jadx.gui.(enclosing)?NodeUnder(Caret|mouse)`
* `gui/bookmark.jadx.kts` - Shows the usefulness of `jadx.gui.open` by implementing a single bookmark support in jadx. It can be extended to provide a normal bookmarks feature.

I'm not fully familiar with the codebase, so I hope those are the current implementations for those functionalities.

### DIFF
--- a/jadx-core/src/main/java/jadx/api/plugins/gui/JadxGuiContext.java
+++ b/jadx-core/src/main/java/jadx/api/plugins/gui/JadxGuiContext.java
@@ -49,4 +49,15 @@ public interface JadxGuiContext {
 	 * Access to GUI settings
 	 */
 	JadxGuiSettings settings();
+
+	ICodeNodeRef getNodeUnderCaret();
+	ICodeNodeRef getNodeUnderMouse();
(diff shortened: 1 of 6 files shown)


skylot (approved):

skylot:
@yoavst thanks! Great work :+1:

Issue #1922, opened 2023-06-20
[feature]open jadx-gui from double click file or open with


chaos-cn:
open jadx-gui from double click file or open with

jpstotz:
What is your OS? 

On Windows 10 to my knowledge assigning Jadx-Gui  (EXE version) to APK files or adding it to the list of "Open With" apps works without problems.

chaos-cn:
> What is your OS?
> 
> On Windows 10 to my knowledge assigning Jadx-Gui (EXE version) to APK files or adding it to the list of "Open With" apps works without problems.

MacOS, I use it like [1308](https://github.com/skylot/jadx/issues/1308)

skylot:
@chaos-cn I merged your PR. I added additional checks because it may throw an exception on unsupported platforms (only macOS is supported for that action :rofl:).

Also, I will try to set up [jdeploy](https://www.jdeploy.com/) for next jadx version to provide platform specific installers. Hope it will simplify jadx install. Use #1308 issue to track progress.

PR #1923, opened 2023-06-20
feature:open jadx-gui from double click file or open with

chaos-cn:
Reference issue: [1922](https://github.com/skylot/jadx/issues/1922)

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/JadxGUI.java
+++ b/jadx-gui/src/main/java/jadx/gui/JadxGUI.java
@@ -1,11 +1,14 @@
 package jadx.gui;
 
+import java.awt.Desktop;
+
 import javax.swing.SwingUtilities;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import jadx.cli.LogHelper;
+import jadx.core.utils.files.FileUtils;
 import jadx.gui.logs.LogCollector;
 import jadx.gui.settings.JadxSettings;
 import jadx.gui.settings.JadxSettingsAdapter;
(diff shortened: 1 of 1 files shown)


jpstotz:
Sorry but I don't think it is a good idea to always register Jadx to all supported file extensions on ever start. Some users use Jadx-Gui as a portable app and in this scenario registering file types is in my opinion not a good idea. Such an option belongs to a.g. a special command in preferences dialog where users can execute it of they want to do so. And even there I would not recommend to register all file types, but instead allow the user to select the file extensions that should be registered.

skylot:
> register Jadx to all supported file extensions

@jpstotz I am sure this handler not assign/register any file types, it is just a macOS specific way to open files in apps.

Also, I saw same [requirements](https://www.jdeploy.com/docs/manual/#_accessing_files_in_java) in [jdeploy](https://www.jdeploy.com/) dev guide, which I plan to use in next jadx version. So I was going to do this change anyway :slightly_smiling_face:

skylot (approved):

chaos-cn:
> > register Jadx to all supported file extensions
> 
> @jpstotz I am sure this handler not assign/register any file types, it is just a macOS specific way to open files in apps.
> 
> Also, I saw same [requirements](https://www.jdeploy.com/docs/manual/#_accessing_files_in_java) in [jdeploy](https://www.jdeploy.com/) dev guide, which I plan to use in next jadx version. So I was going to do this change anyway 🙂

Thanks, I get this solution from [jd-gui](https://github.com/java-decompiler/jd-gui/blob/master/app/src/main/java/org/jd/gui/OsxApp.java).

Issue #1891, opened 2023-06-03
[cli] Passed `--mappings-path` gets immedeatly overriden


NebelNidas:
Here, the path is still present:
https://github.com/skylot/jadx/blob/e9a7ccd00c1a372ed6f6bdbf3701d6cb2bfa0a12/jadx-gui/src/main/java/jadx/gui/JadxWrapper.java#L64

Only to get overridden in `jadx.gui.settings.JadxProject#fillJadxArgs` two lines later:
- https://github.com/skylot/jadx/blob/e9a7ccd00c1a372ed6f6bdbf3701d6cb2bfa0a12/jadx-gui/src/main/java/jadx/gui/JadxWrapper.java#L66
- https://github.com/skylot/jadx/blob/e9a7ccd00c1a372ed6f6bdbf3701d6cb2bfa0a12/jadx-gui/src/main/java/jadx/gui/settings/JadxProject.java#L68

NebelNidas:
IMO, CLI args should have priority over the already present options

skylot:
> CLI args should have priority over the already present options

Agree, but only on first load, otherwise you will not be able open another mapping (it will be overwritten by cli args)
So, we need to update project params from cli args, and next always use project params.

BTW, looks like I forgot to move mapping path and other mapping arg to plugin options, this will also help to use more common approach for handling params.

PR #1925, opened 2023-06-21
Don't override passed mappings path

NebelNidas:
Quick & dirty fix for #1891. I know you want to move it to the plugin soon anyway, but at least for now it would be fixed.


skylot:
Ok, thanks.
I will increase priority of #1891 :slightly_smiling_face:

Issue #1189, opened 2021-06-16
[feature/bug] AndroidManifest.xml not updated when deobfuscating


iscle:
When deobfuscating an apk, the AndroidManifest.xml file does not get updated with the new file names generated by jadx. Having the correct file names would help a lot when trying to find stuff like Services, Activities, Receivers, etc.

MrIkso:
I tested it, it fixed, use latest jax version.
@skylot close it

PR #1927, opened 2023-06-22
fix(gradle): add apache http client support for api level 28+

nitram84:
I would like to propose some patches to fix legacy apache http client support in gradle files. To detect legacy apache http client usage the best way in my opinion is to use classpath information from jcst files. This has the advantage that all warnings like "Can't compare types, unknown hierarchy" should be gone for apks. My first idea was generating jcst files for each optional library. But this is a bad idea due to index based referencing of the jcst file format. To store the origin of classes I had to extend the jcst file format.

The first commit repairs the broken jcst generator.

In the second commit I extended the jcst file format, updated core to api level 34 and added all optional libraries (android.car and org.apache.http.legacy). With this PR the order of arguments matters when updating core.jcst. I updated core.jcst with this arguments:

`
<jadx_root>/jadx-core/src/main/resources/clst/core.jcst <sdk_root>/platforms/android-<api level>/android.jar <sdk_root>/platforms/android-<api level>/optional/android.car.jar <sdk_root>/platforms/android-<api level>/optional/org.apache.http.legacy.jar
`

For already decompiled apps it is important to clear jadx caches to rebuild usage info.

What is missing:
* Api level <28 is not yet supported because manifest needs patching (could be done together with #1189)
* Android car needs more testing: with a good test case gradle support for android car could be fixed very quickly.


nitram84:
@skylot Thank you for your feedback. At first I thought of a separate pass for all gradle relevant tasks, where classes have to be processed. But then, just for one simple check, I wasn't sure anymore. You are right, I should fix this, but it will take some time to test everything. Can you give me a good example for a full instructions scan as a reference? Thank you!

skylot:
> Can you give me a good example for a full instructions scan as a reference?

https://github.com/skylot/jadx/blob/a992c931980836be1fd87145cc8ee89e1d8d46ad/jadx-core/src/main/java/jadx/core/dex/visitors/usage/UsageInfoVisitor.java#L112-L118

Same as in UsageInfoVisitor. This instruction visitor allows to quickly traverse all insns and decode only required.
But unlike UsageInfoVisitor you need to check all methods references (IMethodRef) from invoke insns, not only 'resolved' (i.e. from current app).

> At first I thought of a separate pass for all gradle relevant tasks, where classes have to be processed. But then, just for one simple check, I wasn't sure anymore

Yeah, adjusting UsageInfoVisitor looks like the simplest way, but it requires too many changes in lots of places, so in result it is easier to just copy code to new visitor and change it to your needs.

Also, I want to mention, that now there is no way to nicely execute pass only for export. So I suggest adding new jadx pass type with name like `JadxBeforeExportPass` similar to existing [JadxAfterLoadPass](https://github.com/skylot/jadx/blob/7a309ca36797c0d6a3ceaed7f8490dec57798151/jadx-core/src/main/java/jadx/api/plugins/pass/types/JadxAfterLoadPass.java#L6C1-L6C1) and next add it [here](https://github.com/skylot/jadx/blob/683cd76cc555f38808450674ff2f8653f1aa48de/jadx-core/src/main/java/jadx/api/JadxDecompiler.java#L681), get all passes like [this](https://github.com/skylot/jadx/blob/683cd76cc555f38808450674ff2f8653f1aa48de/jadx-core/src/main/java/jadx/api/JadxDecompiler.java#L202) and run them before export.

nitram84:
Thank you! This will help me a lot!

PR #1928, opened 2023-06-23
Parse strings only when they are referenced

### LINKED ISSUE #1926: [core] Avoid reading the string pool eagerly
JADX-GUI gives this error when decoding [AndroidManifest.zip](https://github.com/skylot/jadx/files/11834818/AndroidManifest.zip).
```
Error decode manifest
java.io.IOException: Expected strings start, expected offset not reachable: 0x548, actual: 0x158c
	at jadx.core.xmlgen.ParserStream.skipToPos(ParserStream.java:132)
	at jadx.core.xmlgen.CommonBinaryParser.parseStringPoolNoType(CommonBinaryParser.java:29)
	at jadx.core.xmlgen.BinaryXMLParser.decode(BinaryXMLParser.java:110)
	at jadx.core.xmlgen.BinaryXMLParser.parse(BinaryXMLParser.java:82)
```

Upon further inspection, it seems to be a new obfuscation technique employed by packers to throw off analysis tools: `stringCount` in the string pool is higher than the actual number of strings used in the binary XML file, which means that `stringsOffset` contains more items than needed, overlapping with actual strings, as illustrated in this screenshot:
image

https://github.com/skylot/jadx/blob/afa331e8fc5c826d13f70d602c88c71b41056cda/jadx-core/src/main/java/jadx/core/xmlgen/CommonBinaryParser.java#L26-L29

By the time we reach L29, we have already read past `stringsStart`, which means that this check will fail.
However, the actual Android system seems to be happy with this, as long as you don't reference these extra fake "strings".

I suggest that we copy the strings chunk into an array, and attempt to parse strings only when they are referenced, to align with the actual behavior of Android systems.

chengyuhui:
Fix #1926.

Works well enough on all the APKs available locally to me (malicious or not), including the sample provided in the referenced issue.

BTW, this sample [AndroidManifest.xml.zip](https://github.com/skylot/jadx/files/11844183/AndroidManifest.xml.zip) uses another technique (an extremely large Unicode string in the place of namespace identifier, where Android doesn't care) to try to crash JADX, any idea on how we could defend against this?

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java
@@ -45,7 +45,7 @@ public class BinaryXMLParser extends CommonBinaryParser {
 	private final Map<String, String> tagAttrDeobfNames = new HashMap<>();
 
 	private ICodeWriter writer;
-	private String[] strings;
+	private BinaryXMLStrings strings;
 	private String currentTag = "ERROR";
 	private boolean firstElement;
 	private ValuesParser valuesParser;
@@ -387,8 +387,8 @@ public class BinaryXMLParser extends CommonBinaryParser {
 	}
 
 	private String getString(int strId) {
-		if (0 <= strId && strId < strings.length) {
-			return strings[strId];
+		if (0 <= strId && strId < strings.size()) {
+			return strings.get(strId);
 		}
 		return "NOT_FOUND_STR_0x" + Integer.toHexString(strId);
 	}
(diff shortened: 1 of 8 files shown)


jpstotz:
Thanks for this nice PR. I made some tests as I was interested how large the copied byte array sections will get and how effective the cache is. 

A common APK has 3 different BinaryXMLStrings instances, usually two larger and one small. In total the buffered byte array  of all three is between 1 MB and 5 MB. The used cache has a lot of cache hits when loading the APK file, so the the cache is really useful.

skylot (approved):

skylot:
@chengyuhui thank you for a nice fix :+1:

> ... try to crash JADX, any idea on how we could defend against this?

Maybe we can cut such strings similar to deobfuscation conditions for names in code (i.e. if longer of configured number).

PR #1931, opened 2023-07-01
fix(gradle): avoid BuildConfig conflict (#1847)

nitram84:
I changed my mind about implementing buildConfigFields to resolve BuildConfig conflicts (#1847):

```gradle
buildConfigField('Runnable', 'BACKGROUND_THREAD', 'new Runnable() { /* put all your code here */ public void run() { /* insert code with decompilation issues here */ }}')
```

This would create too many new problems. Here is a simple fix to resolve this issue. This setting will be the default in newer AGP versions anyway.


skylot (approved):

skylot:
Nice! Thanks :+1:

PR #1933, opened 2023-07-01
fix(res): resolve all custom attributes

nitram84:
My previous commit e1aa9f6 was not complete. I had to use more logic of apktool. My test case was https://apkpure.com/awb-rastatt/de.k4systems.awbrastatt (version 7.2.0). Gradle build now fails with a java issue.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
@@ -209,9 +209,10 @@ public class ManifestAttributes {
 			if (ri.getTypeName().equals("attr") && ri.getNamedValues().size() > 1) {
 				RawNamedValue first = ri.getNamedValues().get(0);
 				MAttrType attrTyp;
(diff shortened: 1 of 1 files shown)


skylot:
Ok, so you are just reverting 7f317be3, right?
I have no idea how to correct handle these types, but code looks better :rofl:

nitram84:
This is the reference for your review: https://github.com/iBotPeaches/Apktool/blob/f42ce82f0d4812bf490086b1247c33eb5ce23229/brut.apktool/apktool-lib/src/main/java/brut/androlib/res/data/value/ResAttr.java#L99-L106
In my first fix, when it was only one special case, I thought it would be nice to understand the types, but later after my other resource fixes, more failing types showed up and it became clear, that this is not the right way.
(btw: You can see in this references, that there is still more work needed: we do not support attributes like "max" and "localization" yet. I'll try to fix those issues, when I find apks to test with.)

skylot:
@nitram84 thanks. I'm just trying to make sure this change not break previous cases :slightly_smiling_face:

PR #1938, opened 2023-07-04
feat(lang): update zh_TW translation

5idereal:


skylot:
@5idereal thanks :+1:

Issue #1939, opened 2023-07-05
[feature] Close all tabs to the right


bytemarx:
Fairly simple feature that I find useful in other applications with tabs. I've already implemented this and have been using it for a while now but wanted to gauge the interest here about making it an official feature for everyone else.

skylot:
Implemented by @bytemarx in PR #1940

PR #1940, opened 2023-07-05
feat(gui): close all tabs to the right

bytemarx:
Implements #1939. GUI feature that adds a right-click option on tabs to close all tabs to the right of the selected tab.


bytemarx:
That naughty force-push was just me adding the issue reference in the commit.

skylot (approved):

skylot:
@bytemarx thanks!
Just as a note: you don't need to change variable/field type to `LinkedHashMap` just `Map` is fine here.

bytemarx:
> @bytemarx thanks! Just as a note: you don't need to change variable/field type to `LinkedHashMap` just `Map` is fine here.

Since the logic is dependent on the predictable iteration order provided by the `LinkedHashMap`, changing the declarations to such guarantees stability in this implementation. For example, if `openTabs` in `TabbedPane` is redefined as any other `Map` type, then this implementation will break.

PR #1947, opened 2023-07-09
fix(res): support formatted=false for strings

nitram84:
There is an error message in some cases when a percent sign (%) appears in a resources xml inside a string. The error message is e.g.:

`/app/src/main/res/values/strings.xml:143:4: Multiple substitutions specified in non-positional format of string resource string/certificate_person_details_card_expiration. Did you mean to add the formatted="false" attribute?`

Steps to reproduce:
Export this app to gradle and run a build: https://github.com/corona-warn-app/cwa-app-android / https://m.apkpure.com/de/corona-warn-app/de.rki.coronawarnapp

The problem is causes by strings like:
https://github.com/corona-warn-app/cwa-app-android/blob/d98f9c20c595ca5dc877f87f032b47d8b8175a79/Corona-Warn-App/src/main/res/values/covid_certificate_strings.xml#L34

Unlike the devs of SAP jadx does not place quotes around all strings. Apktool solves the issue by setting `formatted="false"`: https://github.com/iBotPeaches/Apktool/blob/2b8121584624f0a1b1602992a48fcdbf60c0347d/brut.apktool/apktool-lib/src/main/java/brut/androlib/res/data/value/ResStringValue.java#L59-L61

It's possible to borrow and reuse their implementation (https://github.com/iBotPeaches/Apktool/blob/2b8121584624f0a1b1602992a48fcdbf60c0347d/brut.apktool/apktool-lib/src/main/java/brut/androlib/res/xml/ResXmlEncoders.java#L139-L142
and 
https://github.com/iBotPeaches/Apktool/blob/2b8121584624f0a1b1602992a48fcdbf60c0347d/brut.apktool/apktool-lib/src/main/java/brut/androlib/res/xml/ResXmlEncoders.java#L163-L208 ) directly in jadx to solve this issue. I suppressed all warning instead of fixing them - with less changes it's easier to sync with upstream.

(Btw: Fixing all resource issues of this app is one change ahead: set compileSdkVersion to 31, buildToolsVersion to 31.0.0 and update android gradle plugin to e.g. 7.4.0 and all resource issues are fixed)

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ResXmlGen.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ResXmlGen.java
@@ -250,6 +250,11 @@ public class ResXmlGen {
 				cw.add(' ').add(attrName).add("=\"").add(attrValue).add('"');
 			}
 		}
+
+		if (itemTag.equals("string") && valueStr.contains("%") && StringFormattedCheck.hasMultipleNonPositionalSubstitutions(valueStr)) {
+			cw.add(" formatted=\"false\"");
+		}
+
 		if (valueStr.equals("")) {
 			cw.add(" />");
 		} else {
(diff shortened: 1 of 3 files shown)


skylot (approved):

skylot:
Looks good to me. Thanks!

Issue #1423, opened 2022-03-24
[feature] Goto launch activity shortcut


Mino260806:
Simply add a shortcut (for example under _Navigation_) that opens the main activity in Code, and opens it accordingly in files tree.

skylot:
Implemented by @Mino260806 in PR #1965.

PR #1965, opened 2023-07-28
Implemented feature: 'go to main activity'

Mino260806:
Suggested here #1423

### Description
Added a toolbar button and an action under navigation menu, that opens main activity class in a code tab


skylot:
@Mino260806 thank you for great work :+1:
I made a small changes to search main activity class by original name (not renamed alias) and also handle exceptions with additional info about failed step.

About left todos: looks like it is better to make a new class (in jadx-core) to handle all manifest parsing and output [ApplicationParams](https://github.com/skylot/jadx/blob/4217aab93322b0c76f7db9618bbb44041ef6c084/jadx-core/src/main/java/jadx/core/export/ApplicationParams.java#L3) as result, so it will be possible to use it both in export and main activity open. But this needs a lot of refactoring, so new PR is preferred.

skylot (approved):

Mino260806:
> looks like it is better to make a new class (in jadx-core)

Sure. In which package do you suggest to create the new class ?

skylot:
> Sure. In which package do you suggest to create the new class ?

`jadx.core.utils.android` will be fine, it already has various utility classes for work with android resources

PR #1971, opened 2023-07-31
refactor: reuse the same parser in main activity action and gradle export feature

Mino260806:
As explained in pr #1965 in this [comment](https://github.com/skylot/jadx/pull/1965#issuecomment-1656864227)

I had to change [this line](https://github.com/Mino260806/jadx/blob/refactor-manifest-parser/jadx-core/src/main/java/jadx/core/export/ExportGradleTask.java?plain=1#L51), ([original](https://github.com/skylot/jadx/blob/master/jadx-core/src/main/java/jadx/core/export/ExportGradleTask.java?plain=1#L50)), because for some reason, resources.getSubFiles() is returning the files in reverse order, so the first strings.xml that was found did not necessarily have app_name (necessary for applicationLabel).
In master branch, the resources.getSubFiles() returns the files in expected order, but I don't know what change I made causes the them to be reversed. 
Why do you think it happens ?


Mino260806:
Why is the build failing ?

jpstotz:
@Mino260806 See the build log, it is executing the unit tests and three of them are failing:

```
TestApacheHttpClient > test() FAILED
    jadx.core.utils.exceptions.JadxRuntimeException at TestApacheHttpClient.java:16
        Caused by: java.lang.NullPointerException at TestApacheHttpClient.java:16

> Task :jadx-core:checkstyleTest

> Task :jadx-core:test

OptionalTargetSdkVersion > test() FAILED
    jadx.core.utils.exceptions.JadxRuntimeException at OptionalTargetSdkVersion.java:13
        Caused by: java.lang.NullPointerException at OptionalTargetSdkVersion.java:13

VectorDrawablesUseSupportLibrary > test() FAILED
    jadx.core.utils.exceptions.JadxRuntimeException at VectorDrawablesUseSupportLibrary.java:16
        Caused by: java.lang.NullPointerException at VectorDrawablesUseSupportLibrary.java:16
```

jpstotz:
Just execute the tests yourself, they should also fail in your environment.

Mino260806:
> In master branch, the resources.getSubFiles() returns the files in expected order, but I don't know what change I made causes the them to be reversed. Why do you think it happens ?

It's the same problem I described earlier, in my pr ExportGradleTask is returning a wrong appStrings, I didn't change anything in that can disrupt it. I need some help

jadx-core/src/main/java/jadx/core/utils/android/AndroidManifestParser.java
```diff
@@ -0,0 +1,243 @@
...
+					if (stringName.equals(appLabelName)) {
+						return strings.item(i).getTextContent();
+					}
+				}
+			} else {
+				return appLabelName;
+			}
+		}
+
+		return null;
```
skylot, line 143: Previous implementation return "UNKNOWN" if value not found, now it returns null and cause tests to fail.

skylot:
> Should I keep that change?

Sure, it is always better to use more strict filter. And thanks for explanation!

Mino260806:
> Other suggestions:
> 
> * instead int flags using enum with EnumSet is a better way in Java.
> * I think we can move ApplicationParams class to package with new AndroidManifestParser, because it is no longer tied to export
> * (optional) it is better to avoid running code in constructor (especially if it may throw exception). Just run all operations in `ApplicationParams parse()` method, it will be easier to wrap it in try/catch to handle errors.

Anything else ?

skylot (approved):

skylot:
Great. Thank you :+1:

Issue #1948, opened 2023-07-10
[feature] Option to configure line breaks


NebelNidas:
Currently, JADX [hardcodes the the system default](https://github.com/skylot/jadx/blob/ce9645864d9a93705717c408ecb3b2432e4f3502/jadx-core/src/main/java/jadx/api/ICodeWriter.java#L11) for line breaks. In certain situations, it may be desirable to define a specific one though. This is something I wouldn't necessarily expose via `JadxArgs`, but it should be possible to overwrite the default value in `ICodeWriter` by using a custom implementation thereof. If you could change `NL` from being a field to an interface method, like `nl()`, which can be overridden, that would be much appreciated.

PR #1972, opened 2023-07-31
fix(core): fix ResXmlGenTest fails on Windows

jpstotz:
The ResXmlGenTest always failed on Windows because of paths using backslash instead of slash and generated XML content containing \r\n but only \n was expected.

### DIFF
--- a/jadx-core/src/test/java/jadx/core/xmlgen/ResXmlGenTest.java
+++ b/jadx-core/src/test/java/jadx/core/xmlgen/ResXmlGenTest.java
@@ -196,7 +196,7 @@ class ResXmlGenTest {
 	}
 
 	private static String adaptLineEndings(String input) {
-		if (System.lineSeparator().equals("\n")){
+		if (System.lineSeparator().equals("\n")) {
 			return input; // no adaption necessary
 		}
 		return input.replaceAll(System.lineSeparator(), "\n");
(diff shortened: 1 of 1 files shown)


skylot:
Oh, thanks.
Look like we really need to implement line break options (already requested in #1948) and set it to fixed value for tests.
Also, I will add windows build host for tests in github workflows.

skylot (approved):

skylot:
@jpstotz I added tests run on windows host (868e99e), but also I add [`.gitattributes`](https://www.git-scm.com/docs/gitattributes) file to always use unix line breaks. So looks like we don't need [part about autocrlf option](https://github.com/skylot/jadx/wiki/Code-Formatting#windows-and-git-autocrlftrue) in code format wiki. And it will be nice if you can verify that my changes not broke/interfere with git autocrlf option :slightly_smiling_face:

jpstotz:
@skylot Great, that simplifies the whole process for Jadx developers on Windows. Just tried it, performed a fresh git clone (my default git CRLF settings is the system default) and all files were cloned using Linux (LF) only mode. Executing `gradlew spotlessApply` on this repo (which previously caused all line endings to be converted to LF which means all files had changes) and a `git diff` afterwards did not show any changes. Therefore the `.gitattributes` definition seem to work well. I did not know that git provides such an easy way for solving the line ending problem, thanks.

Issue #1721, opened 2022-11-05
Restore the original shortcut key[feature]


javaeryang:
*Describe your idea:*
Ctrl + Alt + left   ===> back
Ctrl + Alt + right   ===> forward

Used for a long time
Please

j4k0xb:
I would rather prefer <kbd>alt</kbd>+<kbd>left</kbd> and <kbd>alt</kbd>+<kbd>right</kbd> because it's so common in other programs (browsers, IDEs, file explorers, ...)

javaeryang:
@j4k0xb It would be better if we could set it ourself

skylot:
Now is possible to change these shortcuts. The feature was implemented by @Mino260806 in PR #1980.

Issue #1718, opened 2022-11-03
BUG: jadx-gui: invalid mapping for horizontal scroll wheel and forward mouse button


S-trace:
Hello.

I'm using jadx-gui v1.4.5.
I have a Logitech MX Master 3 mouse with a horizontal scroll wheel and forward/back thumb buttons.

If I use the horizontal scroll wheel to scroll right or left with the mouse cursor over the code window - jadx-gui does not scroll the code window but reacts as if I pressed the "forward"/"back" mouse buttons (back for left-to-right scroll, forward for right-to-left scroll).

If I press the "back" mouse button - jadx-gui works as it should, and shows the previous open tab

If I press the  "forward" mouse button - jadx-gui does not react to it.

This is confusing.

Thank you.

skylot:
Unfortunately, I don't find any method in java Swing/AWT to get actions bound to mouse buttons. So jadx-gui using raw mouse button numbers ([here](https://github.com/skylot/jadx/blob/93ab0fd80bc5ad81781647f5c5f5f43fa4babd48/jadx-gui/src/main/java/jadx/gui/ui/MainWindow.java#L1558-L1566)), and this approach is very limited. Only way to fix that is to allow user to change keybindings (already requested in #1479). Anyway, I will try to search any other solution, maybe I missed something :+1:

oxodao:
Just discovered jadx recently and having a lot of fun with it. Unfortunately this issue is really bugging me as the sensitivity on the MX Master (1) seems to be quite high. Thank you for this piece of software anyway it's really great to have a good free alternative to JEB

Mino260806:
About this
> If I use the horizontal scroll wheel to scroll right or left with the mouse cursor over the code window - jadx-gui does not scroll the code window but reacts as if I pressed the "forward"/"back" mouse buttons (back for left-to-right scroll, forward for right-to-left scroll).

The issue is here
```
private void registerMouseNavigationButtons() {
	Toolkit toolkit = Toolkit.getDefaultToolkit();
	toolkit.addAWTEventListener(event -> {
		if (event instanceof MouseEvent) {
			MouseEvent mouseEvent = (MouseEvent) event;
			if (mouseEvent.getID() == MouseEvent.MOUSE_PRESSED) {
				int rawButton = mouseEvent.getButton();
				if (rawButton <= 3) {
					return;
				}
				int button = remapMouseButton(rawButton);
				switch (button) {
					case 4:
						tabbedPane.navBack();
						break;
[...]

Issue #1609, opened 2022-08-07
[feature] keyboard shortcut for saving project


Bleuzen:
*Describe your idea:*

So usually in every program the shortcut `Ctrl + S` saves the project. But in JADX this standardized shortcut does... save decompiled sources. And there is no shortcut at all to save the current project.

A simple solution for now could be to move the `Ctrl + S` shortcut from saving sources to saving the project.
Here is why (extremely scientific me study):
- saving decompiled sources is an action one most likely won't need often, maybe only once for an app
- saving the project is an action used much more often (for me after every change so sometimes every few seconds) -> makes more sense to have the shortcut here, not having one is purely annoying

If you don't want to 'just move' the shortcut maybe assign a different one to saving the decompiled sources.
But `Ctrl + S` is a standard for saving the project and I suggest JADX doing the same.

skylot:
Sure! Looks like I totally forgot about #1085, it already has that mapping :rofl:
And for `Save source` I want to set shortcut to <kbd>CTRL</kbd> + <kbd>E</kbd> after merging it with export, but for now <kbd>CTRL</kbd> + <kbd>SHIFT</kbd> + <kbd>E</kbd> is also fine.

skylot:
Not is possible to change these shortcuts. The feature was implemented by @Mino260806 in PR #1980.

Issue #1479, opened 2022-05-08
[feature] Let the user define the keyboard shortcuts


komen205:
I'd love to be able to change my keyboard shortcuts. And I think this would benefit other people.

For example: to go back the default key is the "escape" however it'd be nice to change it to Control + Z. Which is the key that most IDE's use.

yfy-1997:
it's a good idea!

skylot:
Shortcuts customization implemented by @Mino260806 in PR #1980.

PR #1980, opened 2023-08-02
feat(gui): allow user to set custom shortcuts

Mino260806:
Supports keyboard keys and mouse buttons

I kept the default shortcuts for each action
<details>
<summary>Toolbar / Menu shortcuts</summary>

image

</details>

<details>
<summary>Code area</summary>

image

</details>

I created some new classes, please let me know if I should move any to another package

Also, I would like to know if there are other shortcuts that need to be customized. Currently, I added Menu / Toolbar actions and CodeArea.

### A small question

https://github.com/skylot/jadx/blob/868e99e0b57b9efade154317c4e8fadfd8ca2408/jadx-gui/src/main/java/jadx/gui/ui/MainWindow.java#L1125

What is the significance of `SystemInfo.IS_MAC` argument here ? There's no way to keep it to make it customizable like other shortcuts

### Related issues
#1721
#1718
#1609
#1479
#1085


skylot:
Wow, this is big PR, it will take some time for me to check it.

> What is the significance of SystemInfo.IS_MAC argument here ?

Looks like there is some difference in registering modifiers in macOs, so this keystoke was set to run action on key release, check PR #1301 for details. To keep this behavior, we should set `onKeyRelease` to true for all actions with modifiers on macOs.
Also, macOs instead `Ctrl` often uses `Command` buttons in shortcuts (like `Ctrl+F`), check issue #165 and fix in #616. So we should set correct modifier (as initial value) for these actions using `jadx.gui.utils.UiUtils.ctrlButton()` method.

> Also, I would like to know if there are other shortcuts that need to be customized

After quick search for `getKeyStroke(` in project, I think, it worth adding 'script run (F8)' in 'ScriptContentPanel' and auto complete key (`ctrl+space`) in 'ScriptCodeArea' 

And some issues I notice:
- for some reason, any `F<number>` keys cannot be set, because field continue to wait input.
- I think we shouldn't accept left mouse button at all (just ignore it). I agreed to the prompt (it doesn't show the key that will be set) and was not able to do anything after that :rofl:
- changes in JNodeAction class, lead to ignoring node under mouse and always use node under cursor, but these can be different locations
- it is nice to show confirmation about using duplicate shortcut, and show name of other action with same key in confirm (so it is possible to go and change it to something different)

Mino260806:
> To keep this behavior, we should set `onKeyRelease` to true for all actions with modifiers on macOs

Okay, I was confused because it was only used in the line I linked

> Also, macOs instead Ctrl often uses Command buttons in shortcuts (like Ctrl+F), check issue https://github.com/skylot/jadx/issues/165 and fix in https://github.com/skylot/jadx/pull/616. So we should set correct modifier (as initial value) for these actions using jadx.gui.utils.UiUtils.ctrlButton() method.

Yeah I'm already using it (see [ActionModel](https://github.com/Mino260806/jadx/blob/6bef0fe7d2f748b33ffc7bc41bc8cf5abdbbd1a7/jadx-gui/src/main/java/jadx/gui/ui/menu/ActionModel.java) enum)

> for some reason, any F<number> keys cannot be set, because field continue to wait input

It's normal. You probably have Fn lock enabled, trying hold Fn while pressing the F\<number\>

> changes in JNodeAction class, lead to ignoring node under mouse and always use node under cursor, but these can be different locations

I don't understand this one. How can I reproduce this issue ?

Mino260806:
> To keep this behavior, we should set onKeyRelease to true for all actions with modifiers on macOs

I'm unsure if this is what you meant https://github.com/skylot/jadx/pull/1980/commits/db182ee5b3ce63833f197d0e746fd7fd6ab888a9

jadx-gui/src/main/java/jadx/gui/ui/codearea/JNodeAction.java
```diff
@@ -31,21 +30,12 @@ public boolean isActionEnabled(@Nullable JNode node) {
...
-		});
-	}
-
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		runAction(node);
+		node = codeArea.getNodeUnderCaret();
+		if (isActionEnabled(node)) {
+			runAction(node);
+		}
```
skylot, line 38:
```suggestion
		if (JadxGuiAction.COMMAND.equals(e.getActionCommand())) {
			node = codeArea.getNodeUnderCaret();
			if (isActionEnabled(node)) {
				runAction(node);
			}
		} else {
			runAction(node);
		}
```

jadx-gui/src/main/java/jadx/gui/settings/ui/shortcut/ShortcutEdit.java
```diff
@@ -0,0 +1,163 @@
...
+					Shortcut pressedShortcut = Shortcut.keyboard(ev.getKeyCode(), ev.getModifiersEx());
+					if (pressedShortcut.isValidKeyboard()) {
+						tempShortcut = pressedShortcut;
+						refresh(tempShortcut);
+					} else {
+						tempShortcut = null;
+					}
+				} else if (ev.getID() == KeyEvent.KEY_TYPED) {
+					removeFocus();
+				}
```
skylot, line 83: I debug issue with `F*` keys and turns out it doesn't trigger KEY_TYPED event, so focus not removed.
Do we really need to listen KEY_TYPED event, why we can't remove focus in KEY_PRESSED?

skylot (changes requested):

jadx-gui/src/main/java/jadx/gui/ui/codearea/JNodeAction.java
skylot: We don't need to use getNodeUnderCaret if action was triggered by mouse (i.e popup menu)

jadx-gui/src/main/java/jadx/gui/settings/ui/shortcut/ShortcutEdit.java
Mino260806: Good point.

skylot:
Hm, I know why you use KEY_TYPED event: to catch several keys pressed at the same time, like `ctrl+R`. Now this is not working :cry:

Mino260806:
> Hm, I know why you use KEY_TYPED event: to catch several keys pressed at the same time, like `ctrl+R`. Now this is not working So, we should keep focus until non modifier key is pressed.

Or I'll just use KEY_RELEASED instead of KEY_TYPED to remove focus

skylot:
Ok, everything looks good now :+1:

skylot (approved):

Mino260806:
There is still some stuff I need to fix, please don't merge until I confirm

skylot:
> There is still some stuff I need to fix, please don't merge until I confirm

Next time, mark PR as draft if it is not ready for merge :slightly_smiling_face:

Also, I just notice that we can't have both keys and mouse shortcuts for navigation (#1718 case), only one not both :sob:

Mino260806:
> Also, I just notice that we can't have both keys and mouse shortcuts for navigation (#1718 case), only one not both

Yeah, sadly. Do you have an idea in mind to get it back along the current changes ? Maybe I add a checkbox under preferences to have mouse shortcuts bind to navigation ?

skylot:
> Maybe I add a checkbox under preferences to have mouse shortcuts bind to navigation ?

Not sure what you mean, but you can add another action enum which will call same code :rofl:
Anyway, without allowing several shortcuts for same action, it is just a workaround.

Mino260806:
I struggled to make the dummy actions get called without being added to any component.
So I had to create [this small class](https://github.com/Mino260806/jadx/blob/e44dfb38232cae6ebca63d4b2027e4f0d480fe8f/jadx-gui/src/main/java/jadx/gui/ui/menu/HiddenMenuItem.java) as a workaround

Mino260806:
Can you test the shortcuts in script panel ? I don't know how to access it

Other than that I think the PR is ready to be merged now

skylot:
Well, it works fine after resetting settings :rofl:
But without reset, new actions are missing and cause NPE
```
java.lang.IllegalArgumentException: trigger key cannot be null
	at org.fife.ui.autocomplete.AutoCompletion.setTriggerKey(AutoCompletion.java:1174)
	at jadx.gui.plugins.script.ScriptCodeArea.addAutoComplete(ScriptCodeArea.java:37)
	at jadx.gui.plugins.script.ScriptCodeArea.<init>(ScriptCodeArea.java:29)
	at jadx.gui.plugins.script.ScriptContentPanel.<init>(ScriptContentPanel.java:59)
	at jadx.gui.treemodel.JInputScript.getContentPanel(JInputScript.java:41)
	at jadx.gui.ui.TabbedPane.getContentPanel(TabbedPane.java:404)
	at jadx.gui.ui.TabbedPane.showNode(TabbedPane.java:272)
	at jadx.gui.ui.MainWindow.nodeClickAction(MainWindow.java:804)
	at jadx.gui.ui.MainWindow$6.mousePressed(MainWindow.java:1168)
```
So, during load, default shortcuts values should be used if action missing in loaded settings.

To check scripts actions in left tree go to `Inputs->Scripts` right click and in popup select `New script` and save it, next open it to see a script's panel above editor.

Mino260806:
> But without reset, new actions are missing and cause NPE

Issue noted, I'll fix it tomorrow

Mino260806:
Ready for merge

jadx-gui/src/main/java/jadx/gui/plugins/script/ScriptContentPanel.java
```diff
@@ -89,15 +90,13 @@ private void initUI() {
...
-		runAction.attachKeyBindingFor(scriptArea, KeyStroke.getKeyStroke(KeyEvent.VK_F8, 0));
-
-		ActionHandler saveAction = new ActionHandler(scriptArea::save);
-		saveAction.setNameAndDesc(NLS.str("script.save"));
-		saveAction.setIcon(Icons.SAVE_ALL);
-		saveAction.setKeyBinding(KeyStroke.getKeyStroke(KeyEvent.VK_S, UiUtils.ctrlButton()));
+		JadxGuiAction runAction = new JadxGuiAction(ActionModel.SCRIPT_RUN, this::runScript);
+		JadxGuiAction saveAction = new JadxGuiAction(ActionModel.SCRIPT_SAVE, scriptArea::save);
+
+		runAction.setShortcutComponent(scriptArea);
```
skylot, line 96:
```suggestion
```
This line prevents action execution on button click

Mino260806: No, it's responsible for enabling mouse shortcuts. I'll fix it

jadx-gui/src/main/java/jadx/gui/plugins/context/CodePopupAction.java
```diff
@@ -34,12 +34,13 @@ private static class NodeAction extends JNodeAction {
 
 		public NodeAction(CodePopupAction data, CodeArea codeArea) {
 			super(data.name, codeArea);
+			setShortcutComponent(codeArea);
```
skylot, line 37: This action class created from plugins and doesn't have associated ActionModel enum, so actionModel field set to null and cause NPE:
```java
java.lang.NullPointerException: Cannot invoke "jadx.gui.ui.action.ActionModel.getDefaultShortcut()" because "actionModel" is null
	at jadx.gui.settings.data.ShortcutsWrapper.get(ShortcutsWrapper.java:16)
	at jadx.gui.utils.shortcut.ShortcutsController.get(ShortcutsController.java:54)
	at jadx.gui.utils.shortcut.ShortcutsController.bindImmediate(ShortcutsController.java:79)
	at jadx.gui.ui.codearea.JNodePopupBuilder.add(JNodePopupBuilder.java:32)
	at jadx.gui.plugins.context.CommonGuiPluginsContext.appendPopupMenus(CommonGuiPluginsContext.java:71)
	at jadx.gui.ui.codearea.CodeArea.addMenuItems(CodeArea.java:133)
	at jadx.gui.ui.codearea.CodeArea.<init>(CodeArea.java:56)
	at jadx.gui.ui.codearea.ClassCodeContentPanel.<init>(ClassCodeContentPanel.java:48)
	at jadx.gui.treemodel.JClass.getContentPanel(JClass.java:119)
	at jadx.gui.ui.TabbedPane.getContentPanel(TabbedPane.java:404)
	at jadx.gui.ui.TabbedPane.showCode(TabbedPane.java:263)
	at jadx.gui.ui.TabbedPane.codeJump(TabbedPane.java:251)
	at jadx.gui.ui.TabbedPane.codeJump(TabbedPane.java:236)
	at jadx.gui.ui.MainWindow.nodeClickAction(MainWindow.java:801)
```
Can you mark `getActionModel()` as nullable and check all its use places?
Or maybe somehow opt out this class from using new shortcuts model at all.

skylot: To reproduce this NPE, add [this script](https://github.com/skylot/jadx/blob/683cd76cc555f38808450674ff2f8653f1aa48de/jadx-plugins/jadx-script/examples/scripts/gui/custom_frida.jadx.kts#L15) to jadx project and try to open any class.

jadx-gui/src/main/java/jadx/gui/plugins/script/ScriptContentPanel.java
skylot: Fixed.

jadx-gui/src/main/java/jadx/gui/plugins/context/CodePopupAction.java
skylot: Also fixed. Thanks :+1:

skylot:
Ok, I think it is fine to merge. Hope we don't get many regressions :smiley:

Mino260806:
> Ok, I think it is fine to merge. Hope we don't get many regressions

It's fine. You can ping me in an issue related to this PR and I'll fix it

skylot:
@Mino260806 sure :rofl:
And thank you for your help and hard work :+1:

Issue #1981, opened 2023-08-03
java.lang.IndexOutOfBoundsException: Index: 10, Size: 9


syxxjujing:
Please describe what you did before the error occurred.
**IMPORTANT!** If the error occurs with a specific APK file please attach or provide link to apk file!
 * Jadx version: 1.4.7
* Java version: 1.8.0_144
* Java VM: Oracle Corporation Java HotSpot(TM) 64-Bit Server VM
* Platform: Mac OS X (10.16 x86_64)
* Max heap size: 32768 MB
* Program args: -Xms128M -Xmx32g -XX:+UseG1GC -Dawt.useSystemAAFontSettings=lcd -Dswing.aatext=true -Djava.util.Arrays.useLegacyMergeSort=true

```
java.lang.IndexOutOfBoundsException: Index: 10, Size: 9
	at java.util.ArrayList.rangeCheck(ArrayList.java:653)
	at java.util.ArrayList.get(ArrayList.java:429)
	at jadx.gui.device.debugger.RegisterObserver.merge(RegisterObserver.java:41)
	at jadx.gui.device.debugger.DebugController.lambda$null$15(DebugController.java:878)
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
	at jadx.gui.device.debugger.DebugController.lambda$updateAllInfo$16(DebugController.java:877)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)

```

jpstotz:
There is an issue which is very similar (not identical) #1841. Not sure if both issues are really related. May be the changes implemented for v1.4.7 make the difference. If both problems have the same source, in the end the problem seems to be still present.

At a first glace this looks like a multi-threading/synchronization problem, but that is impossible as the affected ArrayList is  created directly in the affected `merge` method. 

Therefore it seems like the arguments supplied to `merge`: `rtRegs` and `smaliRegs` seem to contain compatible lists of registers. `rtRegs` seems to contain a reference to a register that does not exist in the `smaliRegs` list. 

What would be possible is that the list of smali registers is not correct in the dex code or the dex code did not matched the executed APK on the phone?
@syxxjujing can you provide the APK file you was loaded in Jadx and that you were debugging?

Issue #1841, opened 2023-04-24
java.lang.IndexOutOfBoundsException: Index 6 out of bounds for length 2


bluegatar:
Please describe what you did before the error occurred.
**IMPORTANT!** If the error occurs with a specific APK file please attach or provide link to apk file!
 * Jadx version: 1.4.6
* Java version: 17.0.6
* Java VM: Oracle Corporation Java HotSpot(TM) 64-Bit Server VM
* Platform: Mac OS X (13.3.1 x86_64)
* Max heap size: 45888 MB
* Program args: -Xms128M -XX:MaxRAMPercentage=70.0 -XX:+UseG1GC -Dawt.useSystemAAFontSettings=lcd -Dswing.aatext=true -Djava.util.Arrays.useLegacyMergeSort=true

```
java.lang.IndexOutOfBoundsException: Index 6 out of bounds for length 2
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:266)
	at java.base/java.util.Objects.checkIndex(Objects.java:359)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at jadx.gui.device.debugger.RegisterObserver.getInfo(RegisterObserver.java:81)
	at jadx.gui.device.debugger.DebugController.lambda$fetchAllRegisters$12(DebugController.java:675)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)

```

jpstotz:
Please update to Jadx 1.4.7. This new version contains several improvements on the debugger view which might solve this problem. (see #1796).

PR #1982, opened 2023-08-03
Understanding and may be fixing IndexOutOfBoundsException in the debugger

jpstotz:
Related issues: #1981 and #1841

This MR contains changes to the debugger to simplify the code to make it easier to understand (it used a lot of generic containers like SimpleEntry).

Also a central method was created to centralize the code that is going wrong in the referenced issues, not yet sure how to finally fix this...

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/device/debugger/ArtAdapter.java
+++ b/jadx-gui/src/main/java/jadx/gui/device/debugger/ArtAdapter.java
@@ -2,7 +2,7 @@ package jadx.gui.device.debugger;
 
 public class ArtAdapter {
 
-	public interface Debugger {
+	public interface IArtAdapter {
 		int getRuntimeRegNum(int smaliNum, int regCount, int paramStart);
 
 		boolean readNullObject();
@@ -10,7 +10,7 @@ public class ArtAdapter {
 		String typeForNull();
 	}
 
-	public static Debugger getAdapter(int androidReleaseVer) {
+	public static IArtAdapter getAdapter(int androidReleaseVer) {
 		if (androidReleaseVer <= 8) {
 			return new AndroidOreoAndBelow();
 		} else {
@@ -18,7 +18,7 @@ public class ArtAdapter {
 		}
 	}
 
-	public static class AndroidOreoAndBelow implements Debugger {
+	public static class AndroidOreoAndBelow implements IArtAdapter {
 		@Override
 		public int getRuntimeRegNum(int smaliNum, int regCount, int paramStart) {
 			int localRegCount = regCount - paramStart;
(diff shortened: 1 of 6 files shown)


skylot:
Looks good to me. It sure uses too many nested generics :rofl:
And exception now will provide better message.

Also, I think root cause is a weird way to calculate register's number, starts here:
https://github.com/skylot/jadx/blob/912c431511b3b71550889f03c4f421759bafff63/jadx-gui/src/main/java/jadx/gui/device/debugger/DebugController.java#L897
and next uses different methods for different Android versions: 
https://github.com/skylot/jadx/blob/4705194a1d96c54f584d013ff4fb57e530323680/jadx-gui/src/main/java/jadx/gui/device/debugger/ArtAdapter.java#L21-L54
Maybe some versions not covered or mapping was changed again :cry:

jpstotz:
I just realized that this PR is still incomplete. 

@skylot I agree that the way to calculate register's number may be root cause. Therefore it makes sense to print the used ArtAdapter in the error message. Together with some details on the used device for debugging, may be we are able to better understand what environments are causing this problem. Unfortunately the device information available in the Jadx debugger wasn't accessible in `RegisterObserver` class, and I don't know how helpful they are in this case. 

So next time we have to ask issue creator for more device details (may be `adb shell getprop` data?).

skylot:
@jpstotz thanks for improvement :+1:

> Unfortunately the device information available in the Jadx debugger wasn't accessible in RegisterObserver class, and I don't know how helpful they are in this case.

I think we need at least Android/SDK version and method signature which cause issue, and ask for apk sample.

> So next time we have to ask issue creator for more device details (may be adb shell getprop data?).

Not sure which props we need, because if issue is vendor specific we will not be able to reproduce it. Anyway, it is worth to try.

skylot:
@jpstotz I commit a change to move `DeviceSettings` class from `ADBDialog`, it was stored in a static field there, so I convert it to singleton. And now it is easier to use its data. Error message (for Google emulator) will contains device info and android version:
```
Register 6 does not exist (size: 5).
 Device: emulator-5554 device product:sdk_gphone64_x86_64 model:sdk_gphone64_x86_64 device:emulator64_x86_64_arm64 transport_id:1, Android: 12, ArtAdapter: AndroidPieAndAbove
 Method: com.jujing.yphook.LoginActivity.onCreate(Landroid/os/Bundle;)V
```
Also, during testing I saw many various issues:
- `Debug error: (JDWP Error Code:35) Invalid slot. for v1`
- values for method args not shown and actual registers list contains weird entries
- occasional UI update exceptions

Maybe first two related to our issue  :rofl:

skylot (approved):

PR #2018, opened 2023-09-18
Add Indonesian translation

roniragilimankhoirul:

### Description
Add Indonesian translation for jadx-gui


skylot:
@roniragilimankhoirul thanks :+1:

Although, looks like jadx not correctly handle old/obsolete language codes conversion done by Java Locale class.
I commit a fix to always use new codes taken from toLanguageTag method.

For details check docs https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Locale.html:
> Legacy language codes
Locale's constructor has always converted three language codes to their earlier, obsoleted forms: he maps to iw, yi maps to ji, and id maps to in. This continues to be the case, in order to not break backwards compatibility.
The APIs added in 1.7 map between the old and new language codes, maintaining the old codes internal to Locale (so that getLanguage and toString reflect the old code), but using the new codes in the BCP 47 language tag APIs (so that toLanguageTag reflects the new one). This preserves the equivalence between Locales no matter which code or API is used to construct them.

skylot (approved):

PR #2020, opened 2023-09-22
Pixeebot / Introduced protections against "zip slip" attacks 

zcarroll4:
### Description
This change updates all new instances of [ZipInputStream](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/zip/ZipInputStream.html) to protect against malicious entries that attempt to escape their "file root" and overwrite other files on the running filesystem.

Normally, when you're using `ZipInputStream` it's because you're processing zip files. That code might look like this:

```java
File file = new File(unzipTargetDirectory, zipEntry.getName()); // use file name from zip entry
InputStream is = zip.getInputStream(zipEntry); // get the contents of the zip entry
IOUtils.copy(is, new FileOutputStream(file)); // write the contents to the provided file name
```

This looks fine when it encounters a normal zip entry within a zip file, looking something like this pseudo-data:
```binary
path: data/names.txt
contents: Zeus\nHelen\nLeda...
```

However, there's nothing to prevent an attacker from sending an evil entry in the zip that looks more like this:
```binary
path: ../../../../../etc/passwd
contents: root::0:0:root:/:/bin/sh
```

Yes, in the above code, which looks like [every](https://stackoverflow.com/a/23870468) [piece](https://stackoverflow.com/a/51285801) of [zip-processing](https://kodejava.org/how-do-i-decompress-a-zip-file-using-zipinputstream/)  code you can [find](https://www.tabnine.com/code/java/classes/java.util.zip.ZipInputStream) on the [Internet](https://www.baeldung.com/java-compress-and-uncompress), attackers could overwrite any files to which the application has access. This rule replaces the standard `ZipInputStream` with a hardened subclass which prevents access to entry paths that attempt to traverse directories above the current directory (which no normal zip file should ever do.) Our changes end up looking something like this:

```diff
+ import io.github.pixee.security.ZipSecurity;
  ...
- var zip = new ZipInputStream(is, StandardCharsets.UTF_8);
+ var zip = ZipSecurity.createHardenedInputStream(is, StandardCharsets.UTF_8);
```

Powered by: [pixeebot](https://docs.pixee.ai/installing/) (codemod ID: [pixee:java/harden-zip-entry-paths](https://docs.pixee.ai/codemods/java/pixee_java_harden-zip-entry-paths))

### DIFF
--- a/buildSrc/src/main/kotlin/jadx-java.gradle.kts
+++ b/buildSrc/src/main/kotlin/jadx-java.gradle.kts
@@ -23,6 +23,8 @@ dependencies {
 	testRuntimeOnly("org.junit.platform:junit-platform-launcher")
 
 	testCompileOnly("org.jetbrains:annotations:24.0.1")
+	
+	implementation("io.github.pixee:java-security-toolkit:1.0.7")
 }
 
 repositories {
--- a/jadx-core/src/main/java/jadx/core/clsp/ClsSet.java
+++ b/jadx-core/src/main/java/jadx/core/clsp/ClsSet.java
@@ -1,5 +1,6 @@
 package jadx.core.clsp;
 
+import io.github.pixee.security.ZipSecurity;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.DataInputStream;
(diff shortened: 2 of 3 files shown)


jpstotz:
@zcarroll4 Thanks for the PR. But are you sure Jadx is still vulnerable to such attacks? Because we already have a ZIP entry path traversal check in place. 

Our implementation is just not integrated into ZipInputStream https://github.com/skylot/jadx/blob/master/jadx-core/src/main/java/jadx/api/plugins/utils/ZipSecurity.java

skylot:
@zcarroll4 as @jpstotz mention jadx already use similar zip security utils (with additional checks to prevent zip bomb attacks #980). Changes in this PR done in code which used only in jadx dev tools and not used in normal jadx operation. Anyway, to unify work with zip files, I will fix this case to use jadx utils in later commit.
As a result, I don't see any changes in this PR which can be used, so I will close it.

skylot:
Update:
Turns out code for work with zip files in `ClsSet` class was not needed at all, so I just remove it (4d87b44) :rofl:

PR #2024, opened 2023-10-06
chore: add resource entry offset check

jpstotz:
While implementing the changes of https://github.com/skylot/jadx/pull/1807 in a different project I noticed that in Jadx offset calculation for sparse entries was wrong but without effect as the offset was never used (Jadx assumes all entries can be read correctly ordered right after another in the stream).

This PR corrects the offset calculation and adds additional offset checks.


skylot (approved):

skylot:
@jpstotz thanks :+1:

PR #2031, opened 2023-10-10
fix: use fully qualified package names for classes of defpackage

### LINKED ISSUE #2027: [feature] import of classes under the defpackage package
### Describe your idea

When using a class with an empty package name, you cannot jump to the corresponding class because no fully qualified class name is used and no import package is displayed 

for example：

    private  z98 z98 = null;

After optimization：

    private  defpackage.z98 z98 = null;

thanks

rawer886:
fix #2027


skylot:
@rawer886 I think it is better to add correct import to fix this issue.
If you like full class names, you can disable `Use import statements` option in preferences.
I added test cases for several options and adjust code to pass these tests.
Anyway, thank you for issue report and your contribution :+1:

rawer886:
👍🏻

PR #2032, opened 2023-10-10
Introduced protections against predictable RNG abuse

zcarroll4:
### Description
This change replaces all new instances of `java.util.Random` with the marginally slower, but much more secure `java.security.SecureRandom`.

We have to work pretty hard to get computers to generate genuinely unguessable random bits. The `java.util.Random` type uses a method of pseudo-random number generation that unfortunately emits fairly predictable numbers.

If the numbers it emits are predictable, then it's obviously not safe to use in cryptographic operations, file name creation, token construction, password generation, and anything else that's related to security. In fact, it may affect security even if it's not directly obvious.

Switching to a more secure version is simple and our changes all look something like this:

```diff
- Random r = new Random();
+ Random r = new java.security.SecureRandom();
```

<details>
  <summary>More reading</summary>

  * [https://owasp.org/www-community/vulnerabilities/Insecure_Randomness](https://owasp.org/www-community/vulnerabilities/Insecure_Randomness)
  * [https://metebalci.com/blog/everything-about-javas-securerandom/](https://metebalci.com/blog/everything-about-javas-securerandom/)
  * [https://cwe.mitre.org/data/definitions/330.html](https://cwe.mitre.org/data/definitions/330.html)
</details> 

Powered by: [pixeebot](https://docs.pixee.ai/installing/) (codemod ID: [pixee:java/secure-random](https://docs.pixee.ai/codemods/java/pixee_java_secure-random))


jpstotz:
Thanks for your contribution. It is correct that numbers generated by `Random` are not cryptographically secure and thus may be guessable. The locations you have changed Random is used to generate random names for XML attributes that are obfuscated (e.g. contain invalid characters). Therefore predictability is not a problem at all for those generated names, they are just random strings aiming to improve human readability.

skylot:
> Therefore predictability is not a problem at all for those generated names, they are just random strings aiming to improve human readability

@jpstotz right, also I think these names should be completely deterministic, so Random not really needed here, and can be refactored to use just sequential numbers. I will try to prepare commit for such changes, but I can't find a sample to trigger this code and test my fix, if you know one, please share :slightly_smiling_face:

PR #2037, opened 2023-10-24
fix:fixed the issue of resourceIds reuse

markslin:
fixed the issue of resourceIds reuse


skylot (approved):

skylot:
@markslin thank you for the fix, looks good :+1:
And, if possible, please share a sample APK to reproduce this issue :slightly_smiling_face:

markslin:
Thanks for the recognition, but the app hasn't been released yet, so it's not convenient to provide

Issue #765, opened 2019-10-17
Deobfuscation does not know android.support.v4 and similar names


Sonic-Amiga:
I am deobfuscating an app containing android.support library, whose classes are obfuscated as a, b, etc. Automatic deobfuscator doesn't recognize android.support.vX as a valid name and renames vX to autogenerated unique name like C0001vX

jpstotz:
What name would you have expected? 
Note that Jadx does not contain functionality for library identification. Therefore Jadx is is not able to regenerate the original class names. The deobfuscation feature is mainly for classes that has been obfuscated in a way that makes them hard to read by humans.

skylot:
Duplicate of #66

Sonic-Amiga:
Well, this can be seen as a small subset of #66, but this one is easy to fix. Something like if(class_name.startsWith("android.support.v") ) deobfuscate=false;

Sonic-Amiga:
Just to be more clear about my case. "v4" part is preserved and not changed, so obfuscated names look like "android. support. v4.a", "android.support.v4.b", etc. And deobfuscator does not recognize "android.support.v" as legitimate, and replaces "v4" with generated name, which should not be done.

PR #2040, opened 2023-11-01
Implement a whitelist for deobfuscation

nitram84:
The deobfucator of jadx usually obfuscates packages or classes like

android.support.v4.*
android.support.v7.*
android.support.v4.os.*
android.support.annotation.Px
androidx.core.os.*
androidx.annotation.Px

Reproducible with any app (obfuscated or not) which uses support libraries.

Of course those names could be renamed again, but wouldn't it be easier with a configurable whitelist for deobfucation and reasonable defaults of well known classes and packages?

My default list may not be complete and if the list gets much longer in the future it could be better to switch to a file based whitelist.

This PR is a duplicate of https://github.com/skylot/jadx/issues/765 which was rejected, because jadx can't identify libraries and marked as duplicate of #66. In my opinion the opposite is now true: With this PR jadx would be able to identify more libraries correctly when deobfuscation is enabled: 

```kotlin
@file:DependsOn("org.xerial:sqlite-jdbc:3.43.2.1")
// LibCkecker.jadx.kts
/**
* For script usage check https://github.com/skylot/jadx/blob/master/jadx-plugins/jadx-script/README.md
*/

import org.sqlite.JDBC
import java.sql.Connection
import java.sql.ResultSet
import java.sql.Statement
import java.util.Properties
import java.util.regex.Pattern

data class Rule(val id: Int, val name: String, val label: String, val isRegex: Boolean, val regex: String?)

fun loadLibCheckerDatabase(): List<Rule> {
	val props: Properties = System.getProperties()
	val conn: Connection = JDBC.createConnection("jdbc:sqlite::resource:https://github.com/LibChecker/LibChecker-Rules/raw/master/cloud/rules/v3/rules.db", props)
	val stmt: Statement = conn.createStatement()
	val res: ResultSet = stmt.executeQuery("select _id, name, label, isRegexRule, regexName from rules_table where type = 5") // Only packages are supported for now, other resources not accessible via public jadx scripting api
	val rules: MutableList<Rule> = mutableListOf()
	while (res.next()) {
		val rule: Rule = Rule(res.getInt(1), res.getString(2), res.getString(3), res.getBoolean(4), res.getString(5))
		rules.add(rule)
	}
	res.close()
	stmt.close()
	conn.close()
	log.debug { "libChecker database loaded" }
	return rules
}

fun getPackages(): Set<String> {
	val packages: MutableSet<String> = mutableSetOf()
	jadx.classes.forEach {
		var idx = it.fullName.lastIndexOf('.')
		var packageName = it.fullName.substring(0, idx)
		packages.add(packageName)
		idx = packageName.lastIndexOf('.')
		while (idx > 0) {
			packageName = packageName.substring(0, idx)
			if (packages.contains(packageName)) {
				break
			}
			packages.add(packageName)
			idx = packageName.lastIndexOf('.')
		}
	}
	return packages
}

val jadx = getJadxInstance()

jadx.afterLoad {

	val packages: Set<String> = getPackages()
	val rules: List<Rule> = loadLibCheckerDatabase()
	rules.forEach {
		val currentRule: Rule = it
		var rulePattern: Pattern? = null
		if (currentRule.isRegex) {
			rulePattern = Pattern.compile(currentRule.regex)
		}
		packages.forEach {
			if ((currentRule.isRegex && rulePattern != null && rulePattern.matcher(it).matches()) || it.equals(currentRule.name)) {
				log.info { "Lib " + currentRule.label + " found, rule id: " + currentRule.id }
			}
		}
	}
}
```

And library detection might be a first step towards automatic dependency detection as a precondition of #66. 

Off topic question: Do you also accept also sample scripts as PR?


skylot:
> Off topic question: Do you also accept also sample scripts as PR?

Sure PRs are welcome

And LibChecker rules looks good, I definitly needs to check them, thanks :slightly_smiling_face:

skylot:
Ok, I merged PR and my refactoring commit.
I hope I didn't break anything, but it will be nice if someone checks the latest unstable build.

@nitram84 along with small fixes to your PR, I changed separator symbol to space for deobfuscation whitelist option, this way it is closer to similar options: `excluded packages` (now it is gui only) and slightly easier to read.
And I think we need to make a decent editor for such lists in jadx preferences :slightly_smiling_face:

nitram84:
You were right: having modular deobfucation conditions is/was definitively the better choice. Sorry for the work you had due to my PR. I'll  soon fix the translations with the new separator symbol and propose more whitelist entries.

Issue #1862, opened 2023-05-10
[feature][gui] Show list of proposed names for node rename 


skylot:
Currently, jadx can provide names for nodes from various sources like: source file, kotlin intrinsics or metadata, resources. But in result code, only one name will be used, and other names will be discarded. Also, result name can be incorrect or worse than others. So it will be nice to allow users to view and select a better name for each node, with a possible option to use a selected source for all other nodes.

Possible implementation details:
- add `proposeName(String name, String sourceId)` to [IRenameNode](https://github.com/skylot/jadx/blob/f52d2814dfbb23c414e746e7570a569635b7aed7/jadx-core/src/main/java/jadx/api/data/IRenameNode.java#L3)
- store list of <Name, SourceId> pairs in node attributes
- in gui show list of names in popup menu with options for each entry: 'apply' and 'use this source for all nodes'
- add light highlighting for nodes with several proposed names and/or show names list in tooltip
- [optional] sort names using ranking method from [BetterName](https://github.com/skylot/jadx/blob/f52d2814dfbb23c414e746e7570a569635b7aed7/jadx-core/src/main/java/jadx/core/utils/BetterName.java#LL19C28-L19C28)

Suggested by @NebelNidas in https://github.com/skylot/jadx/pull/1860#issuecomment-1542290948

PR #2041, opened 2023-11-10
Kotlin toString renamer produces illegal filenames and may override files during export

nitram84:
Steps to reproduce: 
* Open Jadx GUI (latest git), verify that the option [Plugins -> Kotlin Metadata -> rename fields using toString] is enabled
* Open https://github.com/niranjan94/show-java/releases/download/v3.0.5/ShowJava_v3_0_5.apk (and decompile Jadx itself)
* Navigate in structure tree to package "kotlin.jvm.internal": All class names in that package are valid
* Export sources as gradle project -> now there are two empty items in package "kotlin.jvm.internal"
* Show source code for both items: The class name is " " in both cases and an invalid file "app/src/main/java/kotlin/jvm/internal/ .java" was created.

This PR fixes this issue with a check for valid identifiers before rename. An working alternate fix would be using BetterName, since it includes the valid identifier check, but I wasn't sure if the BetterName metric for java is the same as for kotlin. If you prefer BetterName, please reject my PR and I'll fix this with BetterName.

This PR will fix lots of syntax errors in exported sources.

With other apps I've also seen class names like "." generating a filename "..java", but I haven't found any apps with path traversal attacks yet.


skylot (approved):

skylot:
@nitram84 thanks, nice fix!
Also, we may need to check this plugin for similar issues and maybe extract same code for node rename into a common method.

About BetterName: it is hard to detect which name is better so that check has limited usage, so in such cases it is better to allow user to select from names list as suggested in #1862 feature.

PR #2047, opened 2023-11-25
fix and add

LanBaiCode:
### Description
fix: Resolved an issue with incorrectly generated xposedMethodSnippet when the parameter type is generic.

error

fix

add: Introduced xposedGenerateFieldSnippet.

[Main.class.zip](https://github.com/skylot/jadx/files/13464744/Main.class.zip)


skylot (approved):

skylot:
@LanBaiCode looks good. Thank you!

Issue #1597, opened 2022-08-01
[feature] Load XAPKs directly


yoshimo:
APKPure has multi-arch apps in an XAPK file, which is more or less just a zip archive with multiple apks inside.
It would be nice if we could load them directly without unpacking them first and load the folder instead.

Iscle:
@skylot Hey, any plans on implementing this soon? I could do a PR myself if you're willing to review it and point out any changes you might see are necessary to meet the coding standard of the project.

I've noticed that lately more and more apps are only available as xapk from sites such as apkmirror and it's a pain in the ass to extract them before loading them.

skylot:
@Iscle sure, PR will be great :+1:

Iscle:
I've been working on this (I've decided to create a new plugin instead of modifying the existing jadx-dex-input) and I feel like especially the input plugins could maybe be improved a bit, with some sort of priority and maybe mutual exclusion so that if, for example, the xapk plugin supports the file, the dex plugin does not get executed.

I've also seen repeated functionality such as functions to read all bytes from an InputStream that maybe could be moved to a common project.

Is this something you've considered?

Iscle:
Furthermore, I've seen a few issues on how jadx handles resources and _AndroidManifest.xml_s. As you might know, xapk files are just a zip file with many apks inside and a few other support files such as a manifest.json and icon.png. 
Each of the split apk contains an AndroidManifest.xml, as well as some resources. If I extract the xapk and then load all the extracted files into jadx directly, the AndroidManifest.xml that will appear in the editor will be a random one.

It would be nice if the UI either showed all of them (maybe with a smaller gray text specifying which file it came from) or selected the "base" one (which we know as it's specified in the xapk manifest.json).

In my opinion, a refactor on the loading code logic and structure would be a great improvement to jadx. :)

skylot:
Now every plugin search for supported files from input and output decoded results.
To implement your suggestion we can run a global loop over all input files (with zip traversal) and asking every input plugin for decoding result (or even transform to other file format, so other plugins can decode it) and if file is supported stop asking other plugins. This will also allow plugin ordering. I was planing to implement similar thing to support resources decoding in plugins. Such changes require big refactoring with API changes and all plugins rewrite, so it is a lot of work.

As for xapk support, I thought it should be done in dex input plugin with zip unpacking like it already done for apk. This way it not require big refactoring and use will use already implemented approach.

For common parts we can just use `jadx-core` or `jadx-input-api` modules.

[...]

PR #2064, opened 2023-12-20
Add support for xapk files

iscle:
This PR adds support for xapk loading. It handles both: code and resources.

In order to load code, the xapk plugin makes use of the existing dex-input plugin instance by retrieving it from the plugin manager.

Some existing interfaces have been annotated with `@NotNull` tags for better Kotlin compatibility.

This fixes issue #1597

### DIFF
--- a/jadx-plugins/jadx-xapk-input/src/main/java/jadx/plugins/input/xapk/XapkInputPlugin.kt
+++ b/jadx-plugins/jadx-xapk-input/src/main/java/jadx/plugins/input/xapk/XapkInputPlugin.kt
@@ -11,15 +11,15 @@ class XapkInputPlugin : JadxPlugin {
 	internal var dexInputPlugin: DexInputPlugin? = null
 
 	override fun getPluginInfo() = JadxPluginInfo(
-			"xapk-input",
-			"XAPK Input",
-			"Load .xapk files"
+		"xapk-input",
+		"XAPK Input",
+		"Load .xapk files",
 	)
 
 	override fun init(context: JadxPluginContext) {
 		dexInputPlugin = context.decompiler.pluginManager.allPluginContexts
-				.map { it.plugin }
-				.firstOrNull { it is DexInputPlugin } as? DexInputPlugin
+			.map { it.plugin }
+			.firstOrNull { it is DexInputPlugin } as? DexInputPlugin
 
 		context.addCodeInput(codeInput)
 		context.decompiler.addCustomResourcesLoader(resourcesLoader)
--- a/jadx-plugins/jadx-xapk-input/src/main/java/jadx/plugins/input/xapk/XapkInputPlugin.kt
+++ b/jadx-plugins/jadx-xapk-input/src/main/java/jadx/plugins/input/xapk/XapkInputPlugin.kt
@@ -3,9 +3,12 @@ package jadx.plugins.input.xapk
 import jadx.api.plugins.JadxPlugin
 import jadx.api.plugins.JadxPluginContext
 import jadx.api.plugins.JadxPluginInfo
+import jadx.plugins.input.dex.DexInputPlugin
 
 class XapkInputPlugin : JadxPlugin {
+	private val codeInput = XapkCustomCodeInput(this)
 	private val resourcesLoader = XapkCustomResourcesLoader()
+	internal var dexInputPlugin: DexInputPlugin? = null
 
 	override fun getPluginInfo() = JadxPluginInfo(
 			"xapk-input",
@@ -14,6 +17,11 @@ class XapkInputPlugin : JadxPlugin {
 	)
 
 	override fun init(context: JadxPluginContext) {
+		dexInputPlugin = context.decompiler.pluginManager.allPluginContexts
+				.map { it.plugin }
+				.firstOrNull { it is DexInputPlugin } as? DexInputPlugin
+
+		context.addCodeInput(codeInput)
 		context.decompiler.addCustomResourcesLoader(resourcesLoader)
 	}
 }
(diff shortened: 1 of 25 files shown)


skylot (approved):

jadx-plugins/jadx-xapk-input/src/main/java/jadx/plugins/input/xapk/XapkInputPlugin.kt
```diff
@@ -8,7 +8,7 @@ import jadx.plugins.input.dex.DexInputPlugin
 class XapkInputPlugin : JadxPlugin {
 	private val codeInput = XapkCustomCodeInput(this)
 	private val resourcesLoader = XapkCustomResourcesLoader()
-	internal var dexInputPlugin: DexInputPlugin? = null
+	internal var dexInputPlugin = DexInputPlugin()
```
iscle, line 11: Why create a new instance of the DexInputPlugin if we can use the one already available?

skylot:
@iscle thanks!
Although, I made several changes:
- reformat code, check [this](https://github.com/skylot/jadx/wiki/Code-Formatting)
- remove NotNull annotation, it looks awful if all types have this annotation. I usually use only @Nullable.
- use inner apk name as a name of subdirectory instead of merging everything into root.
- make `CustomResourcesLoader` closeable to handle temp file deletion
- use `DexInputPlugin` directly with a new instance, it is used this way in other plugins

iscle:
> * remove NotNull annotation, it looks awful if all types have this annotation. I usually use only @nullable.

Yeah I agree, Kotlin will "accept" the parameter without an explicit null check if we skip the annotation, but it complains in some situations. However, in this case, it will work just fine!

> * make CustomResourcesLoader closeable to handle temp file deletion

Your solution looks good! I wanted to discuss this in the PR comment but I completely forgot.

jadx-plugins/jadx-xapk-input/src/main/java/jadx/plugins/input/xapk/XapkInputPlugin.kt
skylot: Well, until jadx have some API to do this search nicely, better not to rely on such hacky code :rofl:

skylot:
@iscle can you please quick check my changes, especially resources display :slightly_smiling_face:

iscle:
Sure, I'll check them a bit later. I think there there might be an issue with string resources getting overwritten maybe. (not with your changes, but in the OG implementation)

iscle:
Just checked, I see that now each APK is shown in the tree. Nice!

I only have one question, how will jadx handle the multiple AndroidManifest.xml when, for example, using the "Go to main Activity" button? Will it search through all of them until it finds the launch intent, or get the last one loaded?  
With the apk I tested the last apk is the "base" one but this is coincidental and might not be the case for all xapk files.

In the manifest, we do have an unused field which indicates this, which could be used to avoid searching all the manifests. However, I think it's better not to use it so that we can also handle the case where a user drags multiple .apk files into jadx from a split apk app.

skylot:
> how will jadx handle the multiple AndroidManifest.xml

Hm, now it uses first one found (code [here](https://github.com/skylot/jadx/blob/1e57e85382a230c2c1d070f8d0266ec8d63ac7c3/jadx-core/src/main/java/jadx/core/utils/android/AndroidManifestParser.java#L49-L48)), order same as zip entries. So this can be any other manifest. Not sure how fix this right now, because logic all over jadx didn't expect several manifests or resources.arsc files :sob:

iscle:
Well, sincr this is not part of xapk support I think the PR can be merged and this fixed in another PR.

nitanmarcel:
There are also .apkm files shipped by apkmirror and .apks which is the default format for Google, and the format which SAI (Split APK Installer) backups bundled applications to.

Anyway the unzipping idea is the same, and the naming I'd the apks too:

You have a base.apk which contains the base APK
You have one or multiple architecture apk which contains native libs for a specific architecture.
You have one or multiple DPI apk, containing drawables for one of the 4? available resolutions.
And lastly one or multiple locale apk, containing resources for different locales.


https://github.com/nitanmarcel/apkinjector/blob/c3b3ed2893a71863531d20dfdeaf5558b8129f22/apkinjector/bundle.py#L65

nitanmarcel:
> Just checked, I see that now each APK is shown in the tree. Nice!
> 
> I only have one question, how will jadx handle the multiple AndroidManifest.xml when, for example, using the "Go to main Activity" button? Will it search through all of them until it finds the launch intent, or get the last one loaded? With the apk I tested the last apk is the "base" one but this is coincidental and might not be the case for all xapk files.
> 
> In the xapk manifest.json we do have an unused field which indicates this, which could be used to avoid searching all the manifests. However, I think it's better not to use it so that we can also handle the case where a user drags multiple .apk files into jadx from a split apk app.

The main activity is only in the base.apk AndroidManifest.xml.

nitanmarcel:
And this is the format of a split APK file, not base.apk

```
<?xml version="1.0" encoding="utf-8"?>
<manifest
  xmlns:android="http://schemas.android.com/apk/res/android"
  android:versionCode="112"
  package="com.my.package"
  split="config.arm64_v8a"
  splitTypes="base__abi">
  <application
    android:hasCode="false"
    android:extractNativeLibs="false">
    <meta-data
      android:name="com.android.vending.derived.apk.id"
      android:value="3">
    </meta-data>
  </application>
</manifest>
```

iscle:
Hi @nitanmarcel,

Could you provide sample files for each case?

Thanks,
Iscle

PR #2070, opened 2023-12-28
feat: add an option to generate kotlin code for xposed

iscle:
This PR adds the ability to generate Xposed code in Kotlin instead of Java. 

This saves a lot of time when creating new Xposed modules with the help of jadx, since we can avoid having to manually convert the code or copy/paste the strings.

A new setting was added, bumping the settings version to 19.


skylot (approved):

skylot:
Looks good, thanks :+1:

PR #2074, opened 2023-12-31
feat: improve search bar behavior when using key bindings

iscle:
Before, the search bar would toggle when using the key binding to trigger it (Ctrl + f), which was quite annoying. Now, it replicates IntelliJ's search bar behaviour.

1.1. If the user has selected text, use that as the search text
1.2. Otherwise, use the previous search text (or empty if none)
2. Select all text in the search bar and give it focus

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/codearea/SearchBar.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/codearea/SearchBar.java
@@ -140,6 +140,24 @@ public class SearchBar extends JToolBar {
 		setVisible(false);
 	}
 
+	/*
+	 * Replicates IntelliJ's search bar behavior
+	 * 1.1. If the user has selected text, use that as the search text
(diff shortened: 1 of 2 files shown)


jadx-gui/src/main/java/jadx/gui/ui/codearea/SearchBar.java
```diff
@@ -149,8 +167,8 @@ public void toggle() {
 			if (!StringUtils.isEmpty(preferText)) {
 				searchField.setText(preferText);
 			}
-			searchField.requestFocus();
 			searchField.selectAll();
+			searchField.requestFocus();
```
iscle, line 171: This line was changed so that we don't request focus before having all the text selected (for fast keyboard users :P)

skylot (approved):

skylot:
@iscle thanks!

Issue #1440, opened 2022-04-07
[gui] When using the rename shortcut key N, there is a probability to enter it


AdminHcat:
When using the rename shortcut key N, there is a probability to enter it

There have been many times. When you press the N key, n appears instead of the original name

The following is the keyboard test software, indicating that the keyboard is pressed only once
![image](https://user-images.githubusercontent.com/50044729/162112321-1d972c1c-839a-400b-b015-d2e51a9a6eea.png)

skylot:
@AdminHcat thanks! I also notice this issue, and I agree, it is annoying.
I will try harder to fix this because my previous attempts not succeed, and I still don't understand the cause :slightly_frowning_face:

skylot:
Fixed in PR #2075

PR #2075, opened 2023-12-31
fix: attempt to improve field rename issues with "n" shortcut

iscle:
This is an attempt to fix the issues that sometimes arise when renaming functions or variables using the "n" shortcut as stated in issue #1440.

The reasoning behind the change: The instance creation of the RenameDialog was somehow affecting the UI thread and not allowing for the key release event to be dispatched. By running everything inside the invokeLater block, this might get fixed as it will execute after all previous tasks are finished. We now also only show the dialog after EVERYTHING is set up, not before.

Only time will tell if this is fixed, but I have not been able to trigger the issue after this change.


skylot (approved):

skylot:
@iscle indeed, using new update UI event should resolve issue. Thanks :+1:

PR #2082, opened 2024-01-14
Fixing removeUnreachable not searching in all blocks

Away-pp:
The method `removeUnreachableBlocks` was iterating through all the methods basic blocks but only keeping the last result found (in `toRemove`).

This will create an error later when the method `checkForUnreachableBlocks` is called.

Therefore, the fix is to use only one instance of the object `toRemove` and keep all the information there.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/blocks/BlockProcessor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/blocks/BlockProcessor.java
@@ -644,14 +644,13 @@ public class BlockProcessor extends AbstractVisitor {
 	}
 
 	private static void removeUnreachableBlocks(MethodNode mth) {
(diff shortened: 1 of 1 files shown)


skylot (approved):

skylot:
@Away-pp thanks! It is a very stupid mistake :sob:
Although, removing unreachable blocks is a bad idea in most cases because it is just hiding blocks processing issues.
I will try to review such approach and remove only real dead code on a very early stages before jadx do any changes to blocks tree.

Away-pp:
I understand that there are other issues with unreachable blocks, but I think that maintaining this function is ok, not a big issue.
This commit was the first part to fix a specific decompilation problem (will make the pull request later).

By the way, I don`t think it was a stupid mistake. It just happens.

Thank you for the quick review.

PR #2085, opened 2024-01-20
Fixing wrong order in removeUnreachableBlocks

Away-pp:
The counting of removed blocks and removed instructions is not in the right order in `removeUnreachableBlocks`.
`detachBlock` method will zero the instructions of a block before it is analyzed by the code and no warning of the method can be done.

The fix is to reverse the order

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/blocks/BlockProcessor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/blocks/BlockProcessor.java
@@ -654,13 +654,14 @@ public class BlockProcessor extends AbstractVisitor {
 			return;
 		}
 
-		toRemove.forEach(BlockSplitter::detachBlock);
-		mth.getBasicBlocks().removeAll(toRemove);
 		long notEmptyBlocks = toRemove.stream().filter(block -> !block.getInstructions().isEmpty()).count();
 		if (notEmptyBlocks != 0) {
 			int insnsCount = toRemove.stream().mapToInt(block -> block.getInstructions().size()).sum();
 			mth.addWarnComment("Unreachable blocks removed: " + notEmptyBlocks + ", instructions: " + insnsCount);
 		}
+
+		toRemove.forEach(BlockSplitter::detachBlock);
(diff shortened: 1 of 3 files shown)


skylot (approved):

Away-pp:
Test failed (sorry). I thought it was running when I tested locally, I confused the outcome. Please, do not merge.

skylot:
@Away-pp yeah, fix is correct thought :slightly_frowning_face:

Away-pp:
Let me check what is happening with the test that failed first just to be 100% sure that is not breaking something else.

skylot:
@Away-pp to fix the test, you can remove not needed return after throw here:
https://github.com/skylot/jadx/blob/aad2d24c589c49749410315d859412f6a8c62b16/jadx-core/src/test/smali/conditions/TestTernary4.smali#L137-L136

Also, `detachBlock` method not needed to clear instructions, so it is also better to remove that line.

And, after fix, don't create new PR, just update and reopen this one :slightly_smiling_face:

Away-pp:
So, this makes the test `TestDeadBlockReferencesStart` to fail.
Therefore, I think we need another approach.
I am not so familiar with JADX testing framework, so it may take some time.
But, anyway, I think the correct approach should be to change what exactly `TestTernary4` is checking.

skylot:
> So, this makes the test `TestDeadBlockReferencesStart` to fail.

Hm, it works fine for me, I pushed suggested changes to `pr-2085` branch ([result changes](https://github.com/skylot/jadx/compare/master...pr-2085)).

Away-pp:
Probably I was doing something wrong here (was working with two folders, probably mixed what I was doing).
Anyway, I will follow your changes.

Just to explain,
The `TestTernary4` can be fixed by adding `allowWarnInCode`. Everything seems ok.
I think the default is false, so, it will fail because of the warn produced.
```
	public void test() {
		allowWarnInCode();
		assertThat(getClassNodeFromSmali())
				.code()
				.removeBlockComments()
				.doesNotContain("5")
				.doesNotContain("try");
	}


```

skylot:
@Away-pp thanks :+1:

PR #2086, opened 2024-01-21
Removing unused ExcHandlers blocks

Away-pp:
Removing unused ExcHandlers blocks in order to fix some decompilation errors caused by Unreachable blocks.

The root of this problem (for the current fix) is that some `try`s may never actually throw an exception.
Therefore the related `catch` may become unreachable.

The solution is that at the end of the try/catch processing, to look for unreachable blocks from the exception handlers.

An example of method that was not decompiled before:

**Original code**

[https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/provider/FontsContract.java;l=726?q=prepareFontData](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/provider/FontsContract.java;l=726?q=prepareFontData)

**Decompiled code**

```
    private static Map<Uri, ByteBuffer> prepareFontData(Context context, FontInfo[] fonts, CancellationSignal cancellationSignal) {
        HashMap<Uri, ByteBuffer> out = new HashMap<>();
        ContentResolver resolver = context.getContentResolver();
        for (FontInfo font : fonts) {
            if (font.getResultCode() == 0) {
                Uri uri = font.getUri();
                if (!out.containsKey(uri)) {
                    ByteBuffer buffer = null;
                    try {
                        ParcelFileDescriptor pfd = resolver.openFileDescriptor(uri, "r", cancellationSignal);
                        if (pfd != null) {
                            try {
                                FileInputStream fis = new FileInputStream(pfd.getFileDescriptor());
                                try {
                                    FileChannel fileChannel = fis.getChannel();
                                    long size = fileChannel.size();
                                    buffer = fileChannel.map(FileChannel.MapMode.READ_ONLY, 0L, size);
                                    fis.close();
                                } finally {
                                    break;
                                }
                            } catch (IOException e) {
                            } catch (Throwable th) {
                                if (pfd == null) {
                                    throw th;
                                }
                                try {
                                    pfd.close();
                                    throw th;
                                } catch (Throwable th2) {
                                    th.addSuppressed(th2);
                                    throw th;
                                }
                            }
                        }
                        if (pfd != null) {
                            pfd.close();
                        }
                    } catch (IOException e2) {
                    }
                    out.put(uri, buffer);
                }
            }
        }
        return Collections.unmodifiableMap(out);
    }
```


jadx-core/src/main/java/jadx/core/dex/visitors/blocks/BlockExceptionHandler.java
```diff
@@ -65,6 +65,10 @@ public static boolean process(MethodNode mth) {
 			removeMonitorExitFromExcHandler(mth, eh);
 		}
 		BlockProcessor.removeMarkedBlocks(mth);
+
+		List<BlockNode> sorted = new ArrayList<>(mth.getBasicBlocks().size());
+		BlockUtils.dfsVisit(mth, sorted::add);
+		removeUnusedExcHandlers(mth, tryBlocks, sorted);
```
skylot, line 71: Looks like `sorted` blocks list later used only with `contains` so it is better to use `Set` instead `List`.
Or even better to use `BlockSet` from [here](https://github.com/skylot/jadx/blob/5aaceee978dba7935276672c26f0143e07072919/jadx-core/src/main/java/jadx/core/utils/blocks/BlockSet.java#L13)

jadx-core/src/main/java/jadx/core/dex/visitors/blocks/BlockProcessor.java
```diff
@@ -664,6 +664,26 @@ private static void removeUnreachableBlocks(MethodNode mth) {
...
+		}
+
+		long notEmptyBlocks = toRemove.stream().filter(block -> !block.getInstructions().isEmpty()).count();
+		if (notEmptyBlocks != 0) {
+			int insnsCount = toRemove.stream().mapToInt(block -> block.getInstructions().size()).sum();
+			mth.addWarnComment("Unreachable blocks removed: " + notEmptyBlocks + ", instructions: " + insnsCount);
+		}
+
+		toRemove.forEach(BlockSplitter::detachBlock);
+		mth.getBasicBlocks().removeAll(toRemove);
```
skylot, line 684: Instead code duplication, it is better to move this code into new method.

skylot (changes requested):
Looks good, just small requests to improve code.
Also, please add simple smali test (just copy whole method).

nitram84:
Here is a minimized standalone example for this issue that could be used for a unit test. Steps to reproduce: compile with openjdk8, "use dx/d8 to convert java bytecode" [x], "Plugins > Java convert > convert mode": "d8"

With convert mode "dx" there is still a NPE.

```java
import java.io.FileInputStream;
import java.io.IOException;
import java.nio.ByteBuffer;
import java.nio.channels.FileChannel;

public class UnusedExceptionHandlers1 implements AutoCloseable {
  public static void doSomething(final Object unused1, final Object[] array, final Object o1,
      final Object o2, final Object unused2) {
    for (final Object item : array) {
      ByteBuffer buffer = null;
      try (final UnusedExceptionHandlers1 u = doSomething2(o1, "", o2)) {
        try (final FileInputStream fis = new FileInputStream(u.getFilename())) {
          final FileChannel fileChannel = fis.getChannel();
          buffer = fileChannel.map(FileChannel.MapMode.READ_ONLY, 0, 42);
        } catch (final IOException e) {
          // ignore
        }
      } catch (final IOException e) {
        // ignore
      }
    }
  }

  private String getFilename() {
    return null;
  }

  private static UnusedExceptionHandlers1 doSomething2(final Object o1, final String s,
      final Object o2) {
    return null;
  }

  @Override
  public void close() throws IOException {}
}

```

jadx-core/src/main/java/jadx/core/dex/visitors/blocks/BlockExceptionHandler.java
Away-pp: Done

jadx-core/src/main/java/jadx/core/dex/visitors/blocks/BlockProcessor.java
Away-pp: Done

skylot (approved):

jadx-core/src/test/java/jadx/tests/integration/trycatch/TestUnreachableCatch2.java
```diff
@@ -0,0 +1,58 @@
...
+
+		@Override
+		public void close() throws IOException {
+		}
+	}
+
+	@Test
+	public void test() {
+		allowWarnInCode();
+		setFallback();
```
skylot, line 54:
```suggestion
		// TODO: result code not compilable because of 'break' after 'throw'
		disableCompilation();
```

skylot (approved):

skylot:
@Away-pp great! Thank you for your work :+1:

PR #2087, opened 2024-01-22
Disable XML pretty print

bagipro:

### Description
Hey! 

I created an option to disable printing XML attributes on new lines for CLI because it creates a lot of problems when parsing/displaying it automatically

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ options:
                                          'simple' - simplified instructions (linear, with goto's)
                                          'fallback' - raw instructions without modifications
   --show-bad-code                     - show inconsistent code (incorrectly decompiled)
+  --no-xml-pretty-print               - do not prettify XML
   --no-imports                        - disable use of imports, always write entire package name
   --no-debug-info                     - disable debug info parsing and processing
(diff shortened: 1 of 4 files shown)


skylot (approved):

skylot:
@bagipro ok, but what kind of issues can cause new lines in XML? I am just curious :slightly_smiling_face:.

bagipro:
@skylot 
When I want to display a specific XML element from the start line to the end line, new lines in attributes break the code and I have no idea how to fix that using the built-in Java XML SAX parser :)

skylot:
> to display a specific XML element from the start line to the end line

Oh, I see. I don't use SAX parser for a while :smile:.
Quick googling suggest using Locator with custom handler like [here](https://stackoverflow.com/questions/4915422/get-line-number-from-xml-node-java).

Anyway, thanks, maybe anyone else will find this option useful :slightly_smiling_face:

bagipro:
Yeah, that's pretty similar to what I do. However, it doesn't work properly when the beginning and the ending of the opener tag are on different lines

Thanks!

PR #2095, opened 2024-02-01
Update Build Status shield in README

sgallese:

### Description
Noticed README had an SVG that is no longer available.

Tried to find a Shields.io workflow that showed ALL build workflows, but looks like I can only pick one.

Another alternative is to simply upload the SVG again, but looking at old commits, I couldn't find any actual file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## JADX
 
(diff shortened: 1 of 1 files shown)


skylot (approved):

skylot:
@sgallese thanks :+1:

> Tried to find a Shields.io workflow that showed ALL build workflows, but looks like I can only pick one.

Only this one is fine.
It was broken after I rename/rearrange builds :cry:

PR #2111, opened 2024-02-27
Fixed XML proto parsing for removed debug data

bagipro:
Sometimes external tools remove all the debug data. In this case, attribute names, namespaces, and values are blank leading to XML parsing errors.


skylot (approved):

skylot:
@bagipro thanks :+1:

PR #2112, opened 2024-02-27
Avoid duplicated XML attrs

bagipro:
Hey!

I noticed that some very obfuscated apps contain duplicated XML attributes in resources. That's why parsing of such XML documents fails.

I didn't check how Android internally handles such cases, but the code in PR only dumps only the first entry

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java
@@ -286,13 +286,14 @@ public class BinaryXMLParser extends CommonBinaryParser {
 				}
 			}
 		}
+		Set<String> attrCache = new HashSet<>();
 		boolean attrNewLine = attributeCount != 1 && this.attrNewLine;
 		for (int i = 0; i < attributeCount; i++) {
-			parseAttribute(i, attrNewLine);
+			parseAttribute(i, attrNewLine, attrCache);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


skylot (approved):

skylot:
@bagipro looks good, thanks!
Although, it will be nice if you share link to app to test this issue :slightly_smiling_face:

bagipro:
Hey @skylot,

Do you want me to share it here? You can download the `com.ss.android.ugc.trill` app with the `33.4.1` version (https://apkcombo.com/tiktok-asia/com.ss.android.ugc.trill/download/phone-33.4.1-apk). If you open the `res/layout/q6.xml`. You will observe the following XML:
```xml
<p003X.C87522XaW layout_width="match_parent" layout_height="match_parent" fillViewport="true" layout_height="match_parent" layout_width="match parent">
```

skylot:
@bagipro thank you for the sample :+1:
It is really helpful sometimes to be able to check old samples for regressions, also I plan to start adding tests for resources decoder, maybe I will start with this one :rofl:

bagipro:
Great, but there's also another bug :) You can use the same sample to reproduce it because all attributes miss the `android:` namespace prefix. I've reported this bug to Apktool (in case they find a solution and they did https://github.com/iBotPeaches/Apktool/issues/2420), but I still didn't have enough of time to move it to Jadx :(

skylot:
Oh. Ok, I will check this. I will make a new issue for that issue.

PR #2121, opened 2024-03-11
Update AbstractCodeArea.java

zhongqingsong:
### Description
In general, we need data, not text in code. But now every time you copy the highlighted text, you copy the highlighted quotes as well. This often results in an extra need to delete the quotation marks around the sides, which is confusing. 
Now when copying selected highlighted text, quotes are not copied in.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/codearea/AbstractCodeArea.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/codearea/AbstractCodeArea.java
@@ -267,20 +267,20 @@ public abstract class AbstractCodeArea extends RSyntaxTextArea {
 
 	@Nullable
 	public String getWordUnderCaret() {
-		String tmpTxt = getWordByPosition(getCaretPosition());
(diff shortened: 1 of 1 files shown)


zhongqingsong:
Before this feature, in most scenarios, you wanted to use the copied data. Most of the operations must manually delete the quotation marks on both sides, which is troublesome to use.
For example, "test data from github"
1, if you need to search for this String in the search engine, you need to delete the quotation marks on both sides. The operation required is more cumbersome.
2, on the contrary, if you need double quotes, you must copy the content as a string in the code. Then the user only needs to enter double quotes, and the whole process is very smooth.
Compare the cost of using the two, and remove the copied highlighted text quotes, the benefit is greater.

The highlight copy function works even better. For example, after I copy the data, I can directly paste it into the browser to search, or I directly copy the values I need to the code, or I directly copy the results I get directly in Jadx search.
By removing the quotation marks, the convenience is greatly improved.

skylot (approved):

skylot:
@zhongqingsong thanks :+1:
I moved code to upper method to make copy consistent with highlighting, also this will handle null result safely now.

Issue #1232, opened 2021-08-19
[core]Error decode manifest


xmhwws:
apk link:https://drive.google.com/file/d/1kO75uhZ7hYJ08j-KF_HTYcZMaA40vCkW/view?usp=sharing

**Error decode manifest
java.io.IOException: Decode error: NAMESPACE header is not 0x0010, position: 0xdcbc
	at jadx.core.xmlgen.CommonBinaryParser.die(CommonBinaryParser.java:92)
	at jadx.core.xmlgen.BinaryXMLParser.parseNameSpace(BinaryXMLParser.java:151)
	at jadx.core.xmlgen.BinaryXMLParser.decode(BinaryXMLParser.java:111)
	at jadx.core.xmlgen.BinaryXMLParser.parse(BinaryXMLParser.java:79)**

![image](https://user-images.githubusercontent.com/42903751/130008739-5aa24b42-be7e-43e6-9984-00dadcac7310.png)

jpstotz:
apktool also fails to decode `AndroidManifest.xml` of this app (`java.io.IOException: Invalid chunk type (1310976).`). 

Are you sure the app is usable on an Android device?

xmhwws:
> Are you sure the app is usable on an Android device?

Yes, I'm sure this apk will work fine on Android phones.
This apk uses some kind of protective shell, which should be specially protected against decompiling tools, and that's why decompiling AndroidManifest.xml fails.

jackwpa:
FWIW aapt, aapt2, JEB decode this manifest without problem.

green9317:
I've been poking around and it appears that this manifest file has a chunk header size of 0x14. When reading the first namespace chunk (type 0x100).  Jadx currently searches for a size of 0x10 to ensure a valid header is found which is behavior I see in other apk parsers. 

I am having trouble finding a proper standard for the manifest but I think we could get jadx to read the file by removing the check.  Although, that might not be a great solution without knowing what the extra bytes are.

jpstotz:
I tested the APK in [FlowDroid](https://github.com/secure-software-engineering/FlowDroid) which uses [Sable axml](https://github.com/Sable/axml) library for parsing AndroidManifest.xml. This library seems to ignore the namespace chunk type, it only has an error later as the read namespace prefix id is invalid (way too large) so [namespace parsing](https://github.com/Sable/axml/blob/master/src/pxb/android/axml/AxmlReader.java#L84) fails.

[...]

PR #2122, opened 2024-03-11
fix some manifest decode Error.

CKCat:
1. The elementSize may be larger than the actual size of the element chunk.
2. End namespace chunk size can be any value. also fix issue https://github.com/skylot/jadx/issues/1232 .

[AndroidManifest.zip](https://github.com/skylot/jadx/files/14555279/AndroidManifest.zip)

 issue https://github.com/skylot/jadx/issues/1232  APK file.
[com.vmos.pro.apk.zip](https://github.com/skylot/jadx/files/14555312/com.vmos.pro.apk.zip)


skylot (approved):

skylot:
@CKCat thanks, looks good :+1:
Although, I am not sure if we correctly handle chunk data size in `parseNameSpaceEnd()` now, maybe we should make a more common code to skip trailing bytes, but is definitely for another commit/PR :slightly_smiling_face:

Issue #2184, opened 2024-05-17
[gui] File Chooser doesn't show .jadx.kts files


xxr0ss:
### Issue details

File dialog doesn't show `.jadx.kts` files under `jadx/jadx-plugins/jadx-script/examples/scripts` with the file filter
![image](https://github.com/skylot/jadx/assets/31790712/671e896c-ce3d-48e7-aae0-f59e79b18608)

### Jadx version

latest

### Java version

17

### OS

- [X] Windows
- [ ] Linux
- [ ] macOS

PR #2185, opened 2024-05-17
fix(gui): support filtering files with multiple extensions in file dialog

xxr0ss:
Fixes https://github.com/skylot/jadx/issues/2184

After this commit:
image

The origional filter uses the last segment split by `.` from file name to match files, so extensions like `.jadx.kts` won't be accepted

```java
public final class FileNameExtensionFilter extends FileFilter {
    // ...
    public boolean accept(File f) {
        if (f != null) {
            if (f.isDirectory()) {
                return true;
            }
            // NOTE: we tested implementations using Maps, binary search
            // on a sorted list and this implementation. All implementations
            // provided roughly the same speed, most likely because of
            // overhead associated with java.io.File. Therefor we've stuck
            // with the simple lightweight approach.
            String fileName = f.getName();
            int i = fileName.lastIndexOf('.');
            if (i > 0 && i < fileName.length() - 1) {
                String desiredExtension = fileName.substring(i+1).
                        toLowerCase(Locale.ENGLISH);
                for (String extension : lowerCaseExtensions) {
                    if (desiredExtension.equals(extension)) {
                        return true;
                    }
                }
            }
        }
        return false;
    }
    // ...
```

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/filedialog/CustomFileChooser.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/filedialog/CustomFileChooser.java
@@ -13,7 +13,6 @@ import javax.swing.JDialog;
 import javax.swing.JFileChooser;
 import javax.swing.JOptionPane;
 import javax.swing.UIManager;
-import javax.swing.filechooser.FileNameExtensionFilter;
 
 import jadx.api.plugins.utils.CommonFileUtils;
 import jadx.core.utils.Utils;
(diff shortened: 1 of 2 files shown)


skylot (approved):
Not sure why you use delegate of FileNameExtensionFilter, because it just uses provided description internally.
Everything else looks good, thanks :+1:

xxr0ss:
> Not sure why you use delegate of FileNameExtensionFilter, because it just uses provided description internally. Everything else looks good, thanks 👍

Hmm... the code I put in description is exactly part of **the** `FileNameExtensionFilter`, which doesn't work as expected.
Other parts of it just works fine, I don't want to accidentally break them. `FileNameExtensionFilter` is a `final class` so I have to delegate it instead of inheriting it.

thanks for merging : )

PR #2200, opened 2024-06-11
Allow to check for unstable releases

iscle:
### Description
This PR allows the Jadx updater to check for unstable releases. This will be useful for a later PR I'm working on which will allow for auto-updates, without the need to manually download the zip and extract it.

Many of us are using the unstable build as stable builds typically take a lot of time to be released and we always want the latest fixes and features (and bugs :P)

In a future PR which will be based in this code, I will add the option for jadx to auto-update just like other software do.

A settings option has been added to choose between the STABLE release channel (default, as always has been) or UNSTABLE (new behavior).

I chose to convert the code to kotlin since it looks a bit cleaner and has convenient features when handling Closeables and null checking.


iscle:
It looks like the checks have not passed due to an internal github error while processing the jobs @skylot

skylot (approved):

skylot:
@iscle looks good :+1:

> checks have not passed due to an internal github error

Manual re-run didn't help, looks like github issues not fixed yet. Will try to run again later.

PR #2209, opened 2024-06-27
fix (gui): Update russian translation

SnowVolf:


skylot:
@SnowVolf thanks :+1:

PR #2210, opened 2024-07-03
More lenient AXML parsing: allow for non-standard attributes sizes and avoid index exceptions when decoding some strings

qfalconer:
Almost all malicious APKs are now shipped with corrupted manifest. The corruption seep in where the official Android parse is more lenient, particularly the manifest present in these malware has one or more of the following:

- The strings count in the string chunk is almost always invalid. Jadx is immune to this because it correctly ignores it and decode the string on demand, given an index.
- The attributes size of the XML elements is greater than 20 (0x14). I've only seen 24 (0x18) being used, I don't know if Android extended the format (the extra DWORD seems to always be 0).
- Every XML element has an android:tag referencing an invalid string index.

In this patch the attributes size of an XML element is now accounted for.   
This size must be at least 20 (0x14) bytes but can be greater. Extra bytes are just skipped. 

When decoding a string, if such decoding is impossible a placeholder string is returned instead of throwing an exception. Not all code paths has been inspected, though (just what was necessary to make jadx parse the corrupted manifests at hand).

The attached [test-corrupted-manifest.zip](https://github.com/user-attachments/files/16086410/test-corrupted-manifest.zip) is an empty APK (or ZIP, same thing) with just a corrupted manifest from a real BRATA malware campaign. This can be useful to test this PR.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java
@@ -265,9 +265,10 @@ public class BinaryXMLParser extends CommonBinaryParser {
 			die("startNS's attributeStart is not 0x14");
 		}
 		int attributeSize = is.readInt16();
-		if (attributeSize != 0x14) {
-			die("startNS's attributeSize is not 0x14");
+		if (attributeSize < 0x14) {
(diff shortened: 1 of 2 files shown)


skylot:
@qfalconer nice fix! Thank you :+1:

Issue #2182, opened 2024-05-14
AndroidManifest.xml [Error decode manifest]


icyzs:
### Issue details

Using v1.4.7:
![image](https://github.com/skylot/jadx/assets/58960785/49f2ab0b-2116-4273-b70b-fca9dd5a4dc7)
Understand there is a commit on this issue: https://github.com/skylot/jadx/issues/1926

Using v1.5:
![image](https://github.com/skylot/jadx/assets/58960785/f0702c48-ed1c-4f0f-8362-7c1ac7eb6dbd)
However, i'm now getting this error without any further details. Not sure what's the problem here. 

### Jadx version

1.5.0

### Java version

17.0.2

### OS

- [X] Windows
- [X] Linux
- [X] macOS

icyzs:
I'm trying to analyze a malicious apk sample. It's probably part of their anti-analysis method. This is the apk i'm working on: https://malshare.com/sample.php?action=detail&hash=21e6f7e5112bd2e7d7a76b0c2603b0ae85d24eb4044964bc911198f06631b2ae

iznaka:
It looks like this app was protected with something called RizalProtect.

JEB Pro (latest version 5.12) decodes the manifest. Based on the warnings showing in the logger, it seems like this protector used many encoding hacks.

```<?xml version="1.0" encoding="UTF-8"?>
<manifest
  android:compileSdkVersion="23"
  android:compileSdkVersionCodename="6.0-2438415"
  android:tag=""
  android:versionCode="8035339"
  android:versionName="80.35.3.39"
  package="climb.order.tall"
  platformBuildVersionCode="30"
  platformBuildVersionName="11"
  xmlns:android="http://schemas.android.com/apk/res/android">
  <uses-sdk
    android:minSdkVersion="21"
    android:tag=""
    android:targetSdkVersion="29"/>
  <uses-permission
    android:name="android.permission.SEND_SMS"
    android:tag=""/>
  <uses-permission
    android:name="android.permission.SET_WALLPAPER"
    android:tag=""/>
  <uses-permission
    android:name="android.permission.READ_SMS"
    android:tag=""/>
  <uses-permission
    android:name="android.permission.READ_CALL_LOG"
    android:tag=""/>
  <uses-permission
    android:name="android.permission.READ_CONTACTS"
    android:tag=""/>
  <uses-permission
    android:name="android.permission.GET_ACCOUNTS"
    android:tag=""/>
  <uses-permission
[...]

PR #2211, opened 2024-07-06
More AXML/Manifest parsing improvements

jpstotz:
This PR fixes one of the problems that prevents Jadx from loading AndroidManifest.xml included in the sample APK of issue #2182.

Unfortunately I wasn't able to fix all problems so the APK still doesn't load. But may be the changes in this PR will be helpful for other obfuscated APK/AXML files.


skylot (approved):

skylot:
@jpstotz thanks :+1:

PR #2213, opened 2024-07-08
feat: Add button to go to Application class

### LINKED ISSUE #2208: [feature] Go to Application
### Describe your idea

I hope Jadx can add a navigation button for "Go to Application", such as "Go to Main Activity". The reason is that many important analysis cases need to be processed in the Application. If Jadx could have such a navigation button, it would be faster and more convenient to perform, while also improving the user experience. Thank you to the author and all the maintainers for their consideration!

iscle:
Icons from: https://intellij-icons.jetbrains.design/

Closes #2208


jadx-gui/src/main/resources/i18n/Messages_en_US.properties
```diff
@@ -15,6 +15,7 @@ menu.text_search=Text search
 menu.class_search=Class search
 menu.comment_search=Comment search
 menu.goto_main_activity=Go to main Activity
+menu.go_to_application=Go to Application
```
jpstotz, line 18: Having one label with `goto` and one with `go_to` isn't really nice:
```
menu.goto_main_activity=...
menu.go_to_application=...
```

iscle: I fixed it in a new commit. Renamed "goto" to "go_to" since it follows the pattern better.

skylot (approved):

skylot:
@iscle nice! Thanks :+1:

PR #2218, opened 2024-07-19
launch4j: use heap config from applicationDefaultJvmArgs

### LINKED ISSUE #2217: [feature] Is it possible to set the running RAM of jadx-gui at a fixed value
### Describe your idea

Hi,

When opening a jadx-gui, the running RAM will be select automatically, which is usually influenced by the current RAM left.
If I want the value to be higher(which is enough for the computer), I must close some other software before opening jadx-gui.

So, is it possible to set the running RAM of jadx-gui at a fixed value.

thanks.

jpstotz:
launch4j: use heap config from applicationDefaultJvmArgs - do not let launch4j overwrite it.

It seems like `maxHeapPercent` calculation uses the free RAM as basis for it's calculation. As we already have a working heap configuration `-Xms128M -XX:MaxRAMPercentage=70.0` in `applicationDefaultJvmArgs` I think it is better keep this and do not let launch4j reconfigure the heap size. This would also improve consistency between the launch scripts and launch4j exe launcher.

fixes #2217


skylot:
> It seems like maxHeapPercent calculation uses the free RAM as basis for it's calculation

For me, this looks like a good thing, because if we set max heap size bigger than system free memory and JVM will try to expand heap to this maximum value it will crash with segfault. Also, now jadx-gui monitor memory usage and take actions if it is close to max size (like reducing processing threads count and forcing GC run). So, be close to max heap size is better than crash. Anyway, free system memory can change over time, so such solution is also not ideal.

Another option is to try to monitor both java and system memory usage at the same time, so it will be possible to better handle low memory states. I will check if this possible and how to do this :slightly_smiling_face:

jpstotz:
> For me, this looks like a good thing, because if we set max heap size bigger than system free memory and JVM will try to expand heap to this maximum value it will crash with segfault. 

A Windows system crashing with segfault? Windows systems should always have page file that prevents crashes because of out of memory situations. System speed may go down because of frequent in/out paging operations so user can decide which process should be killed to restore normal operations.

skylot:
> A Windows system crashing with segfault?

JVM will crash, not system. And sure, page file will help.
Ok, assuming most systems have enabled page file/swap, it is safe to merge this PR.

Also, I found [oshi](https://github.com/oshi/oshi) library which can query system free/available memory, so it is possible to monitor system memory usage to improve jadx-gui operations on low memory.

skylot (approved):

PR #2231, opened 2024-07-31
feat: make detailed var info deterministic

pubiqq:
The result of the `SSAVar#getDetailedVarInfo` method can be used in generated code, so it's important to achieve that the method produces the same result for ssaVars with the same properties.


skylot (approved):

skylot:
@pubiqq thanks :+1:

PR #2234, opened 2024-08-01
fix(gui): Make code folding context menu actions appear 

Mino260806:
## The Issue

At first, I was going to mark this as a feature, only to find out this is already implemented. However, the current implementation is erroneous, and never adds the actions. Here's the problem

Code folding is disabled by default. Its value is set using `AbstractCodeArea.setCodeFoldingEnabled()`.
However, AbstractCodeArea prebuilds the context menu in its constructor, fired in this line.
https://github.com/skylot/jadx/blob/61855a7ea1d936c36a894c9f3db9a81e165b54d7/jadx-gui/src/main/java/jadx/gui/ui/codearea/AbstractCodeArea.java#L93-L93

But superclass constructor is called before child constuctor, so AbstractCodeArea calls `isCodeFoldingEnabled()` before child gets to change it, so `isCodeFoldingEnabled()` below always returns its default value of `false` and code folding actions are never added.
https://github.com/skylot/jadx/blob/61855a7ea1d936c36a894c9f3db9a81e165b54d7/jadx-gui/src/main/java/jadx/gui/ui/codearea/AbstractCodeArea.java#L139-L145

## My Solution
I thought the best way to add it is to add the folding menu actions once `setCodeFoldingEnabled(true)` is called. However `RSyntaxTextArea.appendFoldingMenu(JPopupMenu)` only adds the actions to the end, so I made a workaround that removes present actions after target index, adds folding menu, and then adds present actions back.

https://github.com/Mino260806/jadx/blob/9daac915ef08a048cb39581335839229ac9ecf93/jadx-gui/src/main/java/jadx/gui/ui/codearea/AbstractCodeArea.java#L142-L172

I'll open a PR in RSyntaxTextArea that creates `RSyntaxTextArea.appendFoldingMenu(JPopupMenu, int)`, and we would fix this workaround once merged and released.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/codearea/AbstractCodeArea.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/codearea/AbstractCodeArea.java
@@ -12,6 +12,7 @@ import java.awt.event.KeyAdapter;
 import java.awt.event.KeyEvent;
 import java.awt.event.MouseListener;
 import java.awt.event.MouseMotionListener;
+import java.util.ArrayList;
 import java.util.Objects;
 
 import javax.swing.AbstractAction;
@@ -77,6 +78,7 @@ public abstract class AbstractCodeArea extends RSyntaxTextArea {
 	protected JNode node;
 
 	protected volatile boolean loaded = false;
+	private int foldingMenuIndex;
 
 	public AbstractCodeArea(ContentPanel contentPanel, JNode node) {
 		this.contentPanel = contentPanel;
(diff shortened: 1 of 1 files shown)


skylot:
Hm, I think we can run `addWrapLineMenuAction` inside `createPopupMenu()`.
And all code for adding popup menu items should be done in `createPopupMenu()`.
I will check this and prepare a fix.

skylot:
I commit a fix in `fix-folding-menu` branch, commit 8e00de39043e6c20a127a86ee0583c4f37726f73
It works fine for me. The only downside is that using this inheritance/override way it is harder to control order of items, but we still can insert something in the middle if needed. Anyway, current order looks fine to me.
@Mino260806 please check my commit, and you can use it or fix/adapt if you want :smile:

Mino260806:
@skylot it's fine you can merge your branch, I was also hesitant about my approach

PR #2235, opened 2024-08-01
fix: improve checking if methods are inline (for FixAccessModifiers)

pubiqq:


skylot (approved):

skylot:
@pubiqq nice fix! Thank you :+1:

PR #2241, opened 2024-08-05
feat(gui): Quick Tabs Overhaul

### LINKED ISSUE #1078: [feature] Add the ability to bookmark a node for later reference
It would be useful to add a feature to the GUI enabling a user to bookmark nodes using the right click menu.  These nodes could then be cached and ideally a symbol or highlighting is used to indicate which nodes / classes are bookmarked for easy reference.  An option would then be added to the navigation menu to enable the user to jump to any of the bookmarked nodes.  This information would need to be stored in the jadx or jobf file for cases when an app is being looked at over multiple days.

### LINKED ISSUE #1448: [feature] Bookmarks and search history
*Describe your idea:*
Bookmark function: when there are many files, it can quickly mark and locate to the specified place for secondary opening and analysis.
Search history: it is convenient to quickly locate the searched location during the initial analysis of the project.

The above functions hope to support a separate pop-up window display and improve efficiency.
Jadx is great, thanks to the author

Mino260806:
- [x] (internal) Decentralize Tabs Controlling code to a seperate TabsController class. (*1)
- [x] Add Open Tabs parent node in Quick Tabs that allows users to view tabs that are open
- [x] Add Bookmarked Tabs parent node in Quick Tabs that allows users to bookmark tabs and keep a reference even when the tab is closed

~~_Unchecked tasks are WIP_~~

------
(*1) This change is a must. Previously, `TabbedPane` was the only class interacting with tabs. So it was fine that all the code that controls tabs was found there. However now, that we are adding `QuickTabsTree`, (and why not other components), it no longer makes sense that `QuickTabsTree` needs to communicate with `TabbedPane` in order to manage the tabs. Counter example ? There may be tabs which are not open in `TabbedPane` but are present in `QuickTabsTree`, like bookmarked tabs. So if we would keep using old code, we'd have to store in `TabbedPane` a seperate reference to tabs that are present but not open, which would transform the code into a mess. I hope you understood the point.

Now TabsController hold a "generic reference" to each tab, and TabbedPane holds a "component reference".

<details>

<summary>Before</summary>

image

</details>

<details>

<summary>Now</summary>

image

</details>

------

### Affected issues

- closes #1448
- closes #1078

_Icons from https://intellij-icons.jetbrains.design_


jadx-gui/src/main/java/jadx/gui/ui/tab/TabComponent.java
```diff
@@ -56,14 +59,15 @@ public void loadSettings() {
...
+		return tabsController.getMainWindow().getSettings().getFont().deriveFont(Font.BOLD);
 	}
 
 	private void init() {
 		setLayout(new FlowLayout(FlowLayout.LEFT, 0, 0));
 		setOpaque(false);
 
-		JNode node = contentPanel.getNode();
+		JNode node = getNode();
+		TabBlueprint blueprint = getBlueprint();
```
skylot, line 70: `blueprint` not used later in code.

Also, looks like we don't have any visual mark that tab was bookmarked. I think we can add a little bookmark icon in top left corner of tab icon. Please check `jadx.gui.utils.OverlayIcon` and how it used to add small icons for fields.

skylot (changes requested):
Looks good, but I found some issues and request small improvements :slightly_smiling_face:

skylot:
One more thing.
I notice that on reload (or close jadx-gui and open project again) tabs order changes.
To reproduce, move any tab to other position and press reload action (from toolbar or F5) and tabs will reorder.
Not sure, maybe that was broken before this commit, but as soon as we change saving code here it is better to use actual tabs order on project save.

Mino260806:
@skylot all ready

skylot:
Hm, not sure why we need to save index to restore order, is just saving/reading tabs list in correct order not working?

Mino260806:
> Hm, not sure why we need to save index to restore order, is just saving/reading tabs list in correct order not working?

`TabsController` now holds the generic reference to each `TabBlueprint` through a `Map<JNode, TabBlueprint>`, I avoided saving the tabs in the TabbedPane order because
-  it would require a callback to TabbedPane before saving tabs that reorders TabBlueprint and saves them back, and we would need filtering for tabs that are hidden
-  we would anyway use the same procedure (store index) if we wanted to also preserve order for QuickTabs and allow reordering later on

But maybe we can just consider the order of TabbedPane to be the "default" order, ok I'll implement it by saving/reading tabs list in correct order

Mino260806:
@skylot please see commit https://github.com/skylot/jadx/pull/2241/commits/6ca79d48de6665fe32f0e40f905713406ade9940. I implemented saving/reading tabs list in correct order

jadx-gui/src/main/java/jadx/gui/ui/tab/TabComponent.java
```diff
@@ -141,13 +148,26 @@ public void updateCloseOrPinButton() {
...
-		contentPanel.setPinned(!contentPanel.isPinned());
-		updateCloseOrPinButton();
-		if (contentPanel.isPinned()) {
-			tabbedPane.advanceTab(this);
+	public void updateBookmarkIcon() {
+		icon.clear();
+
+		if (getBlueprint().isBookmarked()) {
+			icon.add(Icons.BOOKMARK_OVERLAY_DARK);
+		}
```
skylot, line 156:
```suggestion
		}
		label.repaint();
```

skylot (approved):

skylot:
OK, I think we can merge it now.
Although, here is a small list of thing for future improvements:
- There are still a lot of calls using TabbedPane instead TabsController. Like in MainWindow we call method in TabbedPane next it calls TabsController, and it calls TabbedPane again :rofl:. This is confusing and not follow architecture you draw :sob:
- It feels like `onTabsReorder` method implementation has very high complexity (like 2*n^2) and can take a lot of time for lots of tabs and bookmarks, so it is better to optimize it. And in general feels like that we are very often iterating over lists in search of one element, maybe some additional map will help.
- Please do not use `ArrayList` as a variable type, it is considered a bad practice in Java. I think checkstyle has rule for that ([this](https://checkstyle.sourceforge.io/checks/coding/illegaltype.html)), I will enable it and fix all such places.

Mino260806:
> Although, here is a small list of thing for future improvements

All noted. I'll open a new PR in the near future

PR #2242, opened 2024-08-07
fix(gui): Quick Tabs Optimization

Mino260806:
Follow-up to PR #2241
- [x] Optimize tabs reorder
- [x] Redirect all calls that open tabs to TabsController

Should we also move `JumpManager`, `navBack` and `navForward` to TabsController ? I'm hesitating because it's something particular to `TabbedPane`


skylot:
> Should we also move JumpManager, navBack and navForward to TabsController ?

I think we can make it independent: new class (JumpsController?) to listen tabs changes and code jumps with ITabsStatesListener, record these jumps into JumpManager, and on back/forward actions calls TabsController methods. One downside here is that we need to use `currentPosition()` from TabbedPane to save jump origin, but current position is tied to editor actually not to tabs, so it is already weird to have such method in TabbedPane :rofl:  So maybe we can move this method to new JumpsController :thinking:

Mino260806:
(added NavigationController in last commit, sorry for wrong commit message)

Anything else that needs refactoring ?

skylot:
> Anything else that needs refactoring ?

No, I think all good now.
I will review and test current changes (maybe tomorrow, actually) and we can merge this PR.

skylot:
> I think all good now

And next I trigger some error without exception and spend lots of time debugging it :rofl:
Exception was lost in the threads of background executor, adding more try/catch helps to reveal it (fix in [de26dc9](https://github.com/skylot/jadx/pull/2242/commits/de26dc9a93975806550d41f669fa2877b67c4659)).
Anyway, I made some changes:
- use of default methods in `ITabStatesListener` helps to reduce empty methods in implementations
- make NavigationController listen tabs events, so it is really independent now
- move code of "sync with editor" feature into separate class
- add LogTabStates class to debug tab states events (helps understand order of similar action like select, open, codejump)
- another attempt to impove onTabsReorder
- resolve some NPEs
- resolve merge conflict

skylot (approved):

skylot:
@Mino260806 please check my changes, I think we can merge this PR :slightly_smiling_face:

skylot:
@Mino260806 I will merge this PR. It contains a lot of good changes and fixes.

Mino260806:
Sorry I had no time to check it

PR #2243, opened 2024-08-07
fix: fix inlining synthetic accessors

pubiqq:


skylot:
@pubiqq can you share test case for this issue (smali or apk sample with method names)?
I will prepare a regression test.

skylot:
@pubiqq thanks :+1:
I commit several changes:
- add test (only for 4th case)
- undo adding new field into InsnNode: we don't really needed to enforce args list immutability, also I don't want to change method return type, because it is part of public API and can be used in external plugins/scripts. So I just add a new method to get args list :rofl:
- wrap inline checks code into try/catch to not fail method decompilation, just skip method inline.

skylot (approved):

PR #2252, opened 2024-08-15
fix: improve checking of access modifiers for methods

pubiqq:


jadx-core/src/main/java/jadx/core/dex/nodes/utils/ClassUtils.java
```diff
@@ -14,44 +12,7 @@ public ClassUtils(RootNode rootNode) {
...
-	}
-
-	private boolean isPrivateAccessible(ClassNode cls, ClassNode callerCls) {
-		return cls.getTopParentClass().equals(callerCls.getTopParentClass());
-	}
-
-	private boolean isSuperType(ClassNode cls, ClassNode superCls) {
-		return root.getClsp().getSuperTypes(cls.getRawName()).stream()
-				.anyMatch(x -> x.equals(superCls.getRawName()));
+		CodeNodeUtils codeNodeUtils = new CodeNodeUtils(root);
```
skylot, line 15: I don't see a reason to create this instance every time. Please store it in final field (same for method utils)
Also, maybe name it something like `AccessChecks` or `AccessCheckUtils` it is big enough and more specific name is better.

jadx-core/src/main/java/jadx/core/dex/nodes/utils/CodeNodeUtils.java
```diff
@@ -0,0 +1,85 @@
...
+
+	private final RootNode root;
+
+	CodeNodeUtils(RootNode rootNode) {
+		this.root = rootNode;
+	}
+
+	boolean isAccessible(ICodeNode targetNode, ICodeNode callerNode) {
+		ClassNode targetCls = getDeclaringClass(targetNode);
+		ClassNode callerCls = getDeclaringClass(callerNode);
```
skylot, line 21: I notice that we can have inner class in another maybe private inner class, so it requires to check all upper level visibility and use more strict one as an result/effective one. This will make code more complex, but it will be nice to check.
Anyway, at this point we might need a test case for this method (just a unit test) :slightly_smiling_face:

skylot (changes requested):

jadx-core/src/main/java/jadx/core/dex/nodes/utils/ClassUtils.java
pubiqq: Done

jadx-core/src/main/java/jadx/core/dex/nodes/utils/CodeNodeUtils.java
pubiqq: I'm not quite sure what you mean. Members of inner classes are accessible from everywhere inside the top class.

<details><summary>Sample</summary>
<p>

```java
public class TestCls {

  private class A {

    private class AA {

      private void testFromAA() {
        testFromA();

        B b = new B();
        b.testFromB();

        B.BB bb = (new B()).new BB();
        bb.testFromBB();
      }
    }

    private void testFromA() {
      AA aa = new AA();
      aa.testFromAA();

      B b = new B();
      b.testFromB();

      B.BB bb = (new B()).new BB();
      bb.testFromBB();
    }
  }

  private class B {

    private class BB {

      private void testFromBB() {
        testFromB();

        A a = new A();
        a.testFromA();

        A.AA aa = (new A()).new AA();
        aa.testFromAA();
      }
    }

    private void testFromB() {
      BB bb = new BB();
      bb.testFromBB();

      A a = new A();
      a.testFromA();

      A.AA aa = (new A()).new AA();
      aa.testFromAA();
    }
  }

  private void test() {
    A a = new A();
    a.testFromA();

    B b = new B();
    b.testFromB();

    A.AA aa = (new A()).new AA();
    aa.testFromAA();

    B.BB bb = (new B()).new BB();
    bb.testFromBB();
  }
}
```

</p>
</details>

skylot:
```java

public class A {
	public void call() {
		B.BB.BBB.test();
	}
}

public class B {
	private static class BB {
		public static class BBB {
			public static void test() {
			}
		}
	}
}
```
Here A and B different top classes and while BBB is public you can't access it with `B.BB.BBB.test();` because `B.BB has private access in B` (java compiler said this).

But sure, such cases looks exotic, so we might be better to fix these in separate PR.

skylot (approved):

jadx-core/src/main/java/jadx/core/dex/nodes/utils/CodeNodeUtils.java
pubiqq: Ok, I'll think about that.

Issue #1913, opened 2023-06-14
[core] Jadx deobfuscates valid identifiers


bagipro:
Hey!

The latest version deobfuscates valid identifiers in apps that are not obfuscated. This behavior wasn't present before. Sample app: https://drive.google.com/file/d/1NEIPWFwukzv1BgJDHpaqKZAznMf25lqW/view?usp=sharing

When you open the `kotlin` package, you will see the following contents:
<img width="489" alt="image" src="https://github.com/skylot/jadx/assets/15173437/dbb3b967-5b36-40e4-8130-fb65ff6dbc06">

However, I believe, those classes shouldn't be deobfuscated

bagipro:
@skylot 
I have the following configuration:
```json
{
  "flattenPackage": false,
  "checkForUpdates": false,
  "fontStr": "",
  "smaliFontStr": "",
  "editorThemePath": "/org/fife/ui/rsyntaxtextarea/themes/default.xml",
  "lafTheme": "FlatLaf Light",
  "langLocale": {
    "locale": "en_US"
  },
  "autoStartJobs": false,
  "excludedPackages": "",
  "autoSaveProject": false,
  "showHeapUsageBar": false,
  "alwaysSelectOpened": false,
  "useAlternativeFileDialog": false,
  "codeAreaLineWrap": true,
  "srhResourceSkipSize": 3,
  "srhResourceFileExt": ".xml|.html|.js|.json|.txt",
  "searchResultsPerPage": 50,
  "useAutoSearch": true,
  "keepCommonDialogOpen": true,
  "smaliAreaShowBytecode": false,
  "lineNumbersMode": "AUTO",
  "mainWindowVerticalSplitterLoc": 300,
  "debuggerStackFrameSplitterLoc": 300,
  "debuggerVarTreeSplitterLoc": 700,
  "adbDialogPath": "",
  "adbDialogHost": "localhost",
  "adbDialogPort": "5037",
  "codeCacheMode": "MEMORY",
  "usageCacheMode": "NONE",
  "jumpOnDoubleClick": true,
  "treeWidth": 601,
  "dockLogViewer": true,
  "settingsVersion": 18,
  "skipResources": false,
  "skipSources": false,
  "exportAsGradleProject": false,
  "threadsCount": 8,
  "decompilationMode": "AUTO",
  "showInconsistentCode": true,
  "useImports": false,
  "debugInfo": true,
  "addDebugLines": false,
  "inlineAnonymousClasses": false,
[...]

PR #2260, opened 2024-08-29
feat: don't try to guess a better class name when "use source file name" is on

pubiqq:
This PR makes the app ___use source file name as class name alias___ if the "___Use source file name as class name alias___" checkbox is checked, without implicit attempts to guess a better name.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/rename/SourceFileRename.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/rename/SourceFileRename.java
@@ -15,7 +15,6 @@ import jadx.core.dex.attributes.nodes.RenameReasonAttr;
 import jadx.core.dex.nodes.ClassNode;
 import jadx.core.dex.nodes.RootNode;
 import jadx.core.dex.visitors.AbstractVisitor;
-import jadx.core.utils.BetterName;
 import jadx.core.utils.StringUtils;
 import jadx.core.utils.exceptions.JadxException;
 
@@ -61,14 +60,6 @@ public class SourceFileRename extends AbstractVisitor {
 	}
 
 	private static void applyRename(ClassNode cls, String alias) {
-		if (cls.getClassInfo().hasAlias()) {
-			// ignore source name if current alias is "better"
-			String currentAlias = cls.getAlias();
-			String betterName = BetterName.compareAndGet(alias, currentAlias);
-			if (betterName.equals(currentAlias)) {
-				return;
-			}
-		}
 		cls.getClassInfo().changeShortName(alias);
 		cls.addAttr(new RenameReasonAttr(cls).append("use source file name"));
 	}
--- a/jadx-core/src/test/java/jadx/tests/integration/deobf/TestBadSourceFile.java
+++ b/jadx-core/src/test/java/jadx/tests/integration/deobf/TestBadSourceFile.java
@@ -34,7 +34,8 @@ public class TestBadSourceFile extends SmaliTest {
 		args.setDeobfuscationMinLength(100); // rename everything
 		assertThat(searchCls(loadFromSmaliFiles(), "b"))
 				.code()
-				.containsOne("class C0000b {")
-				.containsOne("/* compiled from: a.java */");
+				.containsOne("class a {")
+				.containsOne("/* compiled from: a.java */")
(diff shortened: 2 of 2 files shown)


skylot:
Hm, source file name in most cases contains garbage, so I prefer to keep this renames as safe as possible.
Also in jadx-gui this option is global and enabling it only for some projects can be annoying.
Possible solutions:
1. make this option local or "per project" - this way it can be enabled only if needed in the current project.
  Big downside here is that this requires a lot of changes in jadx-gui settings UI to allow to change and distinguish global and local parameters. I plan to implement this someday along with custom and predefined profiles support.
1. change this option into "enum" with entries like `Always`, `If better`, `Never`.
  Small downside is to compatibility/migration issue, so it is actually easier to create new option `apply source file name` and deprecate old one (internally switching to use only new one).
1. improve `better name` algorithm.
  I don't know what case you encounter, but it will be nice if you share it (in format `original` vs `source` names) :slightly_smiling_face:

I like the second solution, it is just easier. And maybe we can use `If better` option as default (disabled now).

Related to #1913.

pubiqq:
Well, the checkbox is called "Use source file name as class name alias", I explicitly checked it, and I expect it to work that way. The checkbox is unchecked by default, so the PR won't change the default behavior of the program and class names will remain the same as before.

Your proposal can be implemented, but I think that's material for a separate issue/discussion/PR.

> I don't know what case you encounter, but it will be nice if you share it (in format `original` vs `source` names) 🙂

Sure, here are some examples:
```
currentAlias: "C2937kh",         alias: "Room"         => betterName: "C2937kh", but should be "Room"
currentAlias: "C1428qi",         alias: "Fade"         => betterName: "C1428qi", but should be "Fade"
currentAlias: "C4063yi",         alias: "Scene"        => betterName: "C4063yi", but should be "Scene"
currentAlias: "AbstractC3038lk", alias: "PagerAdapter" => betterName: "AbstractC3038lk", but should be "PagerAdapter"
```

skylot:
> I explicitly checked it, and I expect it to work that way

That is why I suggested to use more explicit options/names instead of just boolean, it will be easier to understand what it does.

> The checkbox is unchecked by default, so the PR won't change the default behavior of the program and class names will remain the same as before.

This option works only for some apks/projects, enabling it will affect all projects and might apply worse names and this is the reason why it disabled by default.

> here are some examples

Oh, so we just need to not select names generated by jadx or names with many numbers.
This is actually will be the easiest fix for this issue :rofl:

So, as a conclusion to this PR: I don't want to remove `better name` check. We need to apply suggested solutions, and it looks like we will need all of them eventually.
@pubiqq you can change this PR into one of solutions or close it and open a new one if you want to.

And BTW, another solutions is to provide API to change `better name` implementation and allow users to implement own rules via scripts or plugins (similar as it was done for deobfuscator).

pubiqq:
It looks like you've thought of something more global than what I had in mind in this PR. 😂

> That is why I suggested to use more explicit options/names instead of just boolean, it will be easier to understand what it does.
>
> This option works only for some apks/projects, enabling it will affect all projects and might apply worse names and this is the reason why it disabled by default.

I have no problem understanding the option name.
I'm not suggesting changing the default renaming behavior.

I just want the checkbox with the name 
"___Use source file name as class name alias___" 
to make the app 
___use source file name as class name alias___, 
without hidden mechanics and surprises.

I have other changes to improve the renaming mechanism as well, including the ones you suggested
<details><summary>Trust me</summary>
<p>

</p>
</details> 

, I'm just going to submit them in separate PRs.

--

> And BTW, another solutions is to provide API to change `better name` implementation and allow users to implement own rules via scripts or plugins (similar as it was done for deobfuscator).

It would be great if deobfuscation/renaming could be implemented as a separate plugin (if possible of course).

skylot:
> Trust me, I'm just going to submit them in separate PRs.

Sure, if you have better implementation and want to open a new PR, I will close this one.

> It would be great if deobfuscation/renaming could be implemented as a separate plugin (if possible of course).

It is possible, all renaming happen in jadx passes, so it can be done in a jadx plugin ([short guide](https://github.com/skylot/jadx/wiki/Jadx-plugins-guide))

pubiqq:
Ehmm... you literally ignored... ok, nvm...

Issue #2267, opened 2024-09-06
[core] 16-bit resource entry offsets are not supported


pubiqq:
### Issue details

https://cs.android.com/android/platform/superproject/+/android15-release:frameworks/base/libs/androidfw/include/androidfw/ResourceTypes.h;l=1471-1473

### Jadx version

1.5.0

skylot:
@pubiqq can you share a sample with such flag?

pubiqq:
> can you share a sample with such flag?

android.jar from [Android SDK Platform 35](https://dl.google.com/android/repository/platform-35_r01.zip).

PR #2269, opened 2024-09-06
feat(res): support 16-bit entry offsets

pubiqq:
Fixes https://github.com/skylot/jadx/issues/2267

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ResTableBinaryParser.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ResTableBinaryParser.java
@@ -279,10 +279,6 @@ public class ResTableBinaryParser extends CommonBinaryParser implements IResTabl
 		boolean isSparse = (flags & FLAG_SPARSE) != 0;
 		boolean isOffset16 = (flags & FLAG_OFFSET16) != 0;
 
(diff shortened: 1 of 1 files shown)


skylot:
Hm, with sample provided in https://github.com/skylot/jadx/issues/2267#issuecomment-2334555118 parsing fail at closest
```java
is.checkPos(entriesStart, "Expected first entry start");
```
because of additional 2 bytes, maybe it is padding, so it might be better to change this line to 
```java
is.skipToPos(entriesStart, "Failed to skip to entries start");
```
And, as soon as, sample also use compact entries, it still can't be parsed, so I am not sure if this is a correct solution.
To be honest, it is better to merge this PR with compact entries support, because this is the only way to verify its correctness.

pubiqq:
This PR only adds support for 16-bit entry offsets. Not support for compact entries, not support for non-standard paddings, only 16-bit entry offsets. 

But if you insist on implementing all together, then you can close the PR. I don't think I'll create a PR with compact entries support, because my current implementation of the resource parser is too different from yours.

skylot:
I see.
I will try to implement compact entries support myself and will merge this PR once I managed to parse provided sample.

skylot (approved):

Issue #818, opened 2019-12-26
[feature]Some suggestions to improve jadx


xmhwws:
I have used several decompilation software and found that some aspects of jadx can be improved more:
1. Sometimes users only need to search for `"included strings"` instead of searching for `codes`. I hope jadx can add this function;

2. When I searched, there were multiple pages of search results. I clicked to turn the page and turned to one of them. When I click on search again, I want jadx to retain the previous search status, including page turning status, instead of re-searching, and then positioning to the first page;

3. When I search for the method name, if the class name includes this string, jadx will also render. In fact, I only need jadx to match method names instead of class names;
For example, as shown in the figure:
![方法名](https://user-images.githubusercontent.com/42903751/71474836-4d558d80-2818-11ea-87ff-85293bb63850.png)

4. Right-click on a package and search only on the selected package. Sometimes, the user locates a certain package, but jadx also presents the results of other packages, which interferes with judgment and consumes time;
![右键](https://user-images.githubusercontent.com/42903751/71474843-55adc880-2818-11ea-9803-02a898b54b6d.png)

Thank you for reading, and look forward to a more complete jadx, I wish you a happy life.

PR #2284, opened 2024-09-19
feat(gui): Limit search to a package

### LINKED ISSUE #1408: [feature] [gui] Search in specific path
*Describe your idea:*
Ability to search in specific path (specific package).
For example: search something only in this path (android.support.v4.app)

andyjsmith:
Adds the ability to limit search to a specific package (and sub-packages).

Often, you want to search for some code inside a specific package. The only way to do this currently is to add exclusions for all the packages you don't want, perform your search, and then include the packages again.

This PR adds:
- The ability to specify a package in the search window to limit the search to
- A "Text search" option to the right-click menu of all packages in the tree. This opens the search with the package name pre-filled in the field.

This works for all search providers (except for resources, which just ignores the restriction if enabled).

Build passes with `gradlew clean build dist`

This is my first contribution here so please feel free to let me know your questions/thoughts/concerns!

Resolves #1408 and part of #818

image

image

image

### DIFF
--- a/jadx-core/src/main/java/jadx/api/JavaClass.java
+++ b/jadx-core/src/main/java/jadx/api/JavaClass.java
@@ -252,6 +252,10 @@ public final class JavaClass implements JavaNode {
 		return cls.getPackage();
 	}
 
+	public JavaPackage getJavaPackage() {
+		return cls.getPackageNode().getJavaNode();
+	}
+
 	@Override
 	public JavaClass getDeclaringClass() {
 		return parent;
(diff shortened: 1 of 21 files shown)


skylot:
@andyjsmith great work :+1:

There is one minor possible improvement: split events processing for package search/verify with triggering search. So if package change we first search matching package object and if found than trigger big search. This way package field will be interactive and will notify about incorrect package even if "auto search" is disabled. Not sure if this really needed, so I might apply these changes myself later :slightly_smiling_face:

PR #2306, opened 2024-10-14
fix: don't rename resource entries when useRawResName = true

pubiqq:


skylot (approved):

skylot:
@pubiqq nice fix, thanks.

Issue #2215, opened 2024-07-12
[gui] Why not add code editing features


shinjaxu:
### Issue details

Is it a problem with the positioning of the software?

### Jadx version

1.5

### Java version

12

### OS

- [X] Windows
- [ ] Linux
- [ ] macOS

jpstotz:
As far as I know no such feature is planned. May be because decompiling, editing and recompiling is 100 times more difficult to implement. 

For simple APK files you can use the function "Save as gradle project" which will create a project that you can compile using Android SDK.  

For more complex apps I strongly recommend to use apktool.

Mino260806:
> May be because decompiling, editing and recompiling is 100 times more difficult to implement.
>

@jpstotz what if we only compile one java method at a time (not recompile the entire project, nor the entire class). If we consider how .smali files are structeredn all methods are independant from each others, so each method would still compile fine even when it's isolated.

For example say we have a class Example
```java
package com.example;

public class Example {
    int field1;
    Object field2;
    // ...
    String field_n;

    public void targetMethod() {
        System.out.println("To be edited");
    }

    public void methodOne() {
         // Very long body
    }

    public void methodTwo() {
         // ...
    }

    public void methodThree() {
         // ...
    }

   // and so on
}
```
Say the user would finish editing "targetMethod", Jadx would isolate the method in an internal temp .java file like this

```java
package com.example;

public class Example {
    int field1;
    Object field2;
    // ...
    String field_n;

    public void targetMethod() {
        System.out.println("Editing finished");
    }
}
```

It would also convert other compiled files to some `library.jar`

And then compile with the command
```
javac -cp library.jar com/example/Example.java
```

Then convert it using `baksmali` to .smali file, copy the target method only using simple regex to find it, paste it in the original .smali file and recompile the entire project from .smali files.

The only problem that remains to be solved is for invalid java class and method names (like those starting with digits) which will cause compilation error, maybe we'd do some naming convention that replaces the names back afterwards, but I'm sure it's solvable. Let's have a minimal working example first.

I can give it a shot if you confirm that it may be added

jpstotz:
In general you are correct for Java and class files but Jadx doesn't work on class files, but DEX files. I just outline the problems for modifying Android code as this is the main purpose of Jadx. Complexity increases when different source formats have to be considered.

[...]

PR #2310, opened 2024-10-21
Use jadx-gui as a library in plugins

nitram84:
For my latest plugin (https://github.com/nitram84/jadx-apkspy-plugin) I needed a compile dependency to jadx-gui. To build jadx-gui as a library I had to fix some issues, mostly javadoc issues. I'd like to remove this workaround for my plugin, so I would like to propose my changes as a PR.

### DIFF
--- a/jadx-gui/build.gradle.kts
+++ b/jadx-gui/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
 	id("jadx-kotlin")
 	id("application")
+	id("jadx-library")
 	id("edu.sc.seis.launch4j") version "3.0.6"
 	id("com.gradleup.shadow") version "8.3.3"
 	id("org.beryx.runtime") version "1.13.1"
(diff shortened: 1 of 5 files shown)


skylot:
@nitram84 I also add `jadx-library` for `jadx-cli` and `jadx-script-ide` modules because `jadx-gui` depends on them and gradle build fail if unable to find these.
Although, if anyone tries to start JadxGui from this library and opens preferences, app will crash with class not found error for `JFontChooser`. This class from jar in `jadx-gui/libs`, so it is not uploaded into maven. This is minor issue, and I will fix it later.

Anyway, using `jadx-gui` as a library should be used only as a workaround. Because code in jadx-gui often changes and this can break plugins :cry:

skylot (approved):

skylot:
`1.5.1-SNAPSHOT` updated, please check if everything is working fine :slightly_smiling_face:

nitram84:
@skylot Thank you for your work! Everything is working fine.
I don't have a problem if changes in `jadx-gui` break things in plugins - I'm sure there is a way to fix it in the plugin. Anyway my plugin is just a proof of concept to test if #2215 can be fixed with a plugin approach.

PR #2316, opened 2024-10-22
feat: improve resource names

pubiqq:
- Makes sanitized resource names more human-friendly (before: `jadx_deobf_<id>`, after: `<sanitized_name>_<id>`).
- Improves detection of incorrect characters in resource names.
- Fixes an issue where "better name" determination was not performed for styles.
- Fixes an issue where keywords could be used as a field name in the R class.

### DIFF
--- a/jadx-core/src/test/java/jadx/core/xmlgen/ResNameUtilsTest.java
+++ b/jadx-core/src/test/java/jadx/core/xmlgen/ResNameUtilsTest.java
@@ -1,0 +1,89 @@
+package jadx.core.xmlgen;
+
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ResNameUtilsTest {
+
+	@DisplayName("Check sanitizeAsResourceName(name, postfix, allowNonPrintable)")
+	@ParameterizedTest(name = "({0}, {1}, {2}) -> {3}")
+	@MethodSource("provideArgsForSanitizeAsResourceNameTest")
+	void testSanitizeAsResourceName(String name, String postfix, boolean allowNonPrintable, String expectedResult) {
+		assertThat(ResNameUtils.sanitizeAsResourceName(name, postfix, allowNonPrintable)).isEqualTo(expectedResult);
+	}
+
+	@DisplayName("Check convertToRFieldName(resourceName)")
+	@ParameterizedTest(name = "{0} -> {1}")
+	@MethodSource("provideArgsForConvertToRFieldNameTest")
+	void testConvertToRFieldName(String resourceName, String expectedResult) {
+		assertThat(ResNameUtils.convertToRFieldName(resourceName)).isEqualTo(expectedResult);
+	}
+
+	private static Stream<Arguments> provideArgsForSanitizeAsResourceNameTest() {
+		return Stream.of(
+				Arguments.of("name", "_postfix", false, "name"),
+
+				Arguments.of("/name", "_postfix", true, "_name_postfix"),
+				Arguments.of("na/me", "_postfix", true, "na_me_postfix"),
+				Arguments.of("name/", "_postfix", true, "name__postfix"),
+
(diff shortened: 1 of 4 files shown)


skylot:
@pubiqq this change looks great!
Although, I am hoping that someone also check and test this change :sweat_smile:

zhongqingsong:
> @pubiqq this change looks great! Although, I am hoping that someone also check and test this change 😅

hi, it looks interesting, i want test it. But i am not clear how to test, should I write a test class, or run the test locally? 😅

jpstotz:
@skylot I will try to make a review in the next days. Based on the description it sounds good. Biggest problem is to find good example APKs to test it's functionality.

@zhongqingsong Pull Requests are effectively branches you can check-out, compile. In the end have a modified Jadx version with the PR changes: https://docs.github.com/de/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

jadx-core/src/test/java/jadx/core/xmlgen/ResNameUtilsTest.java
```diff
@@ -0,0 +1,89 @@
...
+				Arguments.of("if", "_postfix", true, "if_postfix"),
+				Arguments.of("default", "_postfix", true, "default_postfix"),
+				Arguments.of("true", "_postfix", true, "true_postfix"),
+				Arguments.of("_", "_postfix", true, "__postfix"));
+	}
+
+	private static Stream<Arguments> provideArgsForConvertToJavaIdentifierTest() {
+		return Stream.of(
+				Arguments.of("ThemeDesign", "ThemeDesign"),
+				Arguments.of("Theme.Design", "Theme_Design"),
```
MrIkso, line 80: dot on style name it is normal, you don't need replace it to _

pubiqq: This test suite refers to the `convertToJavaIdentifier` function, which returns a java identifier.

MrIkso: Okay, what about names inside style.xml?

pubiqq: In resource names, a dot can be used anywhere except as the first character.

jpstotz:
Sorry for the delay, I needed to code some utilities first to be able to process multiple APKs one after another to get a better understanding when renaming is applied. 

I started with jadx.core.xmlgen.ResTableBinaryParser#getNewResName. For the few APK files I tested it the most common case was a renaming because the resource name started with a `$` like this:

`$avd_hide_password__0` renamed to `_avd_hide_password__0_res_0x7f080000` however the new resource name doesn't seem to be used anywhere (this was from the APK linked in #1232).

pubiqq:
>I started with jadx.core.xmlgen.ResTableBinaryParser#getNewResName. For the few APK files I tested it the most common case was a renaming because the resource name started with a $ like this:
>
>`$avd_hide_password__0` renamed to `_avd_hide_password__0_res_0x7f080000`

The dollar sign is not a valid character for resource names in aapt2, so `ResNameUtils#sanitizeAsResourceName` replaces it with an underscore and appends a postfix with the resource id. 

> however the new resource name doesn't seem to be used anywhere

They are used in drawable/avd_hide_password.xml.

jpstotz:
> > I started with jadx.core.xmlgen.ResTableBinaryParser#getNewResName. For the few APK files I tested it the most common case was a renaming because the resource name started with a $ like this:
> > `$avd_hide_password__0` renamed to `_avd_hide_password__0_res_0x7f080000`
> 
> The dollar sign is not a valid character for resource names in aapt2, so `ResNameUtils#sanitizeAsResourceName` replaces it with an underscore and appends a postfix with the resource id.

OK, got it. I saw that it was already renamed before your PR, but the new format used by your PR is indeed much better. 

Could you please extend your PR and add JavaDoc where you describe why something is done? The info what is done can be derived from the code but the reason why is often lost. That would really be helpful for the future...

> They are used in drawable/avd_hide_password.xml.

Seems like Jadx resources search does not yet allow searching for file-names, so I directly hit the blind spot of Jadx-Gui ;)

jpstotz:
@pubiqq I didn't had something special in mind, therefore what you have written is in my opinion sufficient. Thanks for your patience and of course of PR.

jpstotz (approved):

skylot (approved):

Issue #1020, opened 2020-11-18
[core] Some style item names aren't decoded


alissonlauffer:
File `res/values-v29/styles.xml`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <style name="e" parent="@android:style/Theme.Material.Light">
        <item name="android:statusBarColor">@color/f</item>
        <item name="android:navigationBarColor">@color/f</item>
        <item name="?0x10104e0">true</item>
        <item name="?0x101056c">true</item>
        <item name="?0x101056d">@android:color/transparent</item>
        <item name="?0x1010590">true</item>
    </style>
    ...
<resources>
```
The file should be decoded like this:

```xml
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <style name="e" parent="@android:style/Theme.Material.Light">
        <item name="android:statusBarColor">@color/f</item>
        <item name="android:navigationBarColor">@color/f</item>
        <item name="android:windowLightStatusBar">true</item>
        <item name="android:windowLightNavigationBar">true</item>
        <item name="android:navigationBarDividerColor">@android:color/transparent</item>
        <item name="android:isLightTheme">true</item>
    </style>
    ...
<resources>
```

APK: https://apkpure.com/via-browser-fast-light-geek-best-choice/mark.via.gp/download

jpstotz:
The `resources.arsc` file that is included in the resources of jadx-core project seems to be a bit outdated and therefore does not know all attribute names. If I use the `resources.arsc` from API 30 android.jar the attribute names are decoded correctly. 

However file size of the new `resources.arsc` file has increased from 11 to 33MB. Have you included the original resources.arsc file or was the old file a stripped one?

Additionally there is the problem that as far as I know newer API versions do not always include all resource IDs from previous versions. Some times old resource entries are removed (not many just a few) so that it might be useful to use a custom format that can be updated with every newly released Android version by adding all resource entries. @skylot What do you think?

skylot:
> it might be useful to use a custom format

@jpstotz yeah, right now I implementing simple converter from `.arsc` to text file with just mapping id to name (like stated in [this](https://github.com/skylot/jadx/blob/d1e5186d4adeabd51f52b42d5c07dd3cccd60f4b/jadx-core/src/main/java/jadx/core/xmlgen/entry/ValuesParser.java#L41) comment). And, as you mention, looks like it must accept several input `.arsc` files and merge them.

jpstotz:
[...]

PR #2318, opened 2024-10-23
chore: update android attrs (API 35)

pubiqq:


skylot:
@pubiqq thanks for update.
Although, it concerns me to see that too many IDs are shifted in `res-map.txt`. This means that these ids shouldn't be replaced with resource names. I will try to investigate this.

Related to #1020, #1035

pubiqq:
Nothing strange here: `res-map.txt` hasn't been updated for 5 versions of Android, and it still contains inaccessible resources that may not (and do not) have stable ids.

skylot (approved):

skylot:
> and it still contains inaccessible resources that may not (and do not) have stable ids

This is sad, but looks like we can't do anything here.

Anyway, I can reproduce result `res-map.txt`, so it is fine :+1:

PR #2319, opened 2024-10-24

zhongqingsong:
### Description
1. New text is translated simultaneously, two untranslated texts remain (no English text)
2. Adjusted and unified part of the symbol, into the full Angle symbol.


skylot:
@zhongqingsong thanks :+1:

> two untranslated texts remain (no English text)

Oh, I will fix this

PR #2320, opened 2024-10-24
chore: update tlds

pubiqq:


skylot:
@pubiqq looks good to me :+1:
One thing: previous implementation were also forbid rename of any subpackage and was more like allowed package names. Now it is working only for root packages, this way it is closer to original intention, but this might affect someone use case.

Related to #1369.

PR #2328, opened 2024-10-29
feat: create hooking code for all internal methods in the class

RuffaloLavoisier:
### Description

When writing hooking codes for a class, write hooking codes for all internal methods for the class.
 
hook_m


jpstotz:
If you want to hook all methods of a class I would recommend to use `frida-trace` instead of `frida`. There you can select classes and methods by pattern (wildcard).

RuffaloLavoisier:
@jpstotz 
Hi there 😄 
IMHO - It would be great if this feature could be added to JADX. For beginners, it would make writing hooking code quick and easy. Also, the current code snippet for the class simply returns the name in a JavaScript-like format, rather than actually hooking the class, which makes it less meaningful for writing hooking code.

skylot (approved):

PR #2338, opened 2024-11-05
fix: support "fall-through to default" case in switch-over-string

pubiqq:


skylot:
> This code definitely needs refactoring

I agree :rofl:
I somehow managed to get it working at least for simple case, so yeah code is messy, feel free to refactor it :+1:

> but if I refactored it, it would be impossible to understand what was actually changed to fix the issue

Well, if you already refactor it and fix the issue it maybe hard to split changes, so it is fine to mix them.

Also, for decompilation tests please provide some checks, at least for containing some string. And for simple tests it is better to add `check` method (see in other tests), it will be executed on decompiled code to verify result correctness.

skylot (approved):

PR #2358, opened 2024-12-03
feat: replace constants in annotations of method arguments

nitram84:
This PR replaces constants in annotation of method arguments. Example:

Having an annotation like 

```java
@Retention(RetentionPolicy.RUNTIME)
public @interface UsesAndroidResource {
	int value() default 0;
}
```

this class 

```java

public static class TestCls {
	public void test(@UsesAndroidResource(17039370 /* android.R.string.ok */) int i) {
	}
}
```

would be decompiled to

```java
public static class TestCls {
	public void test(@UsesAndroidResource(R.string.ok) int i) {
	}
}
```

I need this fix to generate an issue with AGP > 8.0.0, I'd like to fix later.


skylot (approved):
Nice fix! Thank you :+1:

PR #2361, opened 2024-12-09
fix: check respectBytecodeAccModifiers before changing method access modifier in MarkMethodsForInline

ewt45:
even when option "Respect bytecode access modifiers" is on, some method modifiers will still be changed. 
```java
    /* JADX INFO: Access modifiers changed from: private */
    public static String getPropertyNameOfSetter(String str) {
        if (str.startsWith("set")) {
            return str.substring(3);
        }
        return null;
    }
```

bugged dex: [1.zip](https://github.com/user-attachments/files/18058899/1.zip)

after inspect the code, i found that `Access modifiers changed from` is from `FixAccessModifiers.changeVisibility()`. 

it's a static method, so cannot use the boolean param in the same class. Instead, the caller of `changeVisibility()` should check the option by `root.getArgs().isRespectBytecodeAccModifiers()`

in the other place where `changeVisibility()` is called ([ClassModifier.java](https://github.com/skylot/jadx/blob/828cad32870514afb32bcf08a337eb3afaf36cc7/jadx-core/src/main/java/jadx/core/dex/visitors/ClassModifier.java#L287)), the options is checked, so here we should check it too.


skylot (approved):
Thanks :+1:

chanduBhai88:
Keywords

PR #2364, opened 2024-12-11
fix: inner class parent is wrong when name formatted as Class$method$xxx

ewt45:
inner class name format can be `Class1$InnerClass1`. but it can also be `Class1$method1$InnerClass1` ([example](https://github.com/skylot/jadx/blob/master/jadx-core/src/test/smali/inline/TestSyntheticInline3/TestSyntheticInline3%24onCreate%241.smali)). In [ClassInfo.splitAndApplyNames()](https://github.com/skylot/jadx/blob/47f2e516e54be4e91eee1fc56dd260ee634d0cfd/jadx-core/src/main/java/jadx/core/dex/info/ClassInfo.java#L161), only the former situation is considered. 

This pr is trying to cover the latter situation. If the inner class ClassInfo see its parent class as `Class1$method1`, parent classNode will not be found. In this case `resolveParentClass()` will try to split it again, into `Class1` and `method1`, make sure `Class1` exists and contains method `method1`, then set `Class1` as inner class's parent.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/nodes/RootNode.java
+++ b/jadx-core/src/main/java/jadx/core/dex/nodes/RootNode.java
@@ -290,7 +290,7 @@ public class RootNode {
 		List<ClassNode> updated = new ArrayList<>();
 		for (ClassNode cls : inner) {
 			ClassInfo clsInfo = cls.getClassInfo();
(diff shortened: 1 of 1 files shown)


ewt45:
it seems like there're some more situation to considerate. I'll check with more examples.

skylot (approved):
Nice fix! Thanks :+1:

PR #2367, opened 2024-12-12
fix: inline anonymous class wrongly handles field as Classname.this

ewt45:
If option "inline anonymous class" is checked, sometimes the field of the anonymous class is wrongly recoginized as the instance of its parent class, which actually is instance of a random class. This leads to the field being added `CLASS_INSTANCE` type `FieldReplaceAttr` and displayed as `FieldClassName.this` rather than its normal name.

[example dex](https://github.com/user-attachments/files/18104562/classes.dex.zip)
decompiled code: notice that `another` is displayed as `AnotherClass.this`
```java
public class Parent {
    public static Parent test(Class<?> cls) {
        final AnotherClass another = new AnotherClass();
        return new Parent() { // from class: jadxtest.Parent.1
            @Override // jadxtest.Parent
            public String func() {
                return AnotherClass.this.toString();
            }
        };
    }

    public String func() {
        return "";
    }
}
```

this pr extracts `Objects.equals(parentClass, fieldsCls.getClassInfo())`, and only add CLASS_INSTANCE replace attr if it is true.


skylot (approved):
Great! Thank you :+1:

Issue #1967, opened 2023-07-30
[core] Package-info won't get decompiled/generated


nitram84:
In https://github.com/skylot/jadx/issues/1795 I mentioned some special cases if the source name is package-info.java or module-info.java. In a package-info it's possible to write package private classes or to annotate the package. Jadx renames package-info and module-info to "packageinfo" and "moduleinfo" but doesn't generate code for these special classes. When I created the issue I thought package-info isn't android relevant, because usually package-infos are ignored when building a final apk. But now I found an apk containing package-infos: https://java-program.apk.gold/ (version 1.8) e.g. package "google.android.material.internal" or "google.android.material.internal"

In my opinion jadx should consider "package-info" and "module-info" as valid source/class names and not try to rename it and jadx should generate code for it.

It's possible to simulate this issue with plain java:

```java
@XmlSchema(elementFormDefault = XmlNsForm.QUALIFIED, namespace = "http://schemas.android.com/apk/res/android")
package jadx.packageinfo;

import jakarta.xml.bind.annotation.XmlSchema;
import jakarta.xml.bind.annotation.XmlNsForm;
```

or simpler without dependencies

```java
@Deprecated
package jadx.packageinfo;
```

For both package-infos jadx doesn't produce any output and apps would not work as expected (first case).

skylot:
Done for `package-info` :+1:
Implementation for `module-info` is harder, and I suggest opening a separate issue.

PR #2369, opened 2024-12-14
fix: better package-info support

nitram84:
When I filed https://github.com/skylot/jadx/issues/1967 I had only one sample app. With https://apkpure.com/who-becomes-rich/de.sellfisch.android.wwr/download/1.15.1 I found another app with package-infos e.g. `javax.annotation.package-info`. This time a package-info is just an interface, nothing else. I removed some checks to support this case too.


skylot:
I think just checking for name `package-info` is enough. Access flags or even class type is not specified in java docs, so we can't rely on them (package info only mentioned in JLS [here](https://docs.oracle.com/javase/specs/jls/se21/html/jls-7.html#jls-7.4.1))
As a guard from protectors, we can also check that there are no fields and methods in that 'class'.

skylot (approved):

skylot:
@nitram84 thanks :+1:

Issue #2353, opened 2024-11-26
[feature] Confirmation dialog when removing scripts


iscle:
### Describe your idea

It would be nice to have a confirmation dialog when removing scripts, to avoid removing them by clicking on the remove option accidentally :)

PR #2375, opened 2024-12-19
feat(gui): added confirm dialog when removing script

MrIkso:
This PR adds confirm dialog when removing script. Closes https://github.com/skylot/jadx/issues/2353


iscle:
Thanks!

PR #2376, opened 2024-12-19
fix(gui):text on popup action change if need update comment & fixes default size on comment diallog 

### LINKED ISSUE #2155: Comment text box not shown
image

There is no way to write comment

MrIkso:
This PR fixes #2155


skylot:
@MrIkso thanks for the fix. I updated it to just rename labels, instead of adding and changing to a new action. That new action also appears in key binding settings, and this it feels not right.

skylot (approved):

MrIkso:
@skylot thanks, I first used `setName ` but replaced it to set the new action. Okay, I'll keep that in mind.

PR #2378, opened 2024-12-21
feat(gui):add export dialog with options

### LINKED ISSUE #1983: [feature] Add export dialog with options
It will be nice to have a single export dialog with options:
- export as a Gradle project
- don't save classes from excluded packages
- don't save resources
- maybe others

Now excluded packages not applied, and other options only available in preferences dialog.
Also, current name `Save All` not clear and can interfere with saving project actions, so it is better to rename it to `Export`.

Original discussion started in https://github.com/skylot/jadx/issues/1085#issuecomment-770338533

MrIkso:
This PR adds export dialog. Closes #1983
Preview dialogue

image


skylot (approved):
Good! Thank you :+1:

Issue #2373, opened 2024-12-19
[gui] res/values/ directory not showing up


razaina:
### Issue details

Hi,

I am trying the new version 1.5.1, and I just figured out that the `res/values/ ` directory is not visible anymore, whereas one can see that when you search for a string that you expect to be - let's say in the `strings.xml` - the search windows shows that it is in `res/values/ strings.xml` .

![a](https://github.com/user-attachments/assets/cabfcccf-f622-44d1-90fe-cf445758d084)

![b](https://github.com/user-attachments/assets/63ecf6e7-1328-43da-8eb9-f533e43f529e)

### Jadx version

1.5.1

### Java version

17.0.12

### OS

- [ ] Windows
- [X] Linux
- [ ] macOS

MrIkso:
This is normal, strings.xml is included in resources.arsc. Click in resources.arsc node and yous seen other values resources.

skylot:
Hm, I agree this is confusing, I think we need to use 'full paths' with source file as a prefix, like this:
`resources.arsc:/res/values/strings.xml`
Such rename should be implemented carefully, because it might affect export.

Another possible approach is to add a dedicated section 'Android Resources' at the left panel and show merged resource files. But this might be also confusing, because it will duplicate some entries from current 'Resources' section (which is actually shows all loaded files and content)

MrIkso:
> like this:
> `resources.arsc:/res/values/strings.xml`

I think it's best variant, showing  it's virtual path only on search window.

skylot:
> resources.arsc:/res/values/strings.xml

Done

Issue #2009, opened 2023-09-15
[gui] Improve result highlight in Usage and Search dialogs


skylot:
### Describe your idea

Now for highlight search result in Usage dialog used text search in code line, but jadx knows exact node place and should use that info to correctly highlight code.
Same approach should be used in Search dialog for Class/Method/Field nodes search.

This issue created from discussion started in #2006.

MrIkso:
Hi, I'm currently trying to improve the search dialog by replacing the table with a tree. So the first results look like this, it will also fix #2373 and maybe other issues. Work not finished

![image](https://github.com/user-attachments/assets/8b572021-54e6-4f8d-a4b4-53936395c39e)
![image](https://github.com/user-attachments/assets/719500b9-d7bf-4a0a-933b-3794a6a148cd)

skylot:
@MrIkso not sure why you decided to redone UI, because this issue is more about highlighting in code view for usage dialog and only partially in search dialog. And #2373 is just fixing resource name.

Anyway, main focus in table version of UI was on code view, because often class names are obfuscated and don't help in understanding code logic. So I don't think such rework is really needed, I suggest opening a draft PR for this change, maybe someone else will share another opinion or suggestion.

Issue #1130, opened 2021-03-07
Search Comments and Resources does not show full path


green9317:
Currently when searching on class, method, field, and code Jadx shows the full node path within the Text Search window.  When searching on Resource and Comments Jadx only shows the class / file without the path.  This makes it a bit harder to identify which entry I am looking for and it feels a bit odd that its not the same functionality for all search options.

green9317:
That is true.  Perhaps another good issue / suggestion would be to make a custom key bindings page so i can see the current bindings and change them if desired.

Obviously, this is a very minor issue that really only helps with polishing Jadx.

Issue #2237, opened 2024-08-02
[gui] Use a sophisticated docking framework


Mino260806:
### Describe your idea

This idea origined from PR #2230 
A docking framework would allow users to move code tabs and optional panels freely inside jadx gui, where they would be able to make custom split views in any direction, or move a tab to a seperate window.

@skylot suggested using [ModernDocking](https://github.com/andrewauclair/ModernDocking), but it's still unstable, (at least on linux, see https://github.com/andrewauclair/ModernDocking/issues/218 for example, it's unusable for me).

I'm ready to implement this feature, but I don't know what framework I should use

PR #2383, opened 2024-12-27
feat(gui):improve search and usage dialog

MrIkso:
This PR improves Search and Usage dialogs. Releated to #2009 , #2373, #1130

**Main features:**

1. Added Whole Word find option
2. Highlights text for found Classes/Methods/Filelds
3. Tree structure for code/comment/resource results

**Removed**

1. Open button - unuseless feature for curently view
2. Sort button - results sorted by name defaults

**Limitations**

1. Curently search available only one selected option
2. When Method/Field name used highligt type not formatted html

<details>
  <summary>Preview</summary>

|                 | Old | New |
|-----------------|-----|-----|
| Class Search    | Old_Search_Class_2024-12-27_02-14 | Search_Class_2024-12-27_02-14 |
| Methods Search  |  Old_Search_Method_2024-12-27_02-14   | Search_Method_2024-12-27_02-14    |
| Field Search    |  Old_Search_Field_2024-12-27_02-14 |  Search_Fileld_2024-12-27_02-14   |
| Code Search     |   Old_Search_Code_2024-12-27_02-14  |  Search_Code_2024-12-27_02-14 |
| Resource Search |  Old_Sesrch_Res_2024-12-27_02-57   |  Search_Ressource1_2024-12-27_02-14 Search_Ressource2_2024-12-27_02-14   |
| Usage |   Old_Usage_2024-12-27_02-50  |    Usage_2024-12-27_02-50 |

</details>

P.S maybe there will be more ideas how to make the appearance of the search results the best, maybe combine the table and the tree (TreeTable).

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/search/SearchManager.java
+++ b/jadx-gui/src/main/java/jadx/gui/search/SearchManager.java
@@ -1,0 +1,39 @@
+package jadx.gui.search;
+
+import java.util.regex.Pattern;
+
+public class SearchManager {
+	/**
+	 * Generate the good pattern according to the different boolean
+	 *
+	 * @param exp           the searched expression
+	 * @param caseSensitive boolean
+	 * @param wholeWord     boolean
+	 * @param useRegexp     boolean
+	 * @return the pattern
+	 */
+	public static Pattern generatePattern(String exp, boolean caseSensitive, boolean wholeWord, boolean useRegexp) {
+		String word = exp;
+		if (word != null && !word.isEmpty()) {
+			if (!useRegexp) {
+				word = word.replace("\\E", "\\E\\\\E\\Q");
+				word = "\\Q" + word + "\\E";
+				if (wholeWord && exp.matches("\\b.*\\b")) {
+					word = "\\b" + word + "\\b";
+				}
+			}
+
+			if (!caseSensitive) {
+				word = "(?i)" + word;
+			}
+
+			if (useRegexp) {
+				word = "(?m)" + word;
+			}
+
+			return Pattern.compile(word);
+		} else {
+			return Pattern.compile("");
+		}
+	}
+}
--- a/jadx-gui/src/main/java/jadx/gui/search/ISearchMethod.java
+++ b/jadx-gui/src/main/java/jadx/gui/search/ISearchMethod.java
@@ -3,25 +3,19 @@ package jadx.gui.search;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import org.apache.commons.lang3.StringUtils;
-
 public interface ISearchMethod {
-	int find(String input, String subStr, int start);
+	MatchingPositions find(String input, int start);
 
 	static ISearchMethod build(SearchSettings searchSettings) {
-		if (searchSettings.isUseRegex()) {
-			Pattern pattern = searchSettings.getPattern();
-			return (input, subStr, start) -> {
-				Matcher matcher = pattern.matcher(input);
-				if (matcher.find(start)) {
-					return matcher.start();
-				}
-				return -1;
-			};
-		}
-		if (searchSettings.isIgnoreCase()) {
-			return StringUtils::indexOfIgnoreCase;
-		}
-		return String::indexOf;
+		Pattern pattern = searchSettings.getPattern();
+		return (input, start) -> {
+			Matcher matcher = pattern.matcher(input);
+			if (matcher.find(start)) {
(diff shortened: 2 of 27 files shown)


skylot:
Well, it is sad that this PR don't get any suggestions or comments.

Anyway, I have some thoughts:
At first, we will need a tree representation in usage dialog to implement references window (#1368). That is an improvement of current usage dialog with addition of all type of references like class hierarchy, method overrides, etc. and not only usage. For this, collapsible sections are the good approach, so the tree looks like the best component.
Although in search, code preview looks better in a table component in a separate column, because obfuscated code have short names so classes/methods names became useless and tree pointless.
So, as suggested by @MrIkso, `TreeTable` component will be a nice approach to combine above requirements and allow using various representation as needed.
Some implementation details:
- create a customizable tree table component which allows inserting simple text nodes (needed for references window)
- in code search add a switch to change representation: tree or table (should change a 'model' for new component)
- use new component in usage dialog with a single `Usage` top level label (others will be added later)

Such approach will allow using single component in all places to represent jadx nodes in various ways.

@MrIkso this looks like a big task, so you may change this PR, or I can merge it into a new branch and implement these myself or with your help if you want to.

MrIkso:
@skylot yeah ThreeTable it's best variant. I did something like this once, it's a bit complicated but doable. My current implementation has one problem, if the string found is long, it window freezes. Otherwise everything seems to be working.

iscle:
It looks very nice! I think this was a very much needed task to be performed, I'm glad it's already being worked on.

I have one suggestion. It would be nice if the nesting in packages/classes could be disabled, and have everything shown in a single node like it's implemented right now. In most cases this will never be desired, but sometimes when looking at strings or some common code, it's appreciated to see it next to each others to spot differences quickly.

Thanks!

MrIkso:
@skylot I decided to see how search is implemented, find usages in netbeans ide. I think something similar can be done here? We can even seen code how everything)

<details>
  <summary>Preview</summary>
  
image

image

image

image

image

</details>

There everything is not in a separate dialog but at the bottom of the screen, which in principle can be tried to be implemented here, also with tabs(?). I also thought to take the editor from there, it is more productive, but it implemet very hard.

skylot:
@MrIkso  I see, so results are in tree with preview panel, and this looks very similar to implementation in this PR :slightly_smiling_face:
But, as I mention early, in my opinion, tree representation is not working for obfuscated classes, that is why current implementation heavily focused on code lines, so it is easier to scan results without mixing in classes names.
Also, I think for current implementation preview is better to put at the bottom, because results table is wide.

> There everything is not in a separate dialog but at the bottom of the screen, which in principle can be tried to be implemented here

Sure, there is already an issue for start using dockable framework (#2237), but all docking frameworks are very old and abandoned, only active is a [ModernDocking](https://github.com/andrewauclair/ModernDocking), we might use it once it became stable enough.
As another approach, I implement dock/undock manually for logs, it works fine, but docked position is fixed through.

jadx-gui/src/main/java/jadx/gui/search/SearchManager.java
```diff
@@ -0,0 +1,39 @@
...
+	 * Generate the good pattern according to the different boolean
+	 *
+	 * @param exp           the searched expression
+	 * @param caseSensitive boolean
+	 * @param wholeWord     boolean
+	 * @param useRegexp     boolean
+	 * @return the pattern
+	 */
+	public static Pattern generatePattern(String exp, boolean caseSensitive, boolean wholeWord, boolean useRegexp) {
+		String word = exp;
```
MewtR, line 16: I think it's probably better to use StringBuilder here until you're done processing the string

jadx-gui/src/main/java/jadx/gui/search/SearchManager.java
```diff
@@ -0,0 +1,39 @@
...
+		if (word != null && !word.isEmpty()) {
+			if (!useRegexp) {
+				word = word.replace("\\E", "\\E\\\\E\\Q");
+				word = "\\Q" + word + "\\E";
+				if (wholeWord && exp.matches("\\b.*\\b")) {
```
MewtR, line 21: Why is the second condition needed here? `exp.matches("\\b.*\\b")`? Isn't just the `wholeWord` flag enough?

MewtR:
@MrIkso Thanks for working on this. I was wondering why there was no 'Whole word' search feature so I went ahead and implemented it myself but your version seems better. Really looking forward to it.

Issue #2384, opened 2024-12-30
[core] IfRegionMaker find the wrong outBlock


ewt45:
### Issue details

jadx failed to decompiled a method in [this smali class](https://github.com/user-attachments/files/18272175/TestOutBlock.zip).
```java
/* JADX WARN: Multi-variable type inference failed */
    /* JADX WARN: Removed duplicated region for block: B:10:0x0065  */
    /* JADX WARN: Removed duplicated region for block: B:8:0x0064 A[RETURN] */
    /*
        Code decompiled incorrectly, please refer to instructions dump.
    */
    protected void onCreate(Bundle bundle) {
        ViewFacade viewFacade;
        super.onCreate(bundle);
        ApplicationStateBase applicationState = getApplicationState();
        XServerComponent component = applicationState.getEnvironment().getComponent(XServerComponent.class);
        Class cls = (Class) getIntent().getSerializableExtra("facadeclass");
        if (cls != null) {
            try {
                viewFacade = (ViewFacade) cls.getDeclaredConstructor(XServer.class, ApplicationStateBase.class).newInstance(component.getXServer(), applicationState);
            } catch (Exception unused) {
                Assert.state(ENABLE_TRACING_METHODS);
            }
            getWindow().addFlags(128);
            getWindow().addFlags(4194304);
            setContentView(R.layout.main);
            if (checkForSuddenDeath()) {
                this.viewOfXServer = new ViewOfXServer(this, component.getXServer(), viewFacade, applicationState.getXServerViewConfiguration());
                this.periodicIabCheckTimer.start();
                return;
            }
            return;
        }
        viewFacade = null;
        getWindow().addFlags(128);
        getWindow().addFlags(4194304);
        setContentView(R.layout.main);
        if (checkForSuddenDeath()) {
        }
    }
```

by looking into smali, the correct java code should be like 
```java
if (cls != null) {
    try {
        viewFacade = (ViewFacade) cls.getDeclaredConstructor(XServer.class, ApplicationStateBase.class).newInstance(component.getXServer(), applicationState);
    } catch (Exception unused) {
        Assert.state(ENABLE_TRACING_METHODS);
        viewFacade = null;
    }
} else {
    viewFacade = null;
}
getWindow().addFlags(128);
//...
```

[...]

PR #2385, opened 2024-12-30
fix: IfRegionMaker find the wrong outBlock

ewt45:
this pr tries to solve the issue #2384 

when checking whether the then/else block should be extracted as out block, 
- firstly use `getPathCross` to get the cross point(outBlock) of the two blocks. if it's not null, the codes will simply compare out block with then/else block.  
- otherwise the code will fall back to the old method `allPathsFromIf`.

Inside the method `getPathCross`: 
- start block itself is also considered as possible cross point now. 
- when building excluded bitset, add `handlerBlock.getId()` instead of `handlerBlock.getDomFrontier()`. this is to fix a test class failure after the changing above. I'm not sure if this will break other cases.
- remove `combinedDF.clear(block.getId());` in the while loop. it seems that this line is trying to remove the current block after next domFront blocks are added, but current block is stored at `domFrontBS` rather than `combinedDF`.

----
after these changes, now decompiled result is much better
- out block is correct
- however catch block lacks of a line `viewFacade = null;`
```java
    /* JADX WARN: Multi-variable type inference failed */
    protected void onCreate(Bundle bundle) {
        ViewFacade viewFacade;
        super.onCreate(bundle);
        ApplicationStateBase applicationState = getApplicationState();
        XServerComponent component = applicationState.getEnvironment().getComponent(XServerComponent.class);
        Class cls = (Class) getIntent().getSerializableExtra("facadeclass");
        if (cls != null) {
            try {
                viewFacade = (ViewFacade) cls.getDeclaredConstructor(XServer.class, ApplicationStateBase.class).newInstance(component.getXServer(), applicationState);
            } catch (Exception unused) {
                Assert.state(ENABLE_TRACING_METHODS);
            }
        } else {
            viewFacade = null;
        }
        getWindow().addFlags(128);
        getWindow().addFlags(4194304);
        setContentView(R.layout.main);
        if (checkForSuddenDeath()) {
            return;
        }
        this.viewOfXServer = new ViewOfXServer(this, component.getXServer(), viewFacade, applicationState.getXServerViewConfiguration());
        this.periodicIabCheckTimer.start();
    }
```

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/regions/maker/IfRegionMaker.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/regions/maker/IfRegionMaker.java
@@ -178,6 +178,8 @@ final class IfRegionMaker {
 			info.setOutBlock(null);
 			return info;
 		}
+		// init outblock, which will be used in isBadBranchBlock to compare with branch block
+		info.setOutBlock(BlockUtils.getPathCross(mth, thenBlock, elseBlock));
 		boolean badThen = isBadBranchBlock(info, thenBlock);
 		boolean badElse = isBadBranchBlock(info, elseBlock);
 		if (badThen && badElse) {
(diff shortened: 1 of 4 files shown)


skylot (approved):

skylot:
@ewt45 looks good! Thank you :+1:

Issue #1392, opened 2022-02-25
[feature] Install desktop from file from jadx-gui


yotamN:
It would be nice, just like with Intellij to support adding a desktop entry so people could launch Jadx properly through their desktop environment.

There needs to be several conditions to handle cases when that action is unwanted:
- hide it if the user is not on Linux
- hide it an environment variable/build flag is set (so distributions that handle the desktop entries themselves could disable it)
- hide it if the user already installed it (or maybe add an option to remove it)

I think that best way would be to show a popup the first time Jadx is opened and afterward showing it in the menubar.

skylot:
@yotamN I thought that already done by package maintainers.
But I understand that jadx itself missing a nice packaging for different distros or installer for Windows or macOS.
And I think such packaging is a better way to correctly handle desktop files, icons, and other OS/distro dependent files.

> just like with Intellij 

Do you mean `JetBrains Toolbox` or this is can be done directly from IntelliJ apps? Just asking, maybe I miss something :)


Also, beside packaging, I am thinking about creating something like `jadx-launcher` to:
- auto download latest jadx version and easy switch between release and unstable channels
- download/update the latest or custom Java Runtime version (useful for Windows users)
- provide a way to change java startup settings, like max memory (-Xmx)
- other stuff like desktop file, launch scripts, etc.
- and because I want to manage java, it is better to write this app with *not a java* :rofl:

yotamN:
Ideally Jadx would be packaged in all of the common Linux distributions but we are not there yet. It seems that Java packaging is still a bit of a problem, for example I tried to package Jadx into Fedora but it seems they don't ship Gradle, only Maven, so if you have any idea how to proceed I would love to hear because I prefer the user the package manager.

But even if we could package it into all of the popular distributions, there would still be people that can't use it for whatever reason (maybe they use the minimalist distribution of the month) and for them it would be helpful to create a desktop entry easily. 

---

> Do you mean JetBrains Toolbox or this is can be done directly from IntelliJ apps? Just asking, maybe I miss something :)

You can create a desktop entry from Intellij apps like so:
![Screenshot from 2022-02-25 14-47-56](https://user-images.githubusercontent.com/5249239/155717811-ba78c38c-42e4-4e6a-99b5-844b62cd44b5.png)

---

> Also, beside packaging, I am thinking about creating something like jadx-launcher

[...]

PR #2388, opened 2025-01-01
chore: add .desktop file for gui

Bnyro:
### Description
This pull request adds a `.desktop` file for the GUI which can be copied to `/usr/share/application` on Linux to have a desktop entry to start the program using an application launcher/menu instead of starting it via the terminal via `jadx-gui`. I'm not sure if there's a better place for it, I placed it in a newly created `contrib` folder now as that's what most projects do.

This PR is for having a desktop file entry at Void Linux due to https://github.com/void-linux/void-packages/pull/53772#discussion_r1900208774, as we'd like to upstream desktop files and not vendor them ourselves.

If there are suggestions for improvements, feedback is always welcome! Thanks.

### DIFF
--- a/contrib/jadx-gui.desktop
+++ b/contrib/jadx-gui.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Name=JADX GUI
+Comment=Dex to Java compiler
+Icon=jadx
+Exec=jadx-gui %f
+Terminal=false
+Type=Application
+Categories=Development;Java;
+Keywords=Java;Decompiler;


skylot:
> we'd like to upstream desktop files and not vendor them ourselves

Sure, looks good to me, thanks :+1:

Related to #1392.
Also icons can be found here: https://github.com/skylot/jadx/tree/master/jadx-gui/src/main/resources/logos

skylot (approved):

Bnyro:
> Sure, looks good to me, thanks 👍
> 
> Related to #1392. Also icons can be found here: https://github.com/skylot/jadx/tree/master/jadx-gui/src/main/resources/logos
>
Thanks for the quick merge :+1:

Regarding the linked issue, I'm happy that Jadx is packaged in Void Linux (and apparently in Arch too), so the amount of distros packaging Jadx are increasing slowly but continuouly :)

PR #2404, opened 2025-02-01
feat: add button to install desktop file directly from jadx-gui

Hidoni:
This PR is meant to address #1392.  
While it would be ideal if all distros provided JADX and a matching desktop entry, this is unfortunately still not the case several years after the issue was first created. As someone who uses JADX on a distro without one, I know this is a convenience feature I appreciate JetBrains IDEs having and I think JADX could benefit from it too :)

I implemented this feature by adding a new button to the `Help` dropdown menu at the top of the screen which is only visible on systems where desktop files exist (UNIX, except for MacOS). It uses the standardized xdg commands for installing icons and desktop files to hopefully ensure maximum compatibility.  
While working on the PR I noticed several incorrect values in the already provided desktop file that I changed too:
- The comment was incorrect (compiler -> decompiler)
- There was no StartupWMClass which meant the desktop environment could not map the icon to the process itself correctly (at least from my testing on GNOME).

### DIFF
--- a/jadx-gui/build.gradle.kts
+++ b/jadx-gui/build.gradle.kts
@@ -110,11 +110,19 @@ project.components.withType(AdhocComponentWithVariants::class.java).forEach { c 
 
 tasks.startShadowScripts {
 	doLast {
-		val newContent =
+		val newWindowsScriptContent =
 			windowsScript.readText()
 				.replace("java.exe", "javaw.exe")
 				.replace("\"%JAVA_EXE%\" %DEFAULT_JVM_OPTS%", "start \"jadx-gui\" /B \"%JAVA_EXE%\" %DEFAULT_JVM_OPTS%")
-		windowsScript.writeText(newContent)
+		// Add launch script path as a property
+		val newUnixScriptContent =
+			unixScript.readText()
+				.replace(
+					Regex("DEFAULT_JVM_OPTS=.+", RegexOption.MULTILINE),
+					{ result -> result.value + "\" \\\"-Djadx.launchScript.path=\$(readlink -e -- $0)\\\"\"" },
+				)
+		windowsScript.writeText(newWindowsScriptContent)
+		unixScript.writeText(newUnixScriptContent)
 	}
 }
 
--- a/jadx-gui/build.gradle.kts
+++ b/jadx-gui/build.gradle.kts
@@ -119,7 +119,7 @@ tasks.startShadowScripts {
 			unixScript.readText()
 				.replace(
 					Regex("DEFAULT_JVM_OPTS=.+", RegexOption.MULTILINE),
-					{ result -> result.value + "\" \\\"-Djadx.launchScript.path=\$(readlink -e -- $0)\\\"\"" },
+					{ result -> result.value + "\" \\\"-Djadx.launchScript.path=\$(realpath $0)\\\"\"" },
 				)
 		windowsScript.writeText(newWindowsScriptContent)
 		unixScript.writeText(newUnixScriptContent)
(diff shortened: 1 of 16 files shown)


skylot (approved):

skylot:
@Hidoni looks great! Thank you for your work :+1:

jadx-gui/build.gradle.kts
```diff
@@ -110,11 +110,19 @@ project.components.withType(AdhocComponentWithVariants::class.java).forEach { c
...
 			windowsScript.readText()
 				.replace("java.exe", "javaw.exe")
 				.replace("\"%JAVA_EXE%\" %DEFAULT_JVM_OPTS%", "start \"jadx-gui\" /B \"%JAVA_EXE%\" %DEFAULT_JVM_OPTS%")
-		windowsScript.writeText(newContent)
+		// Add launch script path as a property
+		val newUnixScriptContent =
+			unixScript.readText()
+				.replace(
+					Regex("DEFAULT_JVM_OPTS=.+", RegexOption.MULTILINE),
+					{ result -> result.value + "\" \\\"-Djadx.launchScript.path=\$(readlink -e -- $0)\\\"\"" },
```
skylot, line 122:
```suggestion
					{ result -> result.value + "\" \\\"-Djadx.launchScript.path=\$PRG\\\"\"" },
```
Looks like startup script already has full path to executable in `PRG` variable

jadx-gui/src/main/java/jadx/gui/utils/DesktopEntryUtils.java
```diff
@@ -0,0 +1,130 @@
...
+		}
+		if (!writeDesktopFile(launchScriptPath, desktopTempFile)) {
+			return false;
+		}
+		return installDesktopEntry(desktopTempFile);
+	}
+
+	private static boolean installDesktopEntry(Path desktopTempFile) {
+		try {
+			ProcessBuilder desktopFileInstallCommand = new ProcessBuilder("xdg-desktop-menu", "install", desktopTempFile.toString());
```
skylot, line 62:
```suggestion
			ProcessBuilder desktopFileInstallCommand = new ProcessBuilder("/usr/bin/xdg-desktop-menu", "install", desktopTempFile.toString());
```
Github code scanner suggest to use full path to execute binary, is `/usr/bin/` common for all distros?

jadx-gui/src/main/java/jadx/gui/utils/DesktopEntryUtils.java
```diff
@@ -0,0 +1,130 @@
...
+			return false;
+		}
+		LOG.info("Successfully installed desktop file");
+		return true;
+	}
+
+	private static boolean installIcon(int size, Path iconPath) {
+		try {
+			ProcessBuilder iconInstallCommand =
+					new ProcessBuilder("xdg-icon-resource", "install", "--novendor", "--size", String.valueOf(size), iconPath.toString(),
```
skylot, line 80:
```suggestion
					new ProcessBuilder("/usr/bin/xdg-icon-resource", "install", "--novendor", "--size", String.valueOf(size), iconPath.toString(),
```

skylot (changes requested):
Found some possible improvement for start script and some security issues, please check :slightly_smiling_face:

jadx-gui/build.gradle.kts
Hidoni: Unfortunately it doesn't :(  
$PRG is only resolved if $0 is a symlink, and even then from my testing it will be resolved into a relative path if $0 is one, which means it is not suitable for us as we need the full path in the desktop file  
While checking this I also realized `readlink` is a part of GNU coreutils, so I swapped it out for the similar `realpath` command which is a part of POSIX

jadx-gui/src/main/java/jadx/gui/utils/DesktopEntryUtils.java
Hidoni: Although common, it is not guaranteed  
I added another function named `findExecutablePath` which will check `$PATH` for the executable and use the first valid match, which as far as I know is how you are supposed to find executables whose path is not known

skylot (approved):

PR #2419, opened 2025-02-22
Feature: Tab UI improvements

yyyair:
This PR implements some minor improvements for TabbedPane and TabComponent.

Main features:
1. Added a "Reveal In Tree" option in TabComponent's popup menu, similar to "Reveal In Explorer" in vscode and other IDEs.
2. Fixed a problem that occurs when clicking the margin between the tab and the actual TabComponent. Before the fix it won't register as a click, since the TabComponent is smaller than the tab itself (so you could hover the edge of a tab with your mouse, have it marked as "hovered", but when you right-click it nothing happened).
3. Also added a separator after the bookmark options, since it was missing.


skylot:
Looks good.
But adding new name string for already exist action is not good.
We already have 2 different names for this action and I don't want to make it worse :rofl:
```properties
# same action as your 'Reveal in tree'
menu.sync=Sync with editor

# auto select active tab node in tree
menu.alwaysSelectOpened=Always Select Opened File/Class
```
We need to use the same action name in all places.
Maybe `Sync with editor` is a bad name, because people often ask to add this feature in issues :worried:
For example, name of this action in other apps:
- `Select Opened File` in IntelliJ
- `Reveal in Explorer View` in VS Code
- `Link with Editor` (auto select) in Eclipse

I like `Select in Tree`, but if anyone have a better name, feel free to share.

And for this PR, I removed a new string and replace it with exist `menu.sync`. A name change I will commit later.

skylot (approved):

PR #2344, opened 2024-11-15
BufferedInputStream: Allow jumping backwards to an already processed position 

jpstotz:
Some APK resources.arsc seem to use disordered entries, which collides with our approach to read them sequentially from the InputStream. 

As a workaround this PR introduces a BufferedInputStream and mark/reset support for `ParserStream` which allows us to jump back (up to 32k) in stream to read the disordered type chunk entries.

fixes #2343

Issue #2343, opened 2024-11-14
Jadx can not parse resource type chunks with unordered entries


shifoc:
### Issue details

https://www.apkmirror.com/apk/whatsapp-inc/whatsapp/whatsapp-2-24-24-14-release/whatsapp-messenger-2-24-24-14-android-apk-download/

### Relevant log output or stacktrace

```java
ERROR - Failed to parse 'resources.pb'/'.arsc' file
jadx.core.utils.exceptions.JadxException: Error decode: resources.arsc
	at jadx.api.ResourcesLoader.decodeStream(ResourcesLoader.java:118)
	at jadx.core.dex.nodes.RootNode.loadResources(RootNode.java:212)
	at jadx.api.JadxDecompiler.load(JadxDecompiler.java:128)
	at jadx.cli.JadxCLI.processAndSave(JadxCLI.java:59)
	at jadx.cli.JadxCLI.execute(JadxCLI.java:37)
	at jadx.cli.JadxCLI.main(JadxCLI.java:27)
Caused by: java.io.IOException: Expected start of entry 1, expected offset: 0x1f02f4, actual: 0x1f0304
	at jadx.core.xmlgen.ParserStream.checkPos(ParserStream.java:125)
	at jadx.core.xmlgen.ResTableBinaryParser.parseTypeChunk(ResTableBinaryParser.java:289)
	at jadx.core.xmlgen.ResTableBinaryParser.parsePackage(ResTableBinaryParser.java:164)
	at jadx.core.xmlgen.ResTableBinaryParser.decodeTableChunk(ResTableBinaryParser.java:111)
	at jadx.core.xmlgen.ResTableBinaryParser.decode(ResTableBinaryParser.java:84)
	at jadx.api.ResourcesLoader.decodeTable(ResourcesLoader.java:173)
	at jadx.core.dex.nodes.RootNode.lambda$loadResources$7(RootNode.java:212)
	at jadx.api.ResourcesLoader.decodeStream(ResourcesLoader.java:113)
	... 5 common frames omitted
INFO  - processing ...
INFO  - done
```

### Jadx version

1.5.1

jpstotz:
The expected offset: 0x1f02f4 is smaller than the actual offset 0x1f0304 - assuming that the index is correct that would mean we would have to jump backwards - which is currently not possible as we are using an InputStream that can only jump forward, not backward. 

[...]

PR #2444, opened 2025-03-14
fix(res): chunk parsing was aborted too early for unordered entries

jpstotz:
The changes introduced in #2344 allow to parser resources that are not ordered.

I forgot to also consider the new behavior in the code that stops reading entries that are located outside of the chunk. 

The old behavior with ordered entries was to stop parsing if we have reached the end of the chunk. Now with unordered entries we can't stop parsing when we reached the end of the chunk because the next entry may have a lower offset and thus may be located again inside the chunk. 

Without this change entries may be ignored even so they are fully valid. In the [sample app](https://www.apkmirror.com/apk/whatsapp-inc/whatsapp/whatsapp-2-24-24-14-release/whatsapp-messenger-2-24-24-14-android-apk-download/) mentioned in #2343 all 15 entries of `resources.arsc/res/values/bools.xml` are now present instead of only 5 without this PR.


skylot (approved):
Nice fix, thank you :+1:

PR #2450, opened 2025-03-28
Non-leaf packages with classes are not visible in the sources tree

### LINKED ISSUE #2449: [gui] Some top-level packages are missing in the TREE VIEW
### Issue details

The package org.apache.tools.ant is not in the tree,  none of classes inside it can be listed. It seems to be the top-level package.
But which classes can be opened by searching.
Image

### Jadx version

r2410.4a6784d & 1.5.1

### Java version

23.0.1

### OS

- [x] Windows
- [ ] Linux
- [ ] macOS

jpstotz:
fixes #2449


skylot (approved):

skylot:
@jpstotz thanks :+1:

PR #2460, opened 2025-04-13
fix: erroneous SignatureProcessor resolution of standard non-generic …

### LINKED ISSUE #2459: [core] SignatureProcessor erroneous resolution of type
### Issue details

As the title suggests, this issue is about how SignatureProcessor erroneously resolves types of fields that point to non-generic types and that have a Signature Annotation attached, to the type of the Signature Annotation.

```java

```

```smali
.class public La/B;
.super Ljava/lang/Object;

.field public static L:Lbool;
	.annotation system Ldalvik/annotation/Signature;
        value = {
            "Lint;"
        }
    .end annotation
.end field
```

The code above, since `boolean` is a standard library type that is not a generic type, should decompile to:

```
public class B {

    /* JADX WARN: Field that had annotation with different signature: int */
    public static bool L;
}
```

But it decompiles to:

```
public class B {
    public static int L;
}
```

### Jadx version

dev

nerodesu017:
…types with Annotation Signatures

### Description
This PR aims to fix the issue where SignatureProcessor would erroneously resolve the types of fields whose main type is not a standard generic but which also have a Signature Annotation.

Related to #2459


jadx-core/src/main/java/jadx/core/dex/visitors/SignatureProcessor.java
```diff
@@ -176,6 +180,39 @@ private void parseFieldSignature(FieldNode field) {
...
+				"java.lang.Byte",
+				"java.lang.Long",
+				"java.lang.Short",
+				"java.lang.Number",
+				"java.lang.CharSequence",
+				"java.lang.Double",
+				"java.lang.Float",
+		};
+		for (String genericType : stdNonGenericTypes) {
+			if (genericType.equals(ty)) {
```
jpstotz, line 209: Are you sure this `equals` check works?  `genericType` is a `java.lang.String` and `ty` is an `ArgType`, thus this if expression can never get `true`.

nerodesu017: `.equals` does work, it casts it to String automatically, but I will explicitly call `.toString` just it is clear.

Well, the type can still be a primitive, but in case of primitives, even with annotations, they don't get erroneously resolved, so I just have a check at the beginning and I return `false`

nerodesu017: I have also modified the test so that our smali file also contains a field with a primitive type to showcase primitives are not an issue and should be skipped.

nerodesu017: I have re-read your comment and want to clarify why we have the `isPrimitive()` check at the start and we also check for "primitive" types.

If in our smali file the fields have primitive types but that are expressed using the fully qualified class name (using the `L` prefix), they are not resolved to be Primitive ArgType's, so we still have to check for them.

Example:

```smali
.field public static L:Lint;
```

nerodesu017: You were more than right. I am sorry, but I suck at java. I have managed to solve the issues, re-run the tests locally, and now it should be fine.

skylot:
@nerodesu017 please check my view on the original issue in my comment: https://github.com/skylot/jadx/issues/2459#issuecomment-2800076284

As a result, I think jadx now correctly process types, but confusing class names clash with primitive type names should be fixed in `RenameVisitor` in `checkNames` method:
https://github.com/skylot/jadx/blob/c602b3d967d89cf222bd73bafa58c3e2116fc190/jadx-core/src/main/java/jadx/core/dex/visitors/rename/RenameVisitor.java#L48

nerodesu017:
> @nerodesu017 please check my view on the original issue in my comment: [#2459 (comment)](https://github.com/skylot/jadx/issues/2459#issuecomment-2800076284)
> 
> As a result, I think jadx now correctly process types, but confusing class names clash with primitive type names should be fixed in `RenameVisitor` in `checkNames` method:
> 
> https://github.com/skylot/jadx/blob/c602b3d967d89cf222bd73bafa58c3e2116fc190/jadx-core/src/main/java/jadx/core/dex/visitors/rename/RenameVisitor.java#L48

I am sorry, but I am not at all familiar with the codebase, as I've said above, nor am I that knowledgeable about Java. I wanted to thank both of you for helping me out, and I have read your comment which made it a bit more clear for me. Using `L` makes it so it is an actual class name following it.
Now, I wanted to ask: you've suggested I modify the `checkNames` method from `RenameVisitor`, but it seems that that runs way after the `SignatureProcessor`. I was thinking: Maybe I should modify the parsing of the annotation ArgType and throw an error in case of an invalid identifier name given as the class name?

skylot:
> Maybe I should modify the parsing of the annotation ArgType and throw an error in case of an invalid identifier name given as the class name?

Any identifier is valid in bytecode, because VM don't check names at runtime. And jadx only rename such identifiers and only if options are not forbid this.
Also, right now, jadx renames only known classes (found in app) and that's why types in your examples types not renamed. Main issue here is that renaming unknown or outside classes may break references, and this become unclear what we need to do for such cases.
I checked and jadx **will** rename such classes if they are known, so looks like there is nothing to do.
As a result, I have no idea how to correctly handle such cases, maybe we can just add a warning.
I will try to find a solution, any suggestions are welcome.

Anyway, further discussion is better to continue in original issue and I will close current PR as it not needed anymore.

skylot (approved):

PR #2461, opened 2025-04-14
fix(plugins): Fix plugin loading issue with custom jar plugins

xi0yu:
# Fix plugin loading issue with custom jar plugins

## Problem
When loading plugins from custom jars using `loadFromJar()`, the plugins may be filtered out due to a strict ClassLoader check in `loadFromClsLoader()`. The current implementation only accepts plugins where `p.type().getClassLoader() == classLoader`, which doesn't consider the ClassLoader hierarchy.

## Solution
Added a `isClassLoaderMatch()` method to properly check the ClassLoader hierarchy relationship. This allows plugins to be loaded if their ClassLoader is either:
- The same as the target ClassLoader
- A child ClassLoader of the target ClassLoader

This change maintains the security and duplicate loading prevention of the original implementation while fixing the issue with custom jar plugin loading.

## Changes
- Added `isClassLoaderMatch()` method to check ClassLoader hierarchy
- Modified `loadFromClsLoader()` to use the new matching logic
- No changes to existing plugin loading behavior for built-in plugins

## Testing
Tested with:
- Loading built-in plugins (unchanged behavior)
- Loading custom jar plugins (now works correctly)

### DIFF
--- a/jadx-plugins-tools/src/main/java/jadx/plugins/tools/JadxExternalPluginsLoader.java
+++ b/jadx-plugins-tools/src/main/java/jadx/plugins/tools/JadxExternalPluginsLoader.java
@@ -60,9 +60,24 @@ public class JadxExternalPluginsLoader implements JadxPluginLoader {
 	private void loadFromClsLoader(Map<Class<? extends JadxPlugin>, JadxPlugin> map, ClassLoader classLoader) {
 		ServiceLoader.load(JadxPlugin.class, classLoader)
 				.stream()
-				.filter(p -> p.type().getClassLoader() == classLoader)
+				.filter(p -> isClassLoaderMatch(p.type().getClassLoader(), classLoader))
 				.filter(p -> !map.containsKey(p.type()))
 				.forEach(p -> map.put(p.type(), p.get()));
+	}
+
+	private boolean isClassLoaderMatch(ClassLoader pluginClassLoader, ClassLoader targetClassLoader) {
+		if (pluginClassLoader == targetClassLoader) {
+			return true;
+		}
+		// Check if the plugin's ClassLoader is a child of the target ClassLoader
+		ClassLoader parent = pluginClassLoader;
+		while (parent != null) {
+			if (parent == targetClassLoader) {
+				return true;
+			}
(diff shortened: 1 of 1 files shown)


skylot:
@xi0yu I wonder how you managed to put plugins in another classloader? Do you use some custom code or another plugin system?

Anyway, now method `loadFromClsLoader` used in two places:
1. to load bundled plugins from "current" classloader
2. create a new classloader for every plugin jar

For second case, no other classloaders are expected.
But for first case, we don't need a filter at all. So, maybe it will be better just split this method and remove filtering for first case at all (along with classloader argument in `ServiceLoader.load`). I think this way code will be easier to understand.

xi0yu:
@skylot Thank you very much for your response. I attempted to use the jadx-mcp-plugin jar package in jadx-gui, but found
that when installing plugins through the GUI, the loadFromJar method is executed, which generates a
URLClassLoader. However, this classloader is subsequently filtered out by the loadFromClsLoader method, resulting
in the plugin not being installed correctly. I have attempted to modify the original logic to allow for more
reasonable loading of plugins.

I would like to thank you again for your time and effort in responding to my question. Your expertise and
knowledge are greatly appreciated.

If you believe this approach is incorrect, please let me know what the correct course of action would be.

Please feel free to provide any feedback or suggestions on how to improve the code.

skylot:
>  loadFromJar method is executed, which generates a URLClassLoader. However, this classloader is subsequently filtered out by the loadFromClsLoader method, resulting in the plugin not being installed correctly

I don't understand the issue. Yes, jadx create a new URLClassLoader, load plugin jar into it and collect plugins only from that classloader to get exactly that one plugin. So we don't `filtered out` we `filtered by` and this is a purpose of that code.

Also, I checked https://github.com/mobilehackinglab/jadx-mcp-plugin plugin it installed successfully in both 1.5.1 and dev jadx versions.

So root cause is still unclear for me and it is pointless to make any fixes.

xi0yu:
@skylot Thank you for pointing out the errors in my previous modifications. I have reviewed and corrected them according
to your feedback. I appreciate your patience and guidance throughout this process.

Issue #1735, opened 2022-11-28
[feature] Dark editor theme instance highlight tweak


ghost:
*Describe your idea:*
Just a small update to the syntax highlighting on the dark editor themes, when you click on a variable or something for instance, and it highlights all instances, on the dark themes, particularly Monokai, it's _very_ hard to see all the other highlighted instances, as the background is only just a shade or two lighter grey than the background, would love it to be a bit more obvious like the default theme (white, that highlights other instances of whatever you click on in orange), but not a fan of a light theme as it burns my eye balls out of my head when working in the evening or night.  

Or let me know where the theme is actually stored and I can probably just do it myself. 

Thanks!

PR #2471, opened 2025-05-02
fix(gui): created dynamic theme for code editor

### LINKED ISSUE #1763: [feature] Sync editor and application themes
*Describe your idea:*
Since the jadx-gui uses FlatIJLookAndFeelInfo as the application themes, it is possible to synchronize the app themes and the code editor themes.

FlatIJLookAndFeelInfo(Name, ThemeClass, isDark)

### How I see it

You just need to add 2 new constants
`currentDarkTheme` responsible for storing the values of the current dark theme of the editor
and `currentLightTheme,` respectively
```java
if (flatIJLookAndFeelInfo.isDark()){
    editorTheme = currentDarkTheme;
} else {
    editorTheme = currentLightTheme;
}
```
And update the UI accordingly

I would like to add this myself, but I am an Android developer, and have no experience with Swing and desktop applications at all
I would be glad if it would be added

MrIkso:
This PR add new theme for code editor based on current app theme and dynamic update if theme changed. Maybe fixed  #1763, #1735
Preview: [Googe Drive](https://drive.google.com/file/d/1d0B8ptYq6Rw5URplq7XliomA_SRFKnKi/view?usp=sharing)

https://github.com/user-attachments/assets/4d64b15c-2c90-4825-992c-35f39e55be54

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
+++ b/jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
@@ -102,6 +102,7 @@ public class JadxSettings extends JadxCLIArgs {
 	private boolean showHeapUsageBar = false;
 	private boolean alwaysSelectOpened = false;
 	private boolean useAlternativeFileDialog = false;
+	private boolean useDynamicEditorTheme = false;
 
 	private Map<String, WindowLocation> windowPos = new HashMap<>();
 	private int mainWindowExtendedState = JFrame.NORMAL;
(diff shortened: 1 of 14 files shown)


skylot:
@MrIkso I commit changes to put your dynamic theme in an editor theme combo box. I adjust code so it will be easier to add new themes and also simplify exposing an API for changing editor themes in plugins in the future.

Also, I slightly change your code to use identifiers (most of the text in the code) color from UI theme, this might break some of your intentions, so please check it. And I think we should use less hardcoded colors and use directly or mix more colors from UI theme, this way it will appear much better in my option.

MrIkso:
@skylot nice. I forgot to add default color for annotations in dark theme, add it
scheme.getStyle(Token.ANNOTATION).foreground = Color.decode("#B3AE60");
it from defaut dark color from Intellij Idea. Thanks.

PR #2474, opened 2025-05-03
Added Preview Tab Feature

### LINKED ISSUE #756: [feature] Add an option to have a "preview mode"
Sometimes we only want view a file content quickly and in that cases, on current version of GUI, we need close the tab after do a single click on some node in the tree.

I think that would be nice, if we can have a "preview mode" option. When we do a single click, a tab is opened with the content of the node, but if we click on another node, the content of these tab is replaced by the new one. To open a node content on new tab, we can do that by double-clicking on the node.

[Preview mode on VSC](https://code.visualstudio.com/docs/getstarted/userinterface#_preview-mode)

MrIkso:
This PR adds feature named as "Enable Prewiew Tab" taken from Inteliji Idea. Tab is opened with the content of the node, but if we click on another node, the content of these tab is replaced by the new one and if existing opened node , we a jumped to it.
Close #756

https://github.com/user-attachments/assets/ece963c0-1e2e-4c85-8ced-3070c2fcc097

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
+++ b/jadx-gui/src/main/java/jadx/gui/settings/JadxSettings.java
@@ -101,6 +101,7 @@ public class JadxSettings extends JadxCLIArgs {
 
 	private boolean showHeapUsageBar = false;
 	private boolean alwaysSelectOpened = false;
+	private boolean enablePreviewTab = false;
 	private boolean useAlternativeFileDialog = false;
 	private boolean useDynamicEditorTheme = false;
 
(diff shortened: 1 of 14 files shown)


skylot (approved):

skylot:
@MrIkso looks good, thank you :+1:

Some possible improvements for future :rofl::
- add toolbar icon to switch and view 'preview tab' state
- after reload tab lost a 'preview' state and click on class in tree will open a new preview tab
- 'pin' and 'bookmark' states on preview tab also lost after opening new class
- improve code to reduce `boolean fromTree`, `boolean preview` and other boolean flags in many methods, such code harder to understand. Possible solution is a flags enum or additional class to hold all open tab options.

MrIkso:
@skylot 
> add toolbar icon to switch and view 'preview tab' state

I dont like if toolbar has more buttons. Maybe use hot keys

> 'pin' and 'bookmark' states on preview tab also lost after opening new class

maybe it need (?)

>after reload tab lost a 'preview' state and click on class in tree will open a new preview tab

> improve code to reduce boolean fromTree, boolean preview

these problems are interconnected, I was trying to come up with logic to implement a preview only for opening from the tree, so I quickly wrote it so as not to break the other tab logic, which is somehow strangely implemented

skylot:
> I dont like if toolbar has more buttons. Maybe use hot keys

One additional icon will not hurt.
Also, a new icon is the only way to let users discover this new feature, because no one will check all menus for new entries :rofl:

> maybe it need (?)

Not sure what you mean, but pin and bookmark actions just don't work for preview tabs right now, and this is definitely not good. Especially for bookmark because it lost after open another class :cry:

> tab logic, which is somehow strangely implemented

Well, current approach with listeners of tab events needed because there are many listeners, so I don't know a better solution for now. And 'preview' should be just an additional tab state along with 'pin' and 'bookmark' which passes in these tab events.

Issue #2441, opened 2025-03-10
[core] Missing and invalid throws declarations


nitram84:
### Issue details

Jadx generates a throws declaration for each method annotated with `dalvik.annotation.Throws`. But this annotation is optional an not used at runtime ( https://stackoverflow.com/questions/14062434/throws-and-annotation-for-exception-in-dalvik-bytecode ), therefore can be removed ( https://github.com/facebook/redex/blob/main/config/aggressive.config#L62 ). If throws annotation are missing, the decompiled result has compile errors.

If throws annotation are not used at runtime, there is a possibility for bytecode manipulations. I did some tests: a thrown object can be anything and not necessarily a `Throwable`.

In my opinion jadx has to perform the same checks like a java compiler:
Collect all thrown exceptions for each method,  calculate effective throws declarations, verify all throws annotations:
For each method: Follow the call graph of each method invocation; per method; aggregate all directly thrown exceptions and exceptions thrown by method invocations, merge collected exceptions with existing throws annotations, generate missing annotations and mark method as visited.

Is there an example for a visitor where we are traversing the call graph of each method? What is the best way to mark visited methods? A new `JadxAttr`?

Example:

```smali
###### Class jadx.test.MissingThrowsTest (jadx.test.MissingThrowsTest)
.class public Ljadx/test/MissingThrowsTest;
.super Ljava/lang/Object;
.source "MissingThrowsTest.java"


# direct methods
.method public constructor <init>()V
    .registers 1

    .line 6
    invoke-direct {p0}, Ljava/lang/Object;-><init>()V

    return-void
.end method

.method private exceptionSource()V
    .registers 3

    .line 8
    new-instance v0, Ljava/io/FileNotFoundException;

    const-string v1, ""

    invoke-direct {v0, v1}, Ljava/io/FileNotFoundException;-><init>(Ljava/lang/String;)V

    throw v0
.end method

.method private invalidException()V
    .registers 3
    .annotation system Ldalvik/annotation/Throws;
        value = {
            Ljava/lang/String;
        }
    .end annotation

    .line 36
    new-instance v0, Ljava/io/FileNotFoundException;

    const-string v1, ""

    invoke-direct {v0, v1}, Ljava/io/FileNotFoundException;-><init>(Ljava/lang/String;)V

    throw v0
.end method


# virtual methods
.method public doSomething1(I)V
    .registers 3
    .param p1, "i"    # I

    .line 20
    const/4 v0, 0x1

[...]

PR #2475, opened 2025-05-06
fix: generate missing throws declarations, validate exceptions (#2441)

nitram84:
This PR calculates missing throws declarations and introduces basic validation to fix https://github.com/skylot/jadx/issues/2441.

The algorithm is a bit more complex than a simple call graph traversal: for each method thrown exceptions are calculated and throws-annotations are used as a hint. If throws-annotations are compatible with calculated exceptions, the algorithm takes the exception of the throws-annotations. So in normal cases there won't be any difference.

There are two drawbacks:

 1.) Option  '`no-finally`' is active. In this case jadx generates code like

```
try {
 //...
} catch(Throwable t) {
  throws t;
}
```

The algorithm detects a thrown `Throwable` here, but in this special case where no throws declaration is required. I don't see a way to distinguish former finally clauses from other similar try-catch structures where throws declarations are required. I could only disable the pass for 'no-finally'. What is better unneeded throws declarations or possible decompilation issues?

2.) Unknown class hierarchy
With an unknown class hierarchy (e.g. decompilation of JARs, smali, ...) I can't perform a validation so assume that thrown classes are a Exception but not a `RuntimeException`. In this case unneeded throws declarations might be added too.

Some findings:
- This algorithm operates on full qualified names and signatures, in other words just strings, the common "base" of `ClassNode`/`ClspClass` and `MethodNode`/`ClspMethod`. I'm aware that this types are used to express totally different things, but from a logical perspective they share common concepts:
  - names (full qualified or signatures)
  - classes can have super classes
  - classes can have methods and nested classes 
  - classes are objects, should be a logical `ArgType.ObjectType`

  I could imagine common interfaces for both class types and method types. Some passes could be simplified if they can operate on common interfaces. For now we use even different names to access parents.
- In my unit tests I don't use any modifiers for test classes and methods to prevent sonarqube warnings.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/Consts.java
+++ b/jadx-core/src/main/java/jadx/core/Consts.java
@@ -16,7 +16,9 @@ public class Consts {
 	public static final String CLASS_STRING = "java.lang.String";
 	public static final String CLASS_CLASS = "java.lang.Class";
 	public static final String CLASS_THROWABLE = "java.lang.Throwable";
+	public static final String CLASS_ERROR = "java.lang.Error";
 	public static final String CLASS_EXCEPTION = "java.lang.Exception";
+	public static final String CLASS_RUNTIME_EXCEPTION = "java.lang.RuntimeException";
 	public static final String CLASS_ENUM = "java.lang.Enum";
 
 	public static final String CLASS_STRING_BUILDER = "java.lang.StringBuilder";
--- a/jadx-core/src/main/java/jadx/core/Jadx.java
+++ b/jadx-core/src/main/java/jadx/core/Jadx.java
@@ -35,6 +35,7 @@ import jadx.core.dex.visitors.InitCodeVariables;
 import jadx.core.dex.visitors.InlineMethods;
 import jadx.core.dex.visitors.MarkMethodsForInline;
 import jadx.core.dex.visitors.MethodInvokeVisitor;
+import jadx.core.dex.visitors.MethodThrowsVisitor;
 import jadx.core.dex.visitors.MethodVisitor;
 import jadx.core.dex.visitors.ModVisitor;
 import jadx.core.dex.visitors.MoveInlineVisitor;
@@ -179,6 +180,8 @@ public class Jadx {
 		}
 		passes.add(new ReturnVisitor());
 		passes.add(new CleanRegions());
+
+		passes.add(new MethodThrowsVisitor());
 
 		passes.add(new CodeShrinkVisitor());
 		passes.add(new MethodInvokeVisitor());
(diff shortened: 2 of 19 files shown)


skylot:
@nitram84 thanks! Looks good, although I will need some time to review this PR, couple days I hope :slightly_smiling_face:

> What is better unneeded throws declarations or possible decompilation issues?

Hm, it is still unclear for me why rethrow Throwable from catch don't need `throws`, but `throw new Throwable()` require `throws` statement. Maybe it will be not hard to check if `throw` instruction arg is assigned in catch, I will check this.
Anyway, sure "unneeded throws declarations" is better than decompilation issues.

> the common "base" of ClassNode/ClspClass and MethodNode/ClspMethod

Well there is common [IMethodDetails](https://github.com/skylot/jadx/blob/1efdcd7b10fe2b3984db743a4deec0b472e0e2ce/jadx-core/src/main/java/jadx/core/dex/nodes/IMethodDetails.java#L11) which is implemented by MethodNode and ClspMethod.
But for classes it is harder to made similar, because ClspClass is 'unresolved' and have only `ArgType`, even parents is just array and not a tree like in `ClassNode`.
But sure, we can add common interface for just some common properties.
Also, some common methods already collected in `TypeUtils` and `MethodUtils`. Perhaps `TypeUtils.visitSuperTypes` will be useful for traversing parents in common way.

> classes are objects, should be a logical ArgType.ObjectType

Not sure what you mean, ArgType.ObjectType already here: https://github.com/skylot/jadx/blob/306bc7abc3578ff802107c41022ff7f333d545df/jadx-core/src/main/java/jadx/core/dex/instructions/args/ArgType.java#L224
and can be created using `ArgType.object()` (it will also convert to `.` form, removing `L` and `;`)

skylot:
I reviewed code and have two news:
1. Good one: instead of going through parent classes to check base exception class, we can use `ClspGraph.isImplements` method, it was created just for that and heavily used in type inference. And as soon as ClspGraph merged with app classes, this method can be for all classes. So I refactor your PR to use that method and it works fine. Also, I added a check for rethrow Throwable and it also works fine if arg used directly from catch.
2. Bad one: as I mentioned in https://github.com/skylot/jadx/issues/2441#issuecomment-2711935526 only first level of dependant classes are loaded, so if your instructions traversal method go deeper, method might not have basic blocks loaded and current implementation will just ignore that.
I will try to create a test case for that, maybe it is not so bad like I think, but anyway it will be nice to fix this somehow.

skylot (approved):

skylot:
Ok, I think it is fine to merge this PR. Processing issue is still there, but to resolving it, will require to apply current instruction checks using low-level code scanner, and that might be hard.

PR #2486, opened 2025-05-15
Restore tree state after renaming

### LINKED ISSUE #2485: [gui] Tree fully reflesh after class renamed
### Issue details

I found a problem in the latest build that the class tree collapses when renaming and does not restore state, in release 1.5.1 this problem does not exist. Need to find out on which commit this is broken))

### Jadx version

latest

### Java version

OpenJDK JDK 24.0.1

### OS

- [x] Windows
- [ ] Linux
- [ ] macOS

MrIkso:
This PRs adds restore tree state after renaming (fix #2485)


skylot (approved):
Thanks :smile:

MrIkso:
This is more of a hotfix, it is not known how productive it will be on large projects with a more of nodes in the tree. Maybe you will come up with a better solution

PR #2487, opened 2025-05-15
Added dialog for select methods for generate frida snippet

MrIkso:
This PRs adds the feature to choose which methods to create a snippet for. It also possibly fixes the bug that when there are many methods and one big snippet is generated for all of them, the window may be freeze(?)

image

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/treemodel/JMethod.java
+++ b/jadx-gui/src/main/java/jadx/gui/treemodel/JMethod.java
@@ -1,12 +1,10 @@
 package jadx.gui.treemodel;
 
 import java.util.Comparator;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
 import javax.swing.Icon;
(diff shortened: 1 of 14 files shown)


jpstotz:
Does this PR replace the old way to create snippets? Because it can get pretty difficult selecting the desired methods from a small list in case the APK is obfuscated and all methods are named next to identical. Therefore I would not recommend to possibility that you can directly generate a snippet for the one method you have identified by it's code or references (through the method's context menu). 

IMHO if you want to hook multiple methods `frida-trace` is the better option. And that the code blocks contain redundant code shouldn't be a problem even for developers without much experience.

MrIkso:
>Does this PR replace the old way to create snippets?

No, it only improves #2328

skylot (approved):

MrIkso:
>bulk selection buttons like "Select All" and "Deselect All" maybe useful
Hot keys from keyboard Ctrl+A worked for select/deselect all:)

PR #2489, opened 2025-05-17
fix(gui): update resource.arsc nodes after opened tab with resource ids

### LINKED ISSUE #2488: [gui] Resource from resource.arsc not showing, always loading
### Issue details

I noticed a problem where resources from resources.arsc are not displayed in the tree, always Loading... state, although they were successfully read in the log.

UPD. To repeat, need to double-click on the node resources.arsc

Image
`DEBUG - Resource table parsed: size: 2731, time: 160ms`
Sample: https://github.com/skylot/jadx/files/3066105/Dream11_v3.33.1_apkpure.com.apk.zip

### Jadx version

latest(be1c024)

### Java version

OpenJDK JDK 24.0.1

### OS

- [x] Windows
- [ ] Linux
- [ ] macOS

MrIkso:
fix #2488

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/treemodel/JResource.java
+++ b/jadx-gui/src/main/java/jadx/gui/treemodel/JResource.java
@@ -200,6 +200,7 @@ public class JResource extends JLoadableNode {
 			List<JResource> nodes = ResTableHelper.buildTree(rc);
 			sortResNodes(nodes);
 			subNodes = nodes;
+			update();
 			return codeInfo;
 		}
 		// single node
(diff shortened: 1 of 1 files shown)


jadx-gui/src/main/java/jadx/gui/treemodel/JResource.java
```diff
@@ -200,6 +200,7 @@ private ICodeInfo loadContent() {
 			List<JResource> nodes = ResTableHelper.buildTree(rc);
 			sortResNodes(nodes);
 			subNodes = nodes;
+			update();
```
skylot, line 203:
```suggestion
			UiUtils.uiRun(this::update);
```
`loadContent` method might be called from a background thread.

skylot (approved):
Thanks. Update call here indeed fixed the issue, but looks like instead of calling load or loadTask, getCodeInfo() method was called directly from codepanel constructor. This is not correct behavior, also now hex view loaded even not visible which is unnecessary. I will look into these issues and will try to fix them.

PR #2490, opened 2025-05-17
Notify user if font not display some characters

### LINKED ISSUE #621: Notify user if the used font doesn't not display some characters
As hinted in https://github.com/skylot/jadx/issues/584#issuecomment-486041365.

When seeing some strange characters in the GUI, I would not know it is my fonts issue.

It would be much better if we display a notification, that some characters are not displayable by the currently used font.

MrIkso:
This PRs add warning message, if code has non-displayable character.  Fix #621

image


skylot:
@MrIkso this check is great, but it might be very annoying to see it every time a project loaded or reloaded. Also, if user will go to settings and change editor font, warning may appear again, because it is unclear which characters not supported and what font support them.
Maybe we can collect all problematic characters and show them in editor within warning window, and add a font selection button to allow users to verify that a selected font is better. Or maybe automatically check all available fonts and show only better ones :rofl:

MrIkso:
@skylot
> but it might be very annoying to see it every time

This check only works when deobfuscation is not used and renaming is disabled. 

>Maybe we can collect all problematic characters and show them in editor

I don't know, there are characters that no font can display. Also I don't know how long it would take to fully check the code, if it very bigger with all non-printing characters. As part of the optimization it only goes through once and if at least one non-printing character is found, everything stops and a warning is displayed.
The question is quite interesting.

MrIkso:
>but it might be very annoying to see it every time a project loaded or reloaded
>Maybe we can collect all problematic characters and show them in editor within warning window, and add a font selection button to allow users to verify that a selected font is better

@skylot some improves

https://github.com/user-attachments/assets/390b0742-cbee-4317-8eb3-9272ce0e9db1

skylot (approved):
Looks great :+1:

Issue #1675, opened 2022-09-19
[core] Unproper decoding of various attributes in resource files


JohnFarl:
In the decompiled xml files are present attributes that aren't decoded correctly mismatching the types

Some attributes are decoded like strings without proper references reconstruction, leading to errors like` * is incompatible with attribute type (attr) integer` because it is expected a reference to an integer rather than a string.

E.g. 
- x8_progress_load.xml (line 5)

JADX 1.4.4
`<gradient android:startColor="#00ffffff" android:endColor="#ffffff" android:useLevel="false" android:type="sweep" android:centerY="0.5"/>`

I get  _sweep is incompatible with attribute type (attr) integer_

JEB in the same line leaves directly an integer
`<gradient android:startColor="#00ffffff" android:endColor="#ffffff" android:useLevel="false" android:type="2" android:centerY="0.5"/>`

Another analogous result comparison where tis time the error is  _"radial" is incompatible with attribute type (attr) integer_

- x8_map_marker_with_pio_view.xml (line 2)

JADX 1.4.4 
`<com.fimi.app.x8s.widget.X8MapPointMarkerViewGroup xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto" android:id="@+id/myview" android:layout_width="wrap_content" android:layout_height="wrap_content" app:type="radial" app:margin2="1dp" app:margin3="1dp">`

JEB
`<com.fimi.app.x8s.widget.X8MapPointMarkerViewGroup xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto" android:id="@id/myview" android:layout_width="-2" android:layout_height="-2" com.fimi.android.app:type="1" com.fimi.android.app:margin2="1.0dp" com.fimi.android.app:margin3="1.0dp">`

The result of JEB is not the best solution and also not desirable to understand the code semantic but it respects the correct type.

In the same apk are encountered my error of this kind with types as
```
sweep              
radial
surfaceView
textureView
```

The apk I have tried is [link](https://apkpure.com/it/fimi-navi/com.fimi.android.app/download/20-APK?from=versions%2Fversion) that despite isn't obduscated is giving various throubles to correct recontruction by JADX decompiler.

jackwpa:
JEB (latest) and jadx (latest) return what seems correct to me, that is

(JEB 4.19 below)
```
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #2492, opened 2025-05-18
Fix decode xml attributes value

MrIkso:
Currenty xml attribute value decoding not checks attibute namespace. Attribute like `android:style` and `app:style` it is different attributes, but decode function get only name `style` and trying decode and migt return wrong result. This PR fix it, issue #1675

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
@@ -51,6 +51,10 @@ public class ManifestAttributes {
 			return values;
 		}
 
+		public void addValue(long key, String value) {
+			values.put(key, value);
+		}
+
 		@Override
 		public String toString() {
 			return "[" + type + ", " + values + ']';
@@ -58,6 +62,12 @@ public class ManifestAttributes {
 	}
 
 	private final IJadxSecurity security;
+
+	/**
+	 * Map containing default Android resource attribute definitions.
+	 * Keys are Android attribute names (e.g., "android:layout_width"),
+	 * and values are their corresponding {@link MAttr} objects.
+	 */
 	private final Map<String, MAttr> attrMap = new HashMap<>();
 	private final Map<String, MAttr> appAttrMap = new HashMap<>();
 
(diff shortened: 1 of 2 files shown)


jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
```diff
@@ -135,7 +135,7 @@ private void parseValues(String name, NodeList nodeList) {
 					if (attr == null) {
 						return;
 					}
-					attrMap.put(name, attr);
+					attrMap.put(String.format("android:%s", name), attr);
```
jpstotz, line 138: For performance reasons I would prefer simple string concat here `"android:" + name´. 
For Maps with `String` as key as `attrMap` I am a fan of having a short JavaDoc where the Map is defined where the String key format is described. Could you please add such a JavaDoc comment?

jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
```diff
@@ -164,6 +164,9 @@ private void parseValues(String name, NodeList nodeList) {
 	public String decode(String attrName, long value) {
 		MAttr attr = attrMap.get(attrName);
 		if (attr == null) {
+			if (attrName.contains(":")) {
+				attrName = attrName.split(":")[1];
```
jpstotz, line 168: I would use `attrName.split(":",2)[1];` 

To my knowledge the attribute name after the namespace definition can not contain colons so your code wouldn't make problems, so again this change would only improve execution performance.

jpstotz (changes requested):

jadx-core/src/main/java/jadx/core/xmlgen/ManifestAttributes.java
MrIkso: For serach we use full or short name attr like this code
https://github.com/skylot/jadx/blob/ea2d54c36e6ea806d968ab7e89079b16198e6971/jadx-core/src/main/java/jadx/core/xmlgen/BinaryXMLParser.java#L304-L305

 >To my knowledge the attribute name after the namespace definition can not contain colons
 
App package attributes might contains a `app:someArrt`  or `someAttrt` so me like `package` in AndroidManifest, so this check is needed to avoid an IndexOutOfBoundsException.

jpstotz: Sorry the check was out of scope, my comment was just only about the split call.

jpstotz (approved):

MrIkso:
Hmm, interesting why building fail

skylot (approved):

skylot:
@MrIkso and @jpstotz thanks :+1:

PR #2494, opened 2025-05-19
Update NLSAddNewLines tool

MrIkso:
This PRs adds:

- [x] Update commneted untanslated lines, if it changed in Messages_en_US.properties 
- [x] Delete all keys in other translation, if it deleted in Messages_en_US.properties

This is a good useful changes and solves when it used to have to be make it manually.


skylot (approved):
@MrIkso thanks!

I renamed class and task, because it not only add lines now :tada:

MrIkso:
> @MrIkso thanks!
> 
> I renamed class and task, because it not only add lines now 🎉

Nice, I didn't even think about it.

Issue #584, opened 2019-04-10
Wrong Characters on DexGuarded Classes Names


enovella:
# Problem
![jadx-bug](https://user-images.githubusercontent.com/14809754/55917717-50f0a800-5be8-11e9-853a-0d97e52ef6a2.png)

Also deobfuscation of methods and fields is applied but it wasn't enabled on preferences
```java
package o;

/* renamed from: o.ᴻ */
public final /* synthetic */ class C implements Qn {
    /* renamed from: ˊ */
    private final HttpInterface f19353;
    /* renamed from: ˏ */
    private final DownloadPdfRequest f19354;

    public C(HttpInterface httpInterface, DownloadPdfRequest downloadPdfRequest) {
        this.f19353 = httpInterface;
        this.f19354 = downloadPdfRequest;
    }

    public Object apply(Object obj) {
        return LeagueListingFeature.m29926(this.f19353, this.f19354, (PermissionEnum) obj);
    }
}
```

# Jadx version
Latest unstable

Comments contain the right symbol:
```java
/* renamed from: o.Ч */
public class C extends C {
    /* renamed from: ˋ */
```
 # Sample
- https://apkpure.com/dream11/com.app.dream11Pro/download?from=details
- [Dream11_v3.33.1_apkpure.com.apk.zip](https://github.com/skylot/jadx/files/3066105/Dream11_v3.33.1_apkpure.com.apk.zip)

asashour:
I couldn't understand where are wrong characters, I assume you mean all classes are named `C`.

Well, this happens only in case sensitive systems (e.g. *nix), without enabling deobfuscation. Once you enable deobfuscation, the files are renamed.

In case insensitive systems (e.g. Windows), without deobfuscation, the files are changed to Cxyz.

I guess this is the intended behavior. Please also look into #570, with its pending PR.

Regarding the fields, you are right, I guess they shouldn't be changed, since deobfuscation is disabled.

enovella:
Sorry but i didnt have the time to explain in details. I dont want to enable de-obfuscation because otherwise I cannot reverse the app properly. Having all classes with names `C` is hard.

Will check this pending PR. Thanks!

skylot:
@asashour actually this is a bug. In this case class name have only not "printable" chars so jadx remove them. And as soon as empty name is not correct `C` char was added )) 
But checks for unique names was incorrecly skipped (or it was performed before chars removing).
So this need to be fixed anyway.

asashour:
[...]

Issue #2304, opened 2024-10-12
[feature] Plugins API: allow traverse input files without opening (and decompressing) ZIP files every time


skylot:
### Describe your idea

Now [`JadxCodeInput`](https://github.com/skylot/jadx/blob/9417671606d25959c5bb97254952f2f411282502/jadx-plugins/jadx-input-api/src/main/java/jadx/api/plugins/input/JadxCodeInput.java#L6) interface force every plugin to search code in all input files and in most cases it is apk or jar files, so every plugin parsing and decompressing same zip file, this is very resource consuming and slow down processing.

Solution is to provide tree structured representation of input files with already parsed zip files.
Files content loading should be done in lazy fashion.

Also, as many files maybe encoded or transformed, it will be nice if plugins can replace or patch this input tree structure.
Such replace will reduce common processing code and allow implementing plugins to support various archive and containers format.

Related to #2302.

PR #2495, opened 2025-05-19
Added feature to use headers for detect resource extensions

### LINKED ISSUE #1643: [Resource] decode 9-patch png image extension inconsistently
I use jadx 1.4.3, I found some .9.png files (if you use unzip tool to unzip the apk file), will be shown as a png file in jadx-gui left side, or use jadx API unzip to disk. But if you click that png file on the left side of jadx-gui, the right side of jadx-gui title will show .9.png file.

It may be that file is not .9.png file, but I think they should be shown the same file name as the original since META-INF/CERT.SF and META-INF/MANIFEST.MF all use .9.png as an extension.

MrIkso:
Currently jadx get resource extension from name. But if names hard obfuscated and don't have extensions. This PRs slove this problem for detect some popular resources used on apk. Also closes #1643 with correct detection of .9.png.
Example apk how to test: https://github.com/skylot/jadx/issues/584

### DIFF
--- a/jadx-core/src/main/java/jadx/api/ResourceFile.java
+++ b/jadx-core/src/main/java/jadx/api/ResourceFile.java
@@ -4,6 +4,9 @@ import java.io.File;
 
 import org.jetbrains.annotations.Nullable;
 
+import jadx.core.deobf.FileTypeDetector;
+import jadx.core.utils.StringUtils;
+import jadx.core.utils.exceptions.JadxException;
 import jadx.core.xmlgen.ResContainer;
 import jadx.core.xmlgen.entry.ResourceEntry;
 import jadx.zip.IZipEntry;
@@ -11,7 +14,7 @@ import jadx.zip.IZipEntry;
 public class ResourceFile {
 	private final JadxDecompiler decompiler;
 	private final String name;
-	private final ResourceType type;
+	private ResourceType type;
 
 	private @Nullable IZipEntry zipEntry;
 	private String deobfName;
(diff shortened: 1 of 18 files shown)


jadx-core/src/main/java/jadx/api/ResourceFile.java
```diff
@@ -53,22 +56,48 @@ public ResContainer loadContent() {
...
-		sb.append("/").append(ri.getKeyName());
-		int lastDot = name.lastIndexOf('.');
-		if (lastDot != -1) {
-			sb.append(name.substring(lastDot));
+		sb.append("res/").append(entry.getTypeName()).append(entry.getConfig());
+		sb.append("/").append(entry.getKeyName());
+
+		if (useHeders) {
+			try {
+				byte[] bytes = ResourcesLoader.decodeStream(this, (size, is) -> is.readAllBytes());
```
jpstotz, line 66: Considering that resources can get pretty large I would recommend not to perform file detection on the full content. As far as I know most file type detectors limit their detection on the start of the file, e.g. the first 4K.

jadx-core/src/main/java/jadx/core/deobf/FileTypeDetector.java
```diff
@@ -0,0 +1,128 @@
...
+
+	public static String detectFileExtension(byte[] data) {
+		// detect ext by headers
+		String extByHeaders = detectByHeaders(data);
+		if (!StringUtils.isEmpty(extByHeaders)) {
+			return extByHeaders;
+		}
+
+		// detect ext by readable text
+		String text = new String(data);
```
jpstotz, line 67: please specify the charset (e.g. `StandardCharsets.UTF_8`) so that detection will work deterministic and OS independent

MrIkso: Well, this part can and will be removed if partial file reading is implemented.

skylot (approved):

skylot:
@MrIkso to be honest I don't like a additional decoding run, but since you made this feature opt-in and it also runs only if resource deobfuscation is applying alias, it is fine for now. I hope this will be resolved once a `virtual file tree` feature from #2304 is implemented, and it will be possible just rename already decoded resources :slightly_smiling_face:

skylot:
Oh, sure, instead `decoding` I mean `unzipping` since `ResourcesLoader.decodeStream(` will unzip resource entry if it inside apk.

PR #2498, opened 2025-05-20
feat(gui): A a tree structure usage search information

17307:
When I use jadx, I really like its usage search feature. However, I encountered an issue while using this feature: it can only display the usage information of one function at a time.  

In order to obtain usage information layer by layer, I have to open many usage dialogs, which is very confusing. To solve this, I have added a dialog that displays usage information in a tree structure.

image

image

Right click a node supports jumping to the specified location and copying the call path from the initial node to the current node. When left clicking a node, the code at the current node will be rendered.

image


skylot (approved):
@17307 looks good. Thank you :+1:

Although, I think, we can use an actual tree (`JTree` with lazy loading) to represent usage on deeper levels, instead of opening a new list. This should reduce UI clutter and simplify navigation, but, as always, this is a suggestion for future improvement :slightly_smiling_face:

PR #2499, opened 2025-05-20
fix(res): fix reading resource data

MrIkso:
fix for https://github.com/skylot/jadx/pull/2495#issuecomment-2895756552

### DIFF
--- a/jadx-core/src/main/java/jadx/api/ResourceFile.java
+++ b/jadx-core/src/main/java/jadx/api/ResourceFile.java
@@ -1,12 +1,13 @@
 package jadx.api;
 
 import java.io.File;
+import java.io.IOException;
+import java.util.Objects;
 
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


skylot:
You are too fast :rofl:
This change don't fix the root issue I mention in my comment: jadx now defaults to own parser and `getInputStream()` method on entry actually reads all bytes for that entry and wraps into ByteArrayInputStream:
https://github.com/skylot/jadx/blob/d84f0389ec5356942ce07219799152ecae842b27/jadx-commons/jadx-zip/src/main/java/jadx/zip/parser/JadxZipParser.java#L291

So it doesn't help much trying to read only the start of the entry, because input stream already contains all data.
And to fix this, `JadxZipParser` should implement `getInputStream()` method properly, it is possible, but require carefully check all zip entry fields and use same logic as in `getBytes(` method.

Another approach is to store unzipped entries in a virtual file tree, as suggested in #2304

MrIkso:
Thank you, I now really understand what problem is.

Issue #1590, opened 2022-07-27
[feature] Allow renaming into a different package


NebelNidas:
It's possible to rename classes in JADX, but you can't alter the package they're in. While you can rename already present packages, it's not possible to move individual classes into different (possibly made-up) packages.

I suggest doing it like [Enigma](https://github.com/FabricMC/Enigma): Renaming an imaginary `test.class` to `my/package/test.class` creates the following package structure:
```
├──my/
│   ├── package/
│   │   ├── test.class
```

This is useful if you want to restore heavily obfuscated apps' original structure, or just organize the classes for easier deobfuscation. The YouTube app for example contains almost 30'000 classes, and save for a few hundred they're all top-level without a package.

skylot:
Sure, this issue similar to #1253 and I am already working on package renaming refactor in `jadx-script` branch.

NebelNidas:
Perfect :) I don't know how you plan to implement it, but I'd suggest using Enigma's way of allowing the user to just add slashes into the rename window and resolve each slash-succeeded part as a package. Maybe also provide a GUI option, but please handle it internally by using slashes inside the classes' `alias` name. That should make it instantly compatible with the already present mapping import/export code

dvdface:
> Sure, this issue similar to #1253 and I am already working on package renaming refactor in `jadx-script` branch.

Have you finished it?

skylot:
@dvdface this feature still not merged (now in `jadx-next` branch). But it will be very helpful if you check it and report any issues.

skylot:
This feature now in master, so it is possible to enter new full class name to move class into new package.
Although, there is known issue with missing package tree updates, as workaround you can use project reload (toolbar button of F5).

ed9w2in6:
I agree that this is needed for practical work.
I too frequently observe the existence of a "catch-all" package when decompiling, which holds for multiple different apps.
Most importantly the package usually includes Kotlin Intrinsics, which is often a useful source of meta data.

Since `jadx`'s currently does not allow restructuring:
a. it can get quite messy :: minor impact
b. the Kotlin meta data renaming feature fails to kick in BUT no way to manually enable it again :: potentially huge impact

Regarding b), if this is merged just by renaming Intrinsics classes to their respective locations accordingly a potentially lot of variables can be automatically renamed.  I use potentially as we know the Kotlin meta data might be disabled or obfuscated too.

skylot:
[...]

PR #2500, opened 2025-05-20
fix(gui): fix rename package behaviors

MrIkso:
Related to https://github.com/skylot/jadx/issues/1590#issuecomment-2895788206

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/dialog/RenameDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/dialog/RenameDialog.java
@@ -69,12 +69,7 @@ public class RenameDialog extends CommonDialog {
 	}
 
 	private void initRenameField() {
-		if (node instanceof JRenamePackage) {
-			JRenamePackage renamePackage = (JRenamePackage) node;
-			renameField.setText(renamePackage.getTitle());
-		} else {
-			renameField.setText(node.getName());
-		}
+		renameField.setText(node.getName());
 		renameField.selectAll();
 	}
 
--- a/jadx-gui/src/main/java/jadx/gui/ui/dialog/RenameDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/dialog/RenameDialog.java
@@ -69,7 +69,12 @@ public class RenameDialog extends CommonDialog {
 	}
 
 	private void initRenameField() {
-		renameField.setText(node.getName());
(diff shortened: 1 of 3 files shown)


jadx-gui/src/main/java/jadx/gui/ui/dialog/RenameDialog.java
```diff
@@ -69,7 +69,12 @@ private RenameDialog(MainWindow mainWindow, JRenameNode node) {
 	}
 
 	private void initRenameField() {
-		renameField.setText(node.getName());
+		if (node instanceof JRenamePackage) {
+			JRenamePackage renamePackage = (JRenamePackage) node;
+			renameField.setText(renamePackage.getTitle());
```
skylot, line 74: Well, if you want to see full name instead short name in editor box, you can just change "name" to "fullName" in `JRenamePackage.getName()` method.
But, short name used here intentionally, because detecting changed part in full name is much harder. It is unclear if the middle package name was renamed or rest moved into a new package. Also, packages count can change and that complicate things even more.
I am actually implemented a full package rename, but it is just safer and easier to rename only one part.

I will also look into this issue, maybe I will be able to find a root cause.

skylot (changes requested):

jadx-gui/src/main/java/jadx/gui/ui/dialog/RenameDialog.java
MrIkso: > I will also look into this issue, maybe I will be able to find a root cause.

I thik here root problem
https://github.com/skylot/jadx/blob/a80eefaeea3be4eeb5e1ff274549a5c058bfbcd3/jadx-gui/src/main/java/jadx/gui/utils/pkgs/PackageHelper.java#L64

skylot (approved):

skylot:
I checked current code, and have only bad news :rofl: :
1. current renames serialization approach doesn't support synthetic and default packages renames
2. packages tree still not refreshed after moving class (or package) into different package

Here "synthetic package" is a package created by user during moving a class to a new package, such package doesn't have a “reference” so it can't be saved with current approach. It is possible to use other nodes as references, I will create a new issue for that. For now, I just disable rename for such packages :sob:

Second issue: while your change is fixed tree refresh for simple renames, looks like structural changes to tree still not applied after moving class to different package (I suspect some caching issue). So original #1590 issue still not fixed.

PR #2503, opened 2025-05-22
feat(gui): A a tree structure usage search information - JTree Render

17307:
I tried your [suggestion](https://github.com/skylot/jadx/pull/2498) and replaced the list with a Jtree. The effect feels really good. Now there is more space to display the data.

image

The highlighted part represents the currently selected level.


skylot (approved):

skylot:
@17307 great work! Thank you :+1:

Although, I notice that you hardcoded highlighting color, so text become unreadable in dark themes. It is better to use (or mix in) color from current editor theme, check [DynamicCodeAreaTheme](https://github.com/skylot/jadx/blob/73348e51834ed15cb967222855205a1c829549f5/jadx-gui/src/main/java/jadx/gui/ui/codearea/theme/DynamicCodeAreaTheme.java#L31) class for some hints, like detecting dark theme and getting color from theme (`UIManager.getColor("EditorPane.selectionBackground")`).

To check all current colors (will open FlatLaf UI Inspector window):
- set `JADX_GUI_DEBUG=true` environment variable
- run jadx-gui
- press `ctrl shift alt Y`

Added in jadx by this code:
https://github.com/skylot/jadx/blob/aee1e863982f6f9393be656bf68505a361281bdc/jadx-gui/src/main/java/jadx/gui/ui/MainWindow.java#L1507-L1510

PR #2505, opened 2025-05-23
Option to disable the tooltip that pops up on hover

MewtR:
This PR adds an option in the Preferences menu to disable the tooltip that pops up 
when hovering over a node.
I thought it would be nice to have because the tooltip can steal focus
away from the mouse. Anyways, let me know  what you think.


skylot (approved):
@MewtR nice! Thanks :+1:

PR #2509, opened 2025-05-24
Provide direct InputStream of ZIP entries

jpstotz:
As mentioned recently there was no direct way of getting an InputStream of a ZIP entry without having to extract the full entry data to a byte array in memory. 

This PR provides a way to directly get an InputStream for compressed and uncompressed data in a ZIP file.
Only in one case the only way of reading everything to memory is still present: when using the FallbackZipParser.


skylot:
@jpstotz thanks for implementation :+1:
Although I apply some changes:
- use `LimitedInputStream` if security settings suggest so
- for `getBytes` method use previous implementation without converting to input stream, additional conversion looks unnecessary.

skylot (approved):

PR #2511, opened 2025-05-24
fix(gui): migrate from old unsupported rxjava2 to rxjava3

MrIkso:
This PRs added migration from old RxJava2 to new RxJava3


MrIkso:
@skylot strangely why all contributor to jadx gui didn't using all reactive programming functions, so it could be made faster and interface wouldn't block at some operations). Such a powerful library has been added a long time ago and only a few classes and few operations are performed on it.

skylot:
@MrIkso from my experience while it looks easy to use, it also very easy to run something in wrong thread :rofl:
Also, sometimes it feels overcomplicated and hard to understand, so I stop using it :slightly_frowning_face:

PR #2513, opened 2025-05-24
Use list component witch recent projects on start page

MrIkso:
Simple improved start page. Use list witch recent projects instead of using a button as item added to the scrollview 😄. 

<details>
  <summary>Preview</summary>

image

</details>


skylot:
@MrIkso thanks, nice work!
I moved all classes into one package, it feels better to have everything related in one place.
Also, I removed project extension from project name. Jadx project actually stored project name in `JadxProject.name` but now it calculated from project file path and can't be changed.
And allowing to change project name seem like a good idea :rofl:

PR #2521, opened 2025-06-02
fix(gui): fixed incorrectly handles full window repaint in search dialog

### LINKED ISSUE #2520: [gui] Empty space or options disapper in search dialog
### Issue details

After maximizing search dialog, empty space appear below search options
(from https://github.com/skylot/jadx/pull/2510#issuecomment-2923479756 thanks to @MrIkso):

https://github.com/user-attachments/assets/3e26406a-ff1a-4390-a1f1-45a8bf65447a

Looks like a custom wrap flow layout incorrectly handles full window repaint.

### Jadx version

1.5.2

### Java version

*

### OS

- [x] Windows
- [x] Linux
- [x] macOS

MrIkso:
Workaround for #2520

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/dialog/SearchDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/dialog/SearchDialog.java
@@ -1,7 +1,6 @@
 package jadx.gui.ui.dialog;
 
 import java.awt.BorderLayout;
-import java.awt.Component;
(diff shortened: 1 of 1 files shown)


skylot (approved):
Well, this is a very brute force workaround, and not a "fix".
And will cause a double repaint on all resize :slightly_frowning_face:
So this is acceptable only as a temporal solution.

Issue #2346, opened 2024-11-19
[gui] The class definitions in the JavaScript code generated by the "copy Frida snippet" (of Java class) are duplicated.


poping520:
### Issue details

![image](https://github.com/user-attachments/assets/604f49bd-224f-4d90-b08c-526c4ef37c88)

### Jadx version

1.5.1

### Java version

23.0.1

### OS

- [ ] Windows
- [ ] Linux
- [ ] macOS

jpstotz:
Verified. That is the new functionality, introduced by PR #2328.

PR #2525, opened 2025-06-03
Remove duplicate class of frida snippet

RuffaloLavoisier:
Remove duplicate class of frida snippet.
See #2328, #2487, #2346


skylot (approved):
Thanks :+1:

Issue #2522, opened 2025-06-02
[gui] Now the search results include classes.dex and search for text in it


MrIkso:
### Issue details

![Image](https://github.com/user-attachments/assets/ddf699d4-8479-4ff1-9311-4d0a52cbb200)

This didn't happen before, maybe it was caused by a commit https://github.com/skylot/jadx/commit/e51a7fe417eb6bf4e50822bdc310dea8596eb208 with removed `CODE` from `public static boolean isSupportedForView(ResourceType type) `

### Jadx version

1.5.2

### Java version

OpenJDK JDK 24.0.1

### OS

- [x] Windows
- [ ] Linux
- [ ] macOS

skylot:
oops, definitely not expected this :facepalm:

MrIkso:
Set a separate resource extension without the * fixed it, but clicking on classeses.dex will crash.

```Jadx version : dev
Java version : 24.0.1
Java VM      : Oracle Corporation OpenJDK 64-Bit Server VM
Platform     : Windows 11 (10.0 amd64)
Max heap size: 11264 MB
Program args : -Dawt.useSystemAAFontSettings=lcd -Djava.util.Arrays.useLegacyMergeSort=true -Djdk.util.zip.disableZip64ExtraFieldValidation=true -Dsun.java2d.d3d=false -Dsun.java2d.ddblit=false -Dsun.java2d.ddforcevram=true -Dsun.java2d.noddraw=true -Dswing.aatext=true -Dswing.useflipBufferStrategy=true -XX:MaxRAMPercentage=70.0 -XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.lang=ALL-UNNAMED --enable-native-access=ALL-UNNAMED -XX:MaxRAMPercentage=70.0 -XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.lang=ALL-UNNAMED --enable-native-access=ALL-UNNAMED -Xms128M -Dfile.encoding=UTF-8 -Duser.country=UA -Duser.language=uk -Duser.variant

java.lang.NullPointerException: Cannot invoke "jadx.gui.ui.codearea.AbstractCodeArea.requestFocus()" because "codeArea" is null
	at jadx.gui.ui.tab.TabbedPane.scrollToPos(TabbedPane.java:243)
	at jadx.gui.ui.tab.TabbedPane.showCode(TabbedPane.java:231)
[...]

PR #2526, opened 2025-06-03
New search options to search in text or binary resources

skylot:
New checkboxes to add a `variable` into resource file filter field.
`$TEXT` for text files, `$BIN` for binary. Content type field added to `ResourceType` enum.
Filter field can be edited manually, checkboxes will add/remove variables without resetting current values.
`$TEXT` now a default filter string for resources.

screenshot_2025-06-04-004442

Also, some improvement and fixes done to allow search in binary files:
- scroll to position in hex viewer
- using one byte charset to binary files to match offsets in byte array (and hex viewer)

Resolves issues from #2522

### DIFF
--- a/jadx-core/src/main/java/jadx/api/ResourceType.java
+++ b/jadx-core/src/main/java/jadx/api/ResourceType.java
@@ -4,29 +4,40 @@ import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 
+import jadx.api.resources.ResourceContentType;
 import jadx.core.utils.exceptions.JadxRuntimeException;
 
-public enum ResourceType {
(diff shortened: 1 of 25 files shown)


MrIkso:
@skylot im found bug - scrollToOffset work fine if already opened hexviewer

skylot:
@MrIkso thanks, added selection of matching tab (by file content type) and data loading before scroll to pos.
Also, I start treating unknown file types as binary in search, so now `$TEXT` is only for known types (hardcoded in `ResourceType` class) and `$BIN` for all others. Previously unknown files were treated like text files.

MrIkso:
@skylot nice, bug fixed

PR #2527, opened 2025-06-05
Change let to var in Frida script

TheDauntless:
### Description

When generating Frida hooks, the generated scripts use 'let' for the class reference. Since you're typically prototyping different hooks and copy/pasting the hooks into your script, these let statements clash. Using `var` instead of `let` for the class definition fixes this.

Alternatively we could wrap the generated code in a code block, but this var approach makes more sense to me.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/action/FridaAction.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/action/FridaAction.java
@@ -136,7 +136,7 @@ public final class FridaAction extends JNodeAction {
 		JavaClass javaClass = jc.getCls();
 		String rawClassName = StringEscapeUtils.escapeEcmaScript(javaClass.getRawName());
 		String shortClassName = javaClass.getName();
-		return String.format("let %s = Java.use(\"%s\");", shortClassName, rawClassName);
+		return String.format("var %s = Java.use(\"%s\");", shortClassName, rawClassName);
 	}
 
 	private void showMethodSelectionDialog(JClass jc) {


jpstotz:
Some days ago this problem was already addressed by PR https://github.com/skylot/jadx/pull/2525. It removed the duplicate class declarations.

TheDauntless:
@jpstotz  I saw that, but it only removes duplicates when you export multiple functions at the same time.

With `var` you can just keep generating new snippets and pasting them into your main frida script without having to constantly delete the class definition line or changing it to `var` yourself.

So scenario:

1. Identify interesting function, generate frida script and paste into hook.js
2. Identify second interesting function based off result of first hook, generate frida script and paste into hook.js
3. => Clash, due to duplicate class definition

EDIT
Alternatively, we could just use Java.use() directly for the hook instead of putting it in a variable. It might have a _tiny_ performance impact, but would make it work in all cases:

```javascript
Java.use("com.google.android.material.appbar.AppBarLayout$BaseBehavior$a$a")["createFromParcel"].overload('android.os.Parcel', 'java.lang.ClassLoader').implementation = function (parcel, classLoader) {
    console.log(`C0049a.createFromParcel is called: parcel=${parcel}, classLoader=${classLoader}`);
    let result = this["createFromParcel"](parcel, classLoader);
    console.log(`C0049a.createFromParcel result=${result}`);
    return result;
};
```

RuffaloLavoisier:
Using `var` may seem convenient for copy-pasting multiple snippets, but it introduces several issues that can negatively affect script quality.

- `var` is function-scoped, not block-scoped, which can lead to unexpected variable leaks or overwriting in larger scripts.
- It allows re-declaration, which can silently override existing variables and make debugging harder.

TheDauntless:
I still think var makes sense since the full classname is used as the variable name, so it would be weird if any other value is assigned to it other than a Frida reference to that class.

So the choice is between:

* Generating scripts that break your frida script due to re-definition of let if you use the feature more than once
* Generating scripts that potentially introduce side-effects. I feel like this is not a problem, since **you** are the one choosing to copy code into existing code, and hence the risk is _always_ there
* Generating scripts that never break by not using a variable in the first place
* Generating scripts that never break by wrapping them in a code block

skylot:
Well, I agree, that `let` is better for cleaner and error-prone code, but Frida hooks is just a quick way to get some info, so I think it is fine to use `var` in hooks to simplify adding more methods.
And if such topic will be brought up again, we will use a workaround by inlining class variable as @TheDauntless suggested :rofl:

PR #2543, opened 2025-06-21
refactor(gui): Improve menu icon, such as'Select in Tree' and 'Go to AndroidManifest.xml' buttons

xiaojye:
refactor(gui): Improve menu icon, such as'Select in Tree' and 'Go to AndroidManifest.xml' buttons


zhongqingsong:
cool, Especially for the AndroidManifest icon, the problem of the outer circle of the icon being white and the inner one being dark has been solved. In fact, it seems that the white outer circle is hidden. This leads to a sense of incongruity that, under normal circumstances, the list icon appears to be one circle smaller than other ICONS.

PR #2587, opened 2025-07-28
fix: prevent duplicate Java.use() declarations in Frida class method …

grinwoltwoldorf411:
…snippets

When generating Frida snippets for multiple methods of a class, the code was incorrectly calling generateMethodSnippet() which includes the class declaration, causing Java.use() to be repeated for each method.

Changed generateClassAllMethodSnippet() to call getMethodSnippet() directly, which generates only the method implementation without the class declaration.

This ensures clean output with a single Java.use() statement followed by all method implementations.


jpstotz:
Hi. Thanks for the PR, but I am a bit confused. Is it possible that have developed the patch for an older version of Jadx (e.g. Jadx 1.5.2)? 

Because in the latest Jadx dev version source code the function `generateMethodSnippet` is defiend as:
```
	private String generateMethodSnippet(JavaMethod javaMethod, JClass jc) {
		return getMethodSnippet(javaMethod, jc);
	}
```

And as your PR simply replaces the call of `generateMethodSnippet` with `getMethodSnippet` your PR will have no effect on the generated Frida code snippet at all.

Issue #2588, opened 2025-07-29
[core] Incorrect line endings in CFG.


beaverxsheet:
### Issue details

When generating the CFG using the `--cfg` flag the line endings for the instructions are missing the backslash in front of them and therefore result in a broken graph.

Example Output:
```
...
Node_5 [shape=record,label="{5\:\ 0x0023|lr3.f.addTag(42)landroid.content.Intent r0 = (android.content.Intent) r0.getParcelableExtra(\"nextIntent\")lr3.nextIntent = r0l}"];
...
```
Expected Output:
```
Node_5 [shape=record,label="{5\:\ 0x0023|r3.f.addTag(42)\landroid.content.Intent r0 = (android.content.Intent) r0.getParcelableExtra(\"nextIntent\")\lr3.nextIntent = r0\l}"];
```
This is caused by an incorrect usage of `String.replaceAll()` in `DotGraphVisitor.java`. The replacement string containing a backslash will cause problems.

I have a fix for this and can create a pull request if you like.

### Jadx version

Commit b1f48f1

PR #2592, opened 2025-07-31
fix: use proper newlines when generating cfg from backup

beaverxsheet:
As described in #2588 there is an issue when generating the CFG. This PR fixes that issue by using `Matcher.quoteReplacement()` as suggested in the Java documentation.

I also noticed a bug where there is a missing space between "CFG for" and the function path in the title. I wasn't sure whether I should put that in a separate pull request.

I'm quite new to contributing to OSS, so I apologize if I don't know the proper procedures for all this.


skylot (approved):
Thank you :+1:

PR #2636, opened 2025-09-13
fix: avoid false positive throws (regression of PR #2475)

nitram84:
I'd like to fix a regression I introduced with PR #2475. In two cases a check for excluded exceptions was missing. This may cause compiler errors in overridden methods.

## Steps to reproduce

Recompile the app https://saper-minesweeper.apk.gold/

There are errors in `LionheartGames.Saper.SaperActivity` due to thrown exceptions (false positives).

### Detailed instructions:

Download apk from https://apk.gold/download?file_id=263903/saper-minesweeper

Recompile with:

```sh
jadx -e Saper-1.4.apk 
cd Saper-1.4
# Workaround for R-file conflicts, R files should not be generated in recompilation mode. 
# This should be fixed for a better out-of-the box experience.
mv app/src/main/java/LionheartGames/Saper/R.java app/src/main/java/LionheartGames/Saper/R.java.bak
gradle wrapper --gradle-version 8.13
./gradlew assembleDebug
```

After applying the PR this app can be recompiled and installed without errors. Recompiled game is playable.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/MethodThrowsVisitor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/MethodThrowsVisitor.java
@@ -153,12 +153,12 @@ public class MethodThrowsVisitor extends AbstractVisitor {
 						return;
 					}
 				}
-				visitThrows(mth, exceptionType);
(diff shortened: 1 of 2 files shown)


skylot (approved):
Nice fix. Thank you :+1:

PR #2647, opened 2025-09-29
feat: support .apks file

duzhaokun123:
### Description

support .apks file input


skylot (approved):

skylot:
@duzhaokun123 thanks :+1:

Although, at this point such plugins looks very similar to other 'apk bundles' like 'xapk', 'apkm', etc. So maybe in future it is worth to merge all these plugins into one 'Apk Bundles' plugin.

PR #2654, opened 2025-10-11
[core] fix jadx.plugins.input.java.data.ConstPoolReader.parseString()…

wech71:
### Description
fixes "TODO: parse modified UTF-8" in ConstPoolReader.parseString to follow jvms-4.4.7 rules for encoding annotation strings in class files

(This allows Kotlin Annotation-plugin to work with the Protobuffer byte array encoded as a String in the Kotlin Metadata of Kotlin 2.0 or newer)

(I apologize if this is incorrect now as you said, I should not merge to the master branch, but I could not find or push to a new branch in skylot/jadx. I do not have much experience with github)

### DIFF
--- a/jadx-plugins/jadx-java-input/src/main/java/jadx/plugins/input/java/data/ConstPoolReader.java
+++ b/jadx-plugins/jadx-java-input/src/main/java/jadx/plugins/input/java/data/ConstPoolReader.java
@@ -1,6 +1,5 @@
 package jadx.plugins.input.java.data;
 
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


skylot:
@wech71 thanks :+1:
I move method into other class to simplify test and fixed code style with autoformatter.
Although, I was not able to trigger branch which decodes 6-byte char, looks like my sample chars was splitted into two 3-byte chars. It is not urgent, so we may add a test for that case later.

skylot:
> I apologize if this is incorrect now as you said, I should not merge to the master branch

You do everything right. I was trying to say, that branch in your fork shouldn't be a master. Because on Github master branch is write protected, and I will not be able to add fixes commit, like I do.

wech71:
thanks for accepting my changes :-)

> looks like my sample chars was splitted into two 3-byte chars.

hm, I just had another look at the spec and it seems I made a small mistake. The check for 6-byte encoding according 4.12.  must be done before 3-byte encoding 4.8, because both use the same upper nibble of first byte and the same two bits for byte 2 and 3.

This way the code currently always incorrectly enters 3-byte-encoding.


Either the order of the check for 6-byte and 3-byte encoding has to be changed, or an easy fix woud be to change the line

```
if( (x & 0xF0) == 0xE0 && (y & 0xC0) == 0x80 && (z & 0xC0) == 0x80) {
```

to 

```
if( (x & 0xF0) == 0xE0 && (y & 0xC0) == 0x80 && (z & 0xC0) == 0x80    
     && (x != 0xED || (y & 0xF0) != 0xA0) ) {
```
so the 3-byte -special case will continue to the if 6-byte check.

PR #2655, opened 2025-10-11
jadx-kotling-metadata-plugin: Upgrade to kotlin 2.2 metadata library

wech71:
### Description
Upgraded kotlin metadata library from  kotlinx:kotlinx-metadata-jvm:0.9.0 prerelease to kotlin-metadata-jvm:2.2.20 which can read older and current kotling 2.2 metadata annotations.


skylot:
@wech71 I added a test, but instead your smali file I added a java class file, because we're checking fixed java parser :slightly_smiling_face:

Issue #2139, opened 2024-04-03
Can't encode constant CLASS as encoded value


yooo1999:
### Issue details

When decompile jar with jadx cli in Windows

```
ERROR - Dependency scan failed at insn: 0x0372: INVOKE_CUSTOM in method: io.vavr.CheckedFunction3.$deserializeLambda$(java.lang.invoke.SerializedLambda):java.lang.Object, file: XXXXXX.jar:io/YYY/CheckedFunction3.class
jadx.plugins.input.java.utils.JavaClassParseException: Can't encode constant CLASS as encoded value
        at jadx.plugins.input.java.data.ConstPoolReader.readAsEncodedValue(ConstPoolReader.java:230)
        at jadx.plugins.input.java.data.ConstPoolReader.resolveMethodCallSite(ConstPoolReader.java:117)
        at jadx.plugins.input.java.data.ConstPoolReader.getCallSite(ConstPoolReader.java:97)
        at jadx.plugins.input.java.data.code.JavaInsnData.getIndexAsCallSite(JavaInsnData.java:168)
        at jadx.plugins.input.java.data.code.decoders.InvokeDecoder.decode(InvokeDecoder.java:32)
        at jadx.plugins.input.java.data.code.JavaInsnData.decode(JavaInsnData.java:46)
        at jadx.core.dex.visitors.usage.UsageInfoVisitor.processInsn(UsageInfoVisitor.java:157)
        at jadx.core.dex.visitors.usage.UsageInfoVisitor.lambda$processInstructions$0(UsageInfoVisitor.java:114)
        at jadx.plugins.input.java.data.code.JavaCodeReader.visitInstructions(JavaCodeReader.java:81)
        at jadx.core.dex.visitors.usage.UsageInfoVisitor.processInstructions(UsageInfoVisitor.java:112)
        at jadx.core.dex.visitors.usage.UsageInfoVisitor.processMethod(UsageInfoVisitor.java:97)
        at jadx.core.dex.visitors.usage.UsageInfoVisitor.processClass(UsageInfoVisitor.java:86)
        at jadx.core.dex.visitors.usage.UsageInfoVisitor.buildUsageData(UsageInfoVisitor.java:71)
        at jadx.core.dex.visitors.usage.UsageInfoVisitor.init(UsageInfoVisitor.java:55)
        at jadx.core.dex.nodes.RootNode.runPreDecompileStage(RootNode.java:334)
[...]

PR #2658, opened 2025-10-15
feat: Add support for lambda intersection types in type inference

vicky-dx:
## Summary
This PR completes the type inference support for lambda intersection types that was left as a TODO in #2139.

## Problem
When decompiling lambda expressions with intersection types (e.g., `(TestCls<R> & Memoized) lambda::method`), JADX was:
- Not parsing marker interfaces from altMetafactory bootstrap arguments
- Missing marker interface information during type inference
- Resulting in incorrect or missing casts in decompiled code

The TODO comment in TestLambdaInstance3 documented this: 
> `// TODO: type inference set type for 'of' to Memoized and cast incorrectly removed`

## Solution
This PR implements full marker interface support:

1. **Parse marker interfaces** from altMetafactory args[6-9]:
   - Check flags (arg[6]) for bit 0x02 (markers present)
   - Read marker count (arg[7])
   - Extract marker types (arg[8+])

2. **Store marker interfaces** in `InvokeCustomNode` for use during type inference

3. **Propagate marker types** through type inference, preferring the last marker (typically the additional constraint like `Memoized`) over the functional interface

4. **Preserve marker casts** via new `FixLambdaCastVisitor` that:
   - Updates CHECK_CAST instructions to use marker interfaces
   - Marks them as EXPLICIT_CAST to prevent removal

## Test Results
-  `TestLambdaInstance3.testRaung()` - PASSES showing correct `(Memoized)` cast
-  `TestLambdaInstance3.test()` - PASSES with documented limitation
-  All existing lambda-related tests pass (no regressions)

### Before (main branch):
```java
// TODO: cast incorrectly removed or wrong type
return (TestCls) lazyOf::get;
```
### After (this PR - raung test):):

```java
return (Memoized) lazyOf::get;  //  Correct marker interface
```

Known Limitation
Javac-compiled bytecode generates cascading CHECK_CAST instructions (first to marker, then to functional interface). During region transformation, the second cast can shadow the first. This is documented in the test TODO. The fix works correctly for hand-written bytecode as demonstrated by the raung test passing.

Implementation Details

- **InvokeCustomNode:** Added `markerInterfaces` field
- **CustomLambdaCall:** Implemented `parseMarkerInterfaces()` method
- **TypeUpdate:** Added `handleInvokeCustom()` for lambda type propagation
- **FixLambdaCastVisitor:** New visitor to update CHECK_CAST instructions
- **Jadx:** Registered visitor in both passes lists
- **SimplifyVisitor:** Added debug logging for troubleshooting

Related
Addresses TODO from #2139 - completes lambda intersection type support

### DIFF
--- a/jadx-core/src/main/java/jadx/core/Jadx.java
+++ b/jadx-core/src/main/java/jadx/core/Jadx.java
@@ -29,6 +29,7 @@ import jadx.core.dex.visitors.DotGraphVisitor;
 import jadx.core.dex.visitors.EnumVisitor;
 import jadx.core.dex.visitors.ExtractFieldInit;
 import jadx.core.dex.visitors.FallbackModeVisitor;
+import jadx.core.dex.visitors.FixLambdaCastVisitor;
 import jadx.core.dex.visitors.FixSwitchOverEnum;
 import jadx.core.dex.visitors.GenericTypesVisitor;
 import jadx.core.dex.visitors.IDexTreeVisitor;
@@ -149,6 +150,7 @@ public class Jadx {
 		}
 		passes.add(new ConstInlineVisitor());
 		passes.add(new TypeInferenceVisitor());
+		passes.add(new FixLambdaCastVisitor());
 		if (args.isDebugInfo()) {
 			passes.add(new DebugInfoApplyVisitor());
 		}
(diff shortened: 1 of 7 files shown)


skylot:
@vicky-dx thanks for the fix :+1:
This PR is big and touches important parts of jadx core, so it might take several days to review it :slightly_frowning_face:

skylot:
@vicky-dx ok, I have reviewed this PR and turns out it is a total bullshit.
Your changes not affect mentioned tests at all, so nothing is fixed. Also, most of added code not executed and looks pointless. The only fine part is code for parsing lambda marker interfaces. But it can't be used now because intersection types not supported in jadx.

As a result proper intersection types support should be implemented in both jadx type system and type inference and after that lambda marker interfaces can be used. 
As a possible workaround, I will try to implement inserting correct cast for lambda at codegen stage, but this might conflict with inferred types.

PR #2673, opened 2025-10-27
Update Chinese localization for various menu items

zhongqingsong:
### Description
translate new texts and update some old texts.


skylot:
@zhongqingsong thank you :+1:

Issue #2687, opened 2025-11-04
[core] Compilation issues due to missing generics types (or missed for-each optimization in LoopRegionVisitor?)


nitram84:
### Issue details

I recompiled this game https://fruit-cut-mania.apk.gold/ and I had some compilation issues due to missing generics types e.g.

`class com.mobaxe.fruitcut.screens.GameScreen`

```java
    private void playAgain() {
        /* this.adIsShowed = false; */ // remove ads
        this.healthScore = 3;
        this.score = 0;
        this.drawChopper = true;
        this.fadeTimeAlpha = 0.0f;
        this.fadeTimeAlphaHealth = 0.0f;
        this.fadeTimeAlphaBG = 0.0f;
        this.fadeTimeAlphaBlood = 0.0f;
        this.docPerScreen = 0;
        if (this.dragPos.size() != 0) {
            this.dragPos.clear();
        }
        this.swipeSound.resume();
        this.throwSound.resume();
        this.meatSliceSound.resume();
        this.chopperSound.resume();
        Iterator i$ = this.stage.getActors().iterator(); // should be 'Iterator<Actor> i$ = this.stage.getActors().iterator();
        while (i$.hasNext()) {
            Actor actor = i$.next();
            actor.setVisible(false);
        }
        gameState = GameState.RUNNING;
        /* if (FruitCut.actionResolver != null) { // remove ads
            FruitCut.actionResolver.showAds(1);
        } */
    }

```
I tracked the issue down and indeed in smali the generics type for `Iterator i$` was missing. To reproduce this error I created a minimal sample:

```java
	package generics;

	import java.util.Iterator;

	public class TestMissingGenericsTypes2<T> implements Iterable<T> {

		@Override
		public Iterator<T> iterator() {
			return null;
		}

		public void test(TestMissingGenericsTypes2<String> l) {
			Iterator<String> i = l.iterator(); // <-- This generics type was removed in smali
			while (i.hasNext()) {
				String s = i.next();
				doSomething(s);
			}
		}

		private void doSomething(String s) {
		}
	}

```

and I edited the class in smali:

```smali
.class public Lgenerics/TestMissingGenericsTypes2;
.super Ljava/lang/Object;
.source "TestMissingGenericsTypes2.java"

# interfaces
.implements Ljava/lang/Iterable;


# annotations
.annotation system Ldalvik/annotation/Signature;
    value = {
        "<T:",
        "Ljava/lang/Object;",
        ">",
[...]

PR #2689, opened 2025-11-05
fix: detect more for-each loops (also prevent issues with missing generics types)

nitram84:
Here is the PR for issue #2687 (for-each loop).

BTW: There is another possibly related unit test `jadx-core/src/test/smali/generics/MissingGenericsTypesTest.smali`, that does not conform to the naming convention. I called my unit test file `TestMissingGenericsTypes2.smali`, so renaming is possible without conflicts.


skylot (approved):

skylot:
> MissingGenericsTypesTest.smali that does not conform to the naming convention

Yeah, that a very old test and it should be adjusted to new convention and use a common tests approach.
Also, I am not sure, if it possible to implement what tests mean to do. Maybe it is easer just remove it :rofl:

PR #2698, opened 2025-11-11
Invalid cast for overloaded method

nitram84:
The latest git version of jadx decompiles this class (compiled with openjdk8)

```java

public class AbstractItem {

	public void doSomething(Container c, Item i) {
		c.add(i);
	}

	public static class Container {

		public <T extends AbstractItem> int add(T t) {
			return 0;
		}

		public void add(AbstractItem... item) {
		}
	}

	public static class Item extends AbstractItem {
	}
}

```

to

```java
public class AbstractItem {

	public void doSomething(Container c, Item i) {
		c.add((Container)i); // <-- invalid cast
	}

	public static class Container {

		public <T extends AbstractItem> int add(T t) {
			return 0;
		}

		public void add(AbstractItem... item) {
		}
	}

	public static class Item extends AbstractItem {
	}
}
```

It seems the arg offsets were not respected in `TypeUtils.getTypeVarMappingForInvoke()`. After adding the arg offset the cast was gone.


skylot (approved):
Super! Thank you :+1:

PR #2702, opened 2025-11-24
Fix #2701

### LINKED ISSUE #2701: [core] Wrong class implementation displayed when an application contains duplicates
### Issue details

When an application contains several classes with the same name (in different dex files), jadx correctly detect the class duplication, but select a random implementation instead of the one used at runtime by android, both in thecli and gui.

```java

```

All those [applications](https://github.com/user-attachments/files/23700711/examples.zip) use the class with `value = "actual"` at runtime, but Jadx sometime display the other implementation with `value = "dummy"`.

### Jadx version

dev

histausse:
Display the correct class implementation in case of duplication (Fix #2701).

When adding a class already present in the mappings of `RootNode`, compare the dex file names are now compared to select the right class.
The GUI list classes from the `JavaPackage`, which contains all classes implementation and appears to select the last one (?). Fixing this is a little tricky because having all classes in `JavaPackage` might be the desired/expected behavior, but it feels strange to perform additional computations from outside an API class to select the right class, so I added another attribute to `JavaPackage` listing classes without the duplicates.

They are probably other places in the code that select an arbitrary implementation instead of the one used at runtime, but this fix the issue for files generated by the cli and the classes displayed by the gui from the tree menu.

### DIFF
--- a/jadx-core/src/main/java/jadx/api/JavaPackage.java
+++ b/jadx-core/src/main/java/jadx/api/JavaPackage.java
@@ -21,6 +21,7 @@ public final class JavaPackage implements JavaNode, Comparable<JavaPackage> {
 	JavaPackage(PackageNode pkgNode, List<JavaClass> classes, List<JavaPackage> subPkgs) {
 		this(pkgNode, classes, classes, subPkgs);
 	}
+
 	JavaPackage(PackageNode pkgNode, List<JavaClass> classes, List<JavaClass> clsListNoDup, List<JavaPackage> subPkgs) {
 		this.pkgNode = pkgNode;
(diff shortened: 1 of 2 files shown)


skylot:
@histausse thanks for your PR :+1:
Although, I slightly refactor and improve your changes in commit 56ae4a604840fe688a2e63e819755b6b37254954

histausse:
Thank for the merge @skylot. 
You missed an edge case in your refactoring: classes1.dex , classes0.dex and any classes0X.dex are not valid dex names. In `isClassesDex()` you should check that the group captured by the regex is not 1 and that it does not start with a "0".

PR #2707, opened 2025-11-29
chore(gui): update obsolete issue template

gemesa:
### Description

While drafting some issue reports, I noticed that the Jadx GUI issue template is obsolete.


gemesa:
Is there something wrong with this update? The version numbers are not available under Help->About but JadxGUI->About JadxGUI instead.

Issue #2128, opened 2024-03-18
[feature] Support DEX v41


skylot:
### Describe your idea

New DEX file format version 41 add support for "containers" - a way to store several DEX files in one physical file.

Currently, no official documentation available.

References to Google implementations:
- in Smali (read-only): https://github.com/google/smali/commit/26f81368a12e3a349278448b8b949f01c221c281
- in `platform/art`: https://android.googlesource.com/platform/art/+/master/libdexfile/dex/dex_file.h#301

This issue create from discussion started in #2127

MrIkso:
https://android.googlesource.com/platform/art/+/a3b8c73d8ae9943c8aaabf49349302b4c05cea71%5E%21/#F8
Commit with adding v41 support to libdexfile

vova7878:
My sample HelloWorld app with dex file version V041 of 3 parts (the first one intentionally contains one empty class). I created this file with my tool, and checked the launch of the app on a device with Android 15. Just rename zip to apk (github does not allow to attach this type of files)

Unfortunately there is no DebugInfo section here, because I haven't written its processing yet, and no HiddenApi section, because this is not a platform dex.

[app-release-signed.zip](https://github.com/user-attachments/files/18768720/app-release-signed.zip)

I hope this will be useful to you

skylot:
@vova7878 thank you for the sample :+1:

I commit a simple approach to support DEX with container, and your APK parsed correctly now :slightly_smiling_face:

sfncat:
When can it be merged into the version? :)))))

PR #2711, opened 2025-11-30
Use length in header for checksum, fix error in Dex v41

kmod-midori:
Dex v41 (#2128) allows storing multiple dex files inside  a single container. Currently the checksum verification uses the whole file for calculation, which fails if there are multiple files inside a single dex, as the checksum only covers the dex entry that is immediately following the header, not the whole file which is actually 2 or more dex files. This leads to JADX failing to load services.jar in various Android 16 ROMs.

This passes all currently available tests, but I don't know if it will fail on ill-formed dex files out there, one can always disable checksum checking tho.

Example file (taking from Samsung OneUI 8.0): 
[services.zip](https://github.com/user-attachments/files/23840771/services.zip)


skylot:
@kmod-midori thank you for fix and sample :+1:

Issue #2727, opened 2025-12-24
[gui] plugin info not refreshed when loading plugin dialog


ewt45:
### Issue details

I'm writing a [plugin](https://github.com/ewt45/jadx-plugin-cfg-viewer), and want to display localized text depending on `JadxSettings.getLangLocale()`. 
`Locale.getDefault()` is using at the very first, locale from jadx settings is acquired and used  later.

I notice that when opening the `Plugin - Manage plugins` dialog, a new `JadxPlugin` instance is created. `JadxPlugin.getPluginInfo()` is called and jadx locale text is returned. However, the old default locale text is displayed. 
For example, if user select English in Jadx, but default locale is Chinese, Chinese is displayed instead of English in the end.

### Jadx version

latest commit 8629e4c

### Java version

21.0.9

### OS

- [ ] Windows
- [x] Linux
- [ ] macOS

skylot:
> when opening the Plugin - Manage plugins dialog, a new JadxPlugin instance is created

If project already opened (i.e. plugins loaded) new instance should not be created, I will fix this.

@ewt45 to be honest, I didn't expect anyone will be changing description depending on locale.
I plan to support locales in plugins, it will be similar to current jadx implementation, so plugins will be able to have own localization files and strings.
I may add to `JadxPluginInfo` field like `descriptionNLSKey`, jadx will load string for current locale from plugin NLS files.

Although, there is issue with current [plugins list](https://github.com/jadx-decompiler/jadx-plugins-list) implementation: it stores only static strings to show info about not installed plugins. Anyway, this also can be fixed by saving other locale strings.

As a workaround for now, I suggest to show description with both languages in separate text blocks.


As for your PR: `JadxPluginInfo` should be static and immutable so it can be fetched once and cached.
And it is better to wait for NLS plugins API implementation.

ewt45:
Thanks for your reply! With the latest commit, new instance won't be created repeatedly now.

PR #2728, opened 2025-12-24
fix: get plugin info dynamically (#2727)

ewt45:
Get plugin info in real time so that localized description can be applied dynamically.

A quick fix for #2727.


skylot:
Check my reply in base issue: https://github.com/skylot/jadx/issues/2727#issuecomment-3690401520

PR #2741, opened 2026-01-08
gui: Add extra UI text in case a third party plugin is involved

jpstotz:
This is a first try to reduce the number of created issues for Jadx that are actually caused by a third party plugin, not Jadx itself.

@skylot I wasn't sure if only third party plugins actually use a module name lik `jadx-plugin:<pluginname>`. Jadx has several internal plugins/libraries do you know which module they belong to?

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/report/ExceptionDialog.java
+++ b/jadx-gui/src/main/java/jadx/gui/report/ExceptionDialog.java
@@ -14,9 +14,14 @@ import java.io.StringWriter;
 import java.lang.management.ManagementFactory;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


skylot:
@jpstotz well, I already implemented plugin detection in a3df83e60f9638840b0e182f559b1eb995ae72a4 commit, check 
https://github.com/skylot/jadx/blob/a3df83e60f9638840b0e182f559b1eb995ae72a4/jadx-gui/src/main/java/jadx/gui/report/JadxExceptionHandler.java#L34

This method search plugin by classloader prefix and resolve plugin by jar, if plugin github project not found exception dialog will not have report button. I might miss something, but it should work for plugins installed by `locationId` from `Available` section in plugins preferences.

I didn't change texts in exception dialog, so you can edit PR to apply your changes in strings (might require to add more plugin info into `ExceptionData` class).

Issue #2744, opened 2026-01-13
[core] A really obfuscated app consumes a lot of resources and is never decompiled completely


bagipro:
### Issue details

I'm trying to decompile the following app: https://drive.google.com/file/d/1dZlRPLi-gqqOtjFtD6ANyGCJBGRbWhwI/view?usp=sharing

But the JadxGUI consumes a lot of resources and decompilation never finished

### Relevant log output or stacktrace

```java
I ran `jstack`, it dumped the following hangs:

"jadx-task-p-15-6" #45 [42915] prio=5 os_prio=31 cpu=81552.16ms elapsed=110.76s tid=0x0000000c328eea00 nid=42915 runnable  [0x000000017ecaf000]
   java.lang.Thread.State: RUNNABLE
	at java.util.Collection.removeIf(java.base@23.0.1/Collection.java:581)
	at jadx.core.dex.visitors.typeinference.TypeUpdateInfo.rollbackUpdate(TypeUpdateInfo.java:45)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.requestUpdate(TypeUpdate.java:229)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.updateTypeForSsaVar(TypeUpdate.java:198)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.updateTypeChecked(TypeUpdate.java:120)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.allSameListener(TypeUpdate.java:491)
	at jadx.core.dex.visitors.typeinference.TypeUpdate$$Lambda/0x00000f8001162e40.update(Unknown Source)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.runListeners(TypeUpdate.java:247)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.requestUpdate(TypeUpdate.java:227)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.updateTypeForSsaVar(TypeUpdate.java:198)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.updateTypeChecked(TypeUpdate.java:120)
	at jadx.core.dex.visitors.typeinference.TypeUpdate.allSameListener(TypeUpdate.java:484)
	at jadx.core.dex.visitors.typeinference.TypeUpdate$$Lambda/0x00000f8001162e40.update(Unknown Source)
[...]

PR #2745, opened 2026-01-15
chore: speed up initial APK loading

jpstotz:
`java.io.File.getCanonicalFile()` seems be pretty slow under certain conditions. 

Switching to java.nio.file significantly improves execution speed and thus reduces the initial loading time JadxGUI requires

The new way to check for path traversal attacks is the recommended way to implement such a check using java.nio.file:
https://www.heise.de/en/background/Secure-Coding-Best-practices-for-using-Java-NIO-against-path-traversal-9996787.html

fixes issue #2744

### DIFF
--- a/jadx-commons/jadx-zip/src/main/java/jadx/zip/security/JadxZipSecurity.java
+++ b/jadx-commons/jadx-zip/src/main/java/jadx/zip/security/JadxZipSecurity.java
@@ -2,6 +2,8 @@ package jadx.zip.security;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


skylot (approved):

skylot:
@jpstotz thanks, looks good to me.
Previous implementation was straightforward and simple, with new approach we will rely on normalize method which is fine. That method has several implementations and looks like these just search for `..` string in path, but jadx already have a check for that string at start of method, so it looks a little pointless. Anyway, I think, new check is fast and might handle other cases, so keeping it is fine.

> fixes issue #2744

Main issue is an endless loop in type inference visitor in `o.ocx` class, I hope I will be able to fix it.

PR #2752, opened 2026-01-21
fix: avoid more false positive throws

ewt45:
This pr avoids more false positive throws in MethodThrowsVisitor.
In `MethodThrowsVisitor.checkInsn()`, there are two places that call `visitThrows()` to add throws, and the other two places that call `filterExceptions()` to add throws.

Only `visitThrows()` calls `isThrowsRequired()` to ignore runtime exceptions, so sometimes `filterExceptions()` adds redundant throws.

`visitThrows()` and `filterExceptions()` are very similar, the only difference is that the former has two more conditions: `excType.isTypeKnown() && isThrowsRequired(mth, excType)`. So I think it's safe to replace the latter with the former. 

A small reproducible example is added.  
```
    public void noThrownExceptions3() {
		// This line leads to false positive throws.
		// 1. assignment 2. invoked method throws runtime exception
        int i = doSomething3(0);
        System.out.print(i);
    }
    public int doSomething3(int i) throws IllegalArgumentException {
        if (i < 0) {
            throw new IllegalArgumentException();
        }
        return 1;
    }
```

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/MethodThrowsVisitor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/MethodThrowsVisitor.java
@@ -1,7 +1,6 @@
 package jadx.core.dex.visitors;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


skylot:
Looks good to me. Thanks :+1:

PR #2777, opened 2026-02-07
Prevent problems arising by parsing duplicate resource entries from resources.arsc

jpstotz:
fixes #2775

* redefines the `ResourceEntry.id` as offset in the arsc file 
* introduces a check to prevent creating the same ResourceEntry multiple times
* in case we somehow still get identical resource ids make sure not to generate resource names that are getting longer and longer by every rename

Issue #2775, opened 2026-02-05
[core] Jadx fails to decompile resources (OutOfMemoryError)


bagipro:
### Issue details

Hey,

I'm trying to decompile this APK: https://drive.google.com/file/d/1lzSqUd1W83FA1ld1dlzrbCS-kAAv-9J1/view?usp=sharing

However, I'm getting the `OutOfMemoryError`

### Relevant log output or stacktrace

```java
Jadx version : dev
Java version : 25.0.1
Java VM      : Oracle Corporation Java HotSpot(TM) 64-Bit Server VM
Platform     : Mac OS X (26.2 aarch64)
Max heap size: 91776 MB
Program args : -Xms128M -XX:MaxRAMPercentage=70.0 -Dawt.useSystemAAFontSettings=lcd -Dswing.aatext=true -Djava.util.Arrays.useLegacyMergeSort=true -Djdk.util.zip.disableZip64ExtraFieldValidation=true -XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.lang=ALL-UNNAMED -Dsun.java2d.noddraw=true -Dsun.java2d.d3d=false -Dsun.java2d.ddforcevram=true -Dsun.java2d.ddblit=false -Dswing.useflipBufferStrategy=True

java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.Arrays.copyOfRange(Arrays.java:3849)
	at java.base/java.lang.String.<init>(String.java:4863)
	at java.base/java.lang.StringBuilder.toString(StringBuilder.java:478)
	at java.base/java.util.Formatter.toString(Formatter.java:2598)
	at java.base/java.lang.String.format(String.java:4455)
	at jadx.core.xmlgen.entry.ResourceEntry.copyWithId(ResourceEntry.java:36)
	at jadx.core.xmlgen.ResTableBinaryParser.buildResourceEntry(ResTableBinaryParser.java:409)
	at jadx.core.xmlgen.ResTableBinaryParser.parseEntry(ResTableBinaryParser.java:373)
	at jadx.core.xmlgen.ResTableBinaryParser.parseTypeChunk(ResTableBinaryParser.java:297)
	at jadx.core.xmlgen.ResTableBinaryParser.parsePackage(ResTableBinaryParser.java:165)
	at jadx.core.xmlgen.ResTableBinaryParser.decodeTableChunk(ResTableBinaryParser.java:112)
[...]

PR #2788, opened 2026-02-12
fix: back to the old correct resource ID calculation

### LINKED ISSUE #2786: [core] Incorrect ID generation in R file breaks constant replacement
### Issue details

After commit 925503ba04d38e02f35268f27e97f38d529eb9dd, the generated R file produced by Jadx appears to contain incorrect resource IDs.

This seems to be a regression, as previous commits generated consistent and correct IDs. The issue likely originates from an incorrect ID calculation logic introduced in this commit.

As a result, the constant replacement function no longer works correctly, since it relies on properly generated resource IDs.

**Observed Behavior**

The generated R file contains incorrect resource IDs.

The constant replacement logic fails due to mismatched or invalid IDs.

This behavior is reproducible consistently after the referenced commit.

**Expected Behavior**

Resource IDs in the generated R file should be calculated correctly.

The constant replacement function should operate as expected.

**Additional Notes**

The issue may be related to the ID calculation logic introduced or modified in commit 925503ba04d38e02f35268f27e97f38d529eb9dd

If needed, I can help test specific scenarios or provide additional examples.

```java

```

_No response_

### Jadx version

unstable-build

jpstotz:
This PR reverts some of the changes introduced by #2777 regarding the resource ID calculation.

As described in #2786 the resource ID is important for constant replacement. 

This PR also includes changes regarding the protection system to protect us against duplicate resource entry and the renaming cascade and memory problem as described in #2775, it now bases on the index, not on the offset.

fixes #2786

### DIFF
--- a/jadx-core/src/main/java/jadx/core/xmlgen/ResTableBinaryParser.java
+++ b/jadx-core/src/main/java/jadx/core/xmlgen/ResTableBinaryParser.java
@@ -296,16 +296,16 @@ public class ResTableBinaryParser extends CommonBinaryParser implements IResTabl
 		}
 		is.skipToPos(entriesStart, "Failed to skip to entries start");
 		int ignoredEoc = 0; // ignored entries because they are located after end of chunk
-		Set<Integer> processedOffsets = new HashSet<>(offsets.size() * 2);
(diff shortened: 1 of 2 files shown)


skylot:
Hm, looks like we need to add an integration test for this :rofl:

PR #2792, opened 2026-02-16
Fix potential resource leak in JadxExternalPluginsLoader.close()

iamsanjaymalakar:
### Problem:
`JadxExternalPluginsLoader.close()` clears classLoaders even if some `URLClassLoader.close()` calls fail, which drops references to loaders that may still hold open resources and prevents retry/diagnosis. If a close fails, the underlying resources (e.g., jar file handles) can remain open and leak.

### Fix:
Track classloaders that fail to close, log the failure, and keep only those loaders in classLoaders so a later `close()` can retry cleanup.

### Test:
```java
./gradlew build
```


skylot:
@iamsanjaymalakar I understand problem, but don't understand solution.
`JadxExternalPluginsLoader.close()` method will be called only once and that instance will not be used anymore, so collecting class loaders for later closing doesn't make any sense.

PR #2784, opened 2026-02-10
feat: Graph views, code pane sync, and more

gordon-f0:
This PR adds multiple features from a private fork we have been working on:

* Graph views (using GraphViz library) - call graphs, inheritance graph, methods graph
* Context menu option to convert decimal constants to hex and binary
* Synchronization between java/smali code panes in split view 
* Speed up renaming process
* Various fixes for decompilation issues e.g. "failed to find switch out block", "region count limit reached", "code restructure failed: missing block"
* Improved try/catch/finally code generation

PR #2801, opened 2026-02-24
fix(gui): make the Split view checkbox correctly toggle between both modes

jpstotz:
I noticed minor bug: The new "Split view" checkbox works for like a three-state button. When switching from split to normal the checkbox often stays checked. 

The problem seems to be that the Split view checkbox not only exists once but multiple times, one persistent instance for the normal view and one instance that is every time created when you switch to the split view mode.

IMHO the code is with my changes is still very complex and a bit "mind-twisted" but now the checkbox state is consistent.

While playing around with the split view I got the feeling that when you activate split view the smali code pane (and the smali code) is loaded multiple times I think up to five times. Also the code position jumps around a few times (beginning -> code pos -> beginning -> code pos) which is not very nice. But I wasn't able to identify the cause for this.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/codearea/ClassCodeContentPanel.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/codearea/ClassCodeContentPanel.java
@@ -47,6 +47,7 @@ public final class ClassCodeContentPanel extends AbstractCodeContentPanel implem
 	private final AtomicBoolean syncInProgress = new AtomicBoolean(false);
 
 	private boolean splitView = false;
+	private final JCheckBox splitCheckboxNormal;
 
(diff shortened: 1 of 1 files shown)


skylot:
@jpstotz thanks, looks good to me.
I wonder is this a regression from PR #2784, or it was broken before?
I want to prepare new release and not include that big PR, but not sure do I need to backport this change :thinking:

jpstotz:
@skylot Just tested it and it was even more broken without that PR.

skylot:
> it was even more broken without that PR

@jpstotz :rofl:

I backport this PR (into new `stable` branch) and fix repaint and UI updates, it works better now.
Tomorrow (if we don't get another nasty issue) I will prepare release from that branch.

PR #2811, opened 2026-02-28
fix(gui): use UI thread for adding logcat messages

jpstotz:
The logcat messages were also added without proper UI synchronization.

I also made other improvements to adb and logcat handling like never closed sockets and never terminated executor.

Also the LogUtils.escape method was a bit restrictive what it filters out. I was very confused as this "log escaper" modified the ADB commands printed in the log. I relaxed that a bit.

As long term solution we may consider switching to `org.apache.commons.text.StringEscapeUtils.escapeJava(...)`.


skylot (approved):
Looks good, thanks

PR #2814, opened 2026-03-02
fix: handle Kotlin 1.9+ $ENTRIES pattern in enum restoration

clawdbot-silly-waddle:

Kotlin 1.9+ compiles enums with an additional `$ENTRIES` field initialized from the `$VALUES` array via `EnumEntriesKt.enumEntries()`. EnumVisitor did not recognize this pattern, causing enum restoration to fail for most Kotlin 1.9+ enum classes.

The fix handles three cases:
- **RegisterArg in convertToEnum()**: when CodeShrinkVisitor cannot inline the filled-array instruction (2 uses), the assign instruction is reached via RegisterArg instead of InsnWrapArg
- **removeEntriesFieldInit()**: cleans up the `$ENTRIES` SPUT, handling both the wrapped case (INVOKE inlined into SPUT by CodeShrinkVisitor) and the standalone case
- **NPE in checkExternalRegUsage()**: `InsnArg.unwrap()` returns null for RegisterArg, causing NPE when checking FILLED_NEW_ARRAY external uses

Includes a smali test compiled from a real Kotlin 2.3 enum.

Tested on Tumblr v43.4.0.107 (`com.tumblr`, 22,124 classes):
- "Failed to restore enum class" warnings: 363 -> 56 (-84.6%)
- "Unknown enum class pattern" warnings: 330 -> 15 (-95.5%)
- Total enum-related warnings: 693 -> 71 (-89.8%)

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/EnumVisitor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/EnumVisitor.java
@@ -49,6 +49,7 @@ import jadx.core.utils.BlockInsnPair;
 import jadx.core.utils.BlockUtils;
 import jadx.core.utils.InsnRemover;
 import jadx.core.utils.InsnUtils;
+import jadx.core.utils.ListUtils;
 import jadx.core.utils.Utils;
 import jadx.core.utils.exceptions.JadxException;
 import jadx.core.utils.exceptions.JadxRuntimeException;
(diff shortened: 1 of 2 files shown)


obus-globus:
The smali test file was generated from the following Kotlin source using `kotlinc 2.3.10`:

```kotlin
package enums

enum class TestEnumKotlinEntries {
    ALPHA,
    BETA,
    GAMMA
}
```

Compiled with `kotlinc TestEnumKotlinEntries.kt -d out/`, then converted to dex with `d8` and disassembled with `baksmali`. Class name adjusted to match the test package.

skylot:
Thanks, code looks good and succesfully restores enum, but it keeps `getEntries()` method which uses removed `$ENTRIES` field:
```java
public enum TestEnumKotlinEntries {
    ALPHA,
    BETA,
    GAMMA;

    public static EnumEntries<TestEnumKotlinEntries> getEntries() {
        return $ENTRIES;
    }
}
```
And it is better to generate compilable code, so I fix this PR to keep `$ENTRIES` field and use `values()` method instead synthetic `$values()`. Result is correct java code:
```java
public enum TestEnumKotlinEntries {
    ALPHA,
    BETA,
    GAMMA;

    private static final /* synthetic */ EnumEntries $ENTRIES = EnumEntriesKt.enumEntries(values());

    public static EnumEntries<TestEnumKotlinEntries> getEntries() {
        return $ENTRIES;
    }
}
```

PR #2817, opened 2026-03-02
Attempt to improve type update performance

deadYokai:
### Description
- To `FinishTypeInference` added `tryLastResort` to attempt resolve type
- Early reject conficting types in `moveListener`
- Attempt to avoid redundant calls of `requestUpdate`

Edit: maybe i need more testing and fix some to work properly

Edit: very large method that requires >905284 seqNums is now decompiles (at least in my case)

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeUpdate.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeUpdate.java
@@ -56,7 +56,6 @@ public final class TypeUpdate {
 			InsnType.ARITH, InsnType.NEG, InsnType.NOT,
 			InsnType.CHECK_CAST, InsnType.INVOKE, InsnType.CONSTRUCTOR);
 
-
 	public static final class PropCtx {
 		final ArrayDeque<PendingWork> queue = new ArrayDeque<>();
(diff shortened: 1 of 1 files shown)


skylot:
This PR is big and change various jadx parts, also it is hard to review and understand without sample and/or tests.

Anyway, your approach with postponing type updates to prevent stack overflow exception will cause a large amount of invalid types became accepted. Currently, jadx will try all possible types until one is accepted, but to postpone deep updates you accept any type and trying to apply it again later, which it is pointless, because invalid types already accepted and this will prevent from finding correct solution.

Overall your approach might be better than exception, but feels incorrect in general.
I plan to implement type updates without recursion, but that will be a big change, so not sure when it will be complete.

deadYokai:
@skylot good luck

skylot:
@deadYokai can you share a sample and class/method name to reproduce type inference issue?
Or maybe open new issues (with samples) for cases you addressed in this PR.

deadYokai:
> @deadYokai can you share a sample and class/method name to reproduce type inference issue? Or maybe open new issues (with samples) for cases you addressed in this PR.

unfortunately, i cannot share it because of possible license violations,  i'll try to create a dummy that will trigger the bug

deadYokai:
@skylot here mega link with .dex file and dummy smali file
https://mega.nz/folder/3UFxiD4B#N3o59gZhvU_vZDk0aiJzYw

PR #2824, opened 2026-03-14
Resolve "launchScriptPath" is null & Add mimetype for desktop file

peasoft:

### Description
1. Resolve `java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because "launchScriptPath" is null` when creating desktop entry without setting `jadx.launchScript.path`.
2. Add mimetype for desktop entry.


skylot (approved):
Thanks

PR #2828, opened 2026-03-21
Fix: resolve class/package name conflicts on case-insensitive filesystems

johnwas:
# Fix: resolve class/package name conflicts on case-insensitive filesystems

## Problem

On case-insensitive filesystems (Windows, macOS with default HFS+), class or package names that differ only by case are treated as the same path. This causes two issues:

1. **Export corruption**: When two classes resolve to the same file path (e.g. `okio/Sink.java` and `okio/sink.java`), the second written file silently overwrites the first — one class is permanently lost.
2. **GUI confusion**: Both classes appear in the tree, but only one can be exported, and users see unexpected content depending on load order.

Common real-world trigger: Kotlin metadata renames an obfuscated class to a lowercase name (e.g. `okio.sink`) that collides with an existing class (e.g. `okio.Sink`) on the same case-insensitive filesystem.

## Solution

Extended `RenameVisitor.checkNames()` with two additional passes, both gated on `!fsCaseSensitive && isRenameCaseSensitive`:

**Package-level conflict detection** (runs before class check):
- Collects all alias package full names lowercased into a `HashSet`
- On collision, renames the conflicting package via `aliasProvider.forPackage()` and loops until the generated name is also unique (same `while`-loop pattern as the class check)

**Class-level conflict detection** (runs after package rename, so package paths are already stable):
- Collects all `getAliasFullPath().toLowerCase()` values into a `HashSet`
- On collision, renames via `aliasProvider.forClass()` + `changeShortName()` and loops until unique
- Adds `RenameReasonAttr` with reason `"case insensitive filesystem"`

Both checks are conditioned on `args.isFsCaseSensitive() == false` (default on Windows) and `args.isRenameCaseSensitive() == true` (included in the default `EnumSet.allOf(RenameEnum.class)`), so existing behavior on Linux is unchanged.

## Tests

**`TestCaseSensitivePkgChecks`** — packages that differ only by case:
- Smali: `com.Example.Foo` + `com.example.Foo` (previously used `Bar` vs `Foo` which never conflicted — fixed to use the same class name so the paths genuinely collide when lowercased)
- Case-insensitive FS: asserts all `getAliasFullPath().toLowerCase()` values are distinct after rename
- Case-sensitive FS: asserts original paths are preserved

**`TestCaseSensitiveClassInPkgChecks`** — classes in the same named package that differ only by case:
- Smali: `com.example.User` + `com.example.user`
- Same two-mode assertions as above

## Changed Files

| File | Change |
|------|--------|
| `jadx-core/src/main/java/jadx/core/dex/visitors/rename/RenameVisitor.java` | Added package-level and class-level case-insensitive conflict detection |
| `jadx-core/src/test/java/jadx/tests/integration/names/TestCaseSensitivePkgChecks.java` | Fixed test comment; smali data corrected (Bar→Foo) |
| `jadx-core/src/test/smali/names/TestCaseSensitivePkgChecks/2.smali` | Changed class from `Bar` to `Foo` to create genuine path conflict |
| `jadx-core/src/test/java/jadx/tests/integration/names/TestCaseSensitiveClassInPkgChecks.java` | New test for same-package class case conflict |
| `jadx-core/src/test/smali/names/TestCaseSensitiveClassInPkgChecks/1.smali` | New smali: `com.example.User` |
| `jadx-core/src/test/smali/names/TestCaseSensitiveClassInPkgChecks/2.smali` | New smali: `com.example.user` |

## Commit

`6597a799` — Fix: resolve class/package name conflicts on case-insensitive filesystems

### DIFF
--- a/jadx-core/src/test/java/jadx/tests/integration/names/TestCaseSensitivePkgChecks.java
+++ b/jadx-core/src/test/java/jadx/tests/integration/names/TestCaseSensitivePkgChecks.java
@@ -12,7 +12,8 @@ import static jadx.tests.api.utils.assertj.JadxAssertions.assertThat;
 public class TestCaseSensitivePkgChecks extends SmaliTest {
 	/*
 	 * com.Example.Foo and com.example.Foo - same class name in packages that differ only by case.
-	 * On case-insensitive FS both would map to the same path (com/example/foo), requiring package rename.
+	 * On case-insensitive FS both would map to the same path (com/example/foo), requiring package
+	 * rename.
 	 */
 
 	@Test


skylot (approved):
Looks good. Applied changes to prevent possible endless loop.

PR #2851, opened 2026-04-14
fix: harden XML parser in FileTypeDetector against XML bomb DoS

RuffaloLavoisier:

### Description

A crafted APK containing a resource with nested entity expansion (Billion Laughs) could cause excessive memory consumption and crash jadx during file type detection.


skylot (approved):
Thanks.
Although, this code now looks the same as in
https://github.com/skylot/jadx/blob/d84f0389ec5356942ce07219799152ecae842b27/jadx-core/src/main/java/jadx/api/security/impl/JadxSecurity.java#L99-L105

I will prepare a change to use `parseXml` from jadx security class instead this.

Issue #2770, opened 2026-01-31
[core] switch over string, but the first switchRegion is a ifRegion


ewt45:
### Issue details

`switch(string)` is splitted into two `switch`es in smali. but d8 optimization sometimes [changes the first `switch` into `if`](https://jakewharton.com/d8-optimizations/). Currently SwitchOverStringVisitor doesn't handle this situation.

### Relevant log output or stacktrace
decompiled code
```java
/* JADX WARN: Removed duplicated region for block: B:34:0x0070  */
    /*
        Code decompiled incorrectly, please refer to instructions dump.
    */
    private void changeUIByCurFragment() {
        String tag = getSupportFragmentManager().findFragmentById(2131296368).getTag();
        byte b = 0;
        if (tag == null) {
            setHomeIsActionBack(false);
        }
        if (tag.equals(FRAGMENT_TAG_CONTAINER_PROP) || tag.equals(FRAGMENT_TAG_CHOOSE_FILE)) {
            setHomeIsActionBack(true);
        } else {
            setHomeIsActionBack(false);
        }
        int iHashCode = tag.hashCode();
        if (iHashCode != -2019877892) {
            if (iHashCode != -1058712772) {
                if (iHashCode != -487288020) {
                    b = (iHashCode == 2807996 && tag.equals(FRAGMENT_TAG_INSTALL_NEW)) ? (byte) 2 : (byte) -1;
                } else if (tag.equals(FRAGMENT_TAG_MANAGE_CONTAINERS)) {
                    b = 3;
                }
            } else if (tag.equals(FRAGMENT_TAG_START_MENU)) {
                b = 1;
            }
        } else if (!tag.equals(FRAGMENT_TAG_DESKTOP)) {
        }
        switch (b) {
            case 0:
                this.mNavigationView.setCheckedItem(2131296369);
                break;
            case 1:
                this.mNavigationView.setCheckedItem(2131296373);
                break;
            case 2:
                this.mNavigationView.setCheckedItem(2131296371);
                break;
            case 3:
                this.mNavigationView.setCheckedItem(2131296372);
                break;
        }
    }
```

### Provide sample and class/method full name

[sample smali](https://github.com/user-attachments/files/24976629/1.txt)

### Jadx version

latest (331c4aa)

Issue #2359, opened 2024-12-03
[core] restoreSwitchOverString fails


steelbytes:
### Issue details

switch over string fails

### Relevant log output or stacktrace

```java
/* JADX WARN: Can't fix incorrect switch cases order, some code will duplicate */
    /* JADX WARN: Failed to restore switch over string. Please report as a decompilation issue
    java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "jadx.core.dex.visitors.regions.SwitchOverStringVisitor$SwitchData.getNewCases()" is null
    	at jadx.core.dex.visitors.regions.SwitchOverStringVisitor.restoreSwitchOverString(SwitchOverStringVisitor.java:109)
    	at jadx.core.dex.visitors.regions.SwitchOverStringVisitor.visitRegion(SwitchOverStringVisitor.java:66)
    	at jadx.core.dex.visitors.regions.DepthRegionTraversal.traverseIterativeStepInternal(DepthRegionTraversal.java:77)
    	at jadx.core.dex.visitors.regions.DepthRegionTraversal.traverseIterativeStepInternal(DepthRegionTraversal.java:82)
    	at jadx.core.dex.visitors.regions.DepthRegionTraversal.traverseIterative(DepthRegionTraversal.java:31)
    	at jadx.core.dex.visitors.regions.SwitchOverStringVisitor.visit(SwitchOverStringVisitor.java:60)
     */
    /* JADX WARN: Removed duplicated region for block: B:27:0x0070 A[ORIG_RETURN, RETURN] */
    /* JADX WARN: Removed duplicated region for block: B:5:0x006e A[RETURN, SYNTHETIC] */
    @Override // com.piaggio.bike.data.vehicle.IVehicleModel
    /*
        Code decompiled incorrectly, please refer to instructions dump.
    */
```

### Provide sample and class/method full name

apk from https://apkpure.com/piaggio/it.piaggio.app
class com.piaggio.bike.data.vehicle.VehicleModel method enableClockSync

6x68:
the switch over string visitor seems to have a bug where if you have 3+ nested switches (inside a default block, might work in case blocks too).
if you just want to break the restore switch over string visitor:
```java
switch (v) {
	case "a":
		return 0;
	default:
		switch (v) {
			case "b":
				return 1;
			default:
				switch (v) {
					case "c":
						return 2;
					default:
						return 3;
		}
	}
}
```
[...]

PR #2860, opened 2026-04-27
fix: cover more situations for SwitchOverString

ewt45:
Currently SwitchOverStringVisitor supports the normal format of switch over string, which contains two switches in smali.

Sometimes, android's d8/r8 will make some further modification.
- 1st `switch` could be changed to `if`s to reduce size of dex. (#2770)
- 2nd `switch` could be flattened and removed if there are many cases using a same block. (#2359)

This pr covers these two situations. It splits into 3 parts:
- `collectPart1RegionCases`: Validate switch over string pattern. Store str and num/caseBlock in switchData.
- `prepareMergedSwitchCases`: Create cases according to part1/part2Region. Replace keys with strings.
- `replaceWithMergedSwitch`: Replace with new switch. Remove original code.


skylot (approved):

skylot:
@ewt45 thanks, looks good :+1:
Although, I will do a bit of code clean up later.
There are several places with bad code style choices, like:
```java
Objects.requireNonNull(varName).callMethod()
```
it is same as
```
varName.callMethod()
```
Both will throw NPE if `varName` is null, but second is shorter and latest java will provide better error description.

Also, I prefer to remove these lines in tests
```java
disableCompilation();
allowWarnInCode();
```
because compilation is the best way to validate result code, often it is easy to fix smali code to not use external objects and methods.
And it is better to fix all warnings, it simplifies catching regression in the future.

ewt45:
@skylot Thanks for the explanation. 
`Objects.requireNonNull` is added because of my IDE's warning. `disableCompilation` is mistakenly added when testing. these two can be removed safely.

as for `allowWarnInCode();`, currently it's also used in `TestConditions22.java` and I'm not sure what to do with it.

Please feel free to clean up the code.

PR #2863, opened 2026-05-01
feat: add "copy reference" to context menu

xxr0ss:
### Description

Added "Copy Reference" to the context menu, same as in IDEA — copies `com.example.Class.method` to clipboard. Works on classes, methods, and fields

Shortcut: `R` (open to suggestions if anything else fits better)

### Motivation

When reverse engineering, it's common for people to copy reference for writing scripts (frida/xposed or else), writing documents, etc. 

Currently when copying code reference, we can hover and copy from popup. But this is slow and inconvenient (and the popup widget easily loses focus, even thought I pressed `F2`）

I think it would be very helpful to have an menu option to do this :)

### Preview

image

- class: `com.example.MyClass`
- method: `com.example.MyClass.onCreate`
- field: `com.example.MyClass.TAG`
- method of inner class: `com.example.Outer$Inner.doSomething`

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/action/ActionModel.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/action/ActionModel.java
@@ -107,6 +107,8 @@ public enum ActionModel {
 			Shortcut.keyboard(KeyEvent.VK_F)),
 	XPOSED_COPY(CODE_AREA, "popup.xposed", "popup.xposed", null,
 			Shortcut.keyboard(KeyEvent.VK_Y)),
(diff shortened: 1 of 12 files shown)


xxr0ss:
Quick question: I modified i18n files for `en` and `zh`, for the rest files I left placeholders as `#popup.copy_reference=Copy Reference`. Should I keep the placeholders, or is it okay to fill them in with AI translations? 

Curious which is preferred 🤔

skylot (approved):

skylot:
@xxr0ss thanks, looks great :+1:

> Should I keep the placeholders, or is it okay to fill them in with AI translations?

Commented placeholder is preferred, because translation without context can be incorrect.

Issue #2875, opened 2026-05-15
[gui] Inconsistent pasted file handling in file chooser


mbv06:
### Issue details

I used jadx on Ubuntu for a long time and got used to copying files in the system file manager and pasting them directly into the `File -> Open` file name field. I was surprised to find that this workflow was not handled consistently on the other OS.

_To be honest, I am not sure whether functionality was designed, and it should be labeled as a bug, as it doesn't work correctly, or it's a "feature request"_  😅

### Problem

- on Windows, pasting a copied file into the `File name` field did nothing;
- on macOS, pasting a copied file inserted only the file name, not the full path;
- on Ubuntu, pasting a copied file worked fine, but when multiple copied files were pasted, the files were separated with a blank space; as a result, they were handled as a single file path.

### Jadx version

1.55

### Java version

25.0.2

### OS

- [x] Windows
- [x] Linux
- [x] macOS

skylot:
@mbv06 nice PR :+1:
Although, as I understand, most users just drag files into jadx window, and very few will discover this feature.
Anyway, it looks useful, so thank you for your contribution.

PR #2876, opened 2026-05-15
fix(gui): handle pasted file lists in file chooser

mbv06:
### Description

This PR fixes pasting files/folders copied from the system file manager into the Jadx file chooser  https://github.com/skylot/jadx/issues/2875

When copied files are available in the clipboard as `javaFileListFlavor`, jadx now pastes their absolute paths into the `File name` field, wrapping each path in quotes. Regular text paste behavior is left unchanged.

### Fix

Add handling for clipboard lists of files/folders in the Swing file chooser, and paste copied files/folders as quoted absolute paths.

### Testing

I manually tested copying one or more files/folders from the system file manager and pasting them into the jadx file opener on:

- Windows 10, 11
- macOS 15, 26
- Ubuntu 24.04 LTS

In all cases, the copied paths are pasted into the file chooser, and jadx opens the selected files/folders as expected.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/filedialog/CustomFileChooser.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/filedialog/CustomFileChooser.java
@@ -1,19 +1,29 @@
 package jadx.gui.ui.filedialog;
 
 import java.awt.Component;
+import java.awt.Container;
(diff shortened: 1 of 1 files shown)


skylot (approved):
Looks great. Thank you.

PR #2879, opened 2026-05-22
Change from virtualenv to venv

paraeee:
## Summary

This PR changes unified virtual environment used for quark engine

## Problem
Any system other than windows use virtualenv that not guarantee installed in the system or inflict with  system-managed Python environments, 

## Solution
Change the virtual environment to python built in environment

## Test Plan

- [x] Build passes with `./gradlew dist`
- [x] Tested on Linux 7.0.9-arch1-1


skylot (approved):
Looks good to me :+1:

PR #2887, opened 2026-06-14
fix: keep all dimensions for anewarray of an array element type

### LINKED ISSUE #2885: [core] anewarray of an array element drops a dimension (java/class input)
### Issue details

On the java/class input path, `new long[n][]` decompiles to `new long[n]` (the outer dimension is dropped) and the method gets a spurious `Type inference failed` warning.
This happens for any `anewarray` whose element type is itself an array.

`InsnDecoder.makeNewArray` ([InsnDecoder.java:529](https://github.com/skylot/jadx/blob/8c28a8530e0fb2d2caba930b38f0f936f118d490/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java#L529)) cannot tell java `anewarray` (operand is the element type, e.g. `[J` for `new long[n][]`, with dim literal 1) from `multianewarray` (operand is the full result type); both arrive as `Opcode.NEW_ARRAY`.
The `if (indexType.isArray())` shortcut ([InsnDecoder.java:536](https://github.com/skylot/jadx/blob/8c28a8530e0fb2d2caba930b38f0f936f118d490/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java#L536), added in 7576f9cd) treats any array operand as already the full type, so `anewarray` of an array element keeps only the element type and loses the outer dimension.
The dex path is unaffected (dalvik `new-array` carries the full type with dim 0).
`multianewarray` (e.g. `new int[a][b]`) is correct and must stay correct.

### Sample

Compile with `javac` and run jadx on the `.class`:

```java
public class ArrBug {
    public static long[][] makeLong(int n) { return new long[n][]; }
    public static String[][] makeStr(int n) { return new String[n][]; }
    public static int[][][] makeDeep(int n) { return new int[n][][]; }
    public static int[][] makeMulti(int a, int b) { return new int[a][b]; }
}
```

Current output on master (8c28a85):

```java
public static long[][] makeLong(int i) {
    return new long[i];      // expected: new long[i][]
}
public static String[][] makeStr(int i) {
    return new String[i];    // expected: new String[i][]
}
public static int[][][] makeDeep(int i) {
    return new int[i][];     // expected: new int[i][][]
}
public static int[][] makeMulti(int i, int i2) {
    return new int[i][i2];   // correct (multianewarray)
}
```

Each broken method also gets a `JADX WARN: Type inference failed ...` comment.

### Jadx version

master `8c28a85` (current HEAD)

obus-globus:

### Description

On the java/class input path, `anewarray` of an array element (e.g. `new long[n][]`) lost its outer dimension and decompiled to `new long[n]`, with a spurious `Type inference failed` warning.

`makeNewArray` is shared by all inputs and wraps the operand type by the NEW_ARRAY literal. `anewarray`/`newarray` already pass the element type with literal 1, and dalvik `new-array` passes the full array type with literal 0. `multianewarray` also passes the full array type, but was setting the literal to its dimension count, so `makeNewArray` relied on an `indexType.isArray()` shortcut to avoid double-wrapping it. That shortcut also caught `anewarray` whose element is itself an array, dropping the outer dimension.

Set the `multianewarray` literal to 0 (the operand is already the full type, like `new-array`; the dimension count comes from regsCount) and drop the now-redundant `isArray()` shortcut, so `makeNewArray` just wraps by the literal. The dex path is unaffected.

Includes an integration test.

Fixes #2885

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java
+++ b/jadx-core/src/main/java/jadx/core/dex/instructions/InsnDecoder.java
@@ -528,18 +528,10 @@ public class InsnDecoder {
 
 	private InsnNode makeNewArray(InsnData insn) {
 		ArgType indexType = ArgType.parse(insn.getIndexAsType());
+		// NEW_ARRAY literal = dimensions to wrap the operand by: 0 if it is already the full array type
+		// (dalvik new-array, java multianewarray), 1 for newarray/anewarray (operand is the element type)
 		int dim = (int) insn.getLiteral();
-		ArgType arrType;
-		if (dim == 0) {
-			arrType = indexType;
-		} else {
-			if (indexType.isArray()) {
-				// java bytecode can pass array as a base type
-				arrType = indexType;
-			} else {
-				arrType = ArgType.array(indexType, dim);
-			}
-		}
+		ArgType arrType = dim == 0 ? indexType : ArgType.array(indexType, dim);
 		int regsCount = insn.getRegsCount();
 		NewArrayNode newArr = new NewArrayNode(arrType, regsCount - 1);
 		newArr.setResult(InsnArg.reg(insn, 0, arrType));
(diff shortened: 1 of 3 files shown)


skylot (approved):
Nice fix. Thanks :+1:

PR #2888, opened 2026-06-14
fix: erase unresolved generic field type to its bound in field-get assign

### LINKED ISSUE #2886: [core] generic field read through a raw type is compared to 0 instead of null
### Issue details

Reading a generic field through a raw instance type and null-checking it emits `== 0` instead of `== null`, with a `type inference failed` warning.
It affects both a plain type-variable field (`T`) and an array of one (`T[]`).

`TypeBoundFieldGetAssign.getResultType` ([TypeBoundFieldGetAssign.java:42](https://github.com/skylot/jadx/blob/8c28a8530e0fb2d2caba930b38f0f936f118d490/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeBoundFieldGetAssign.java#L42)): when the field's type cannot be resolved from the (raw) instance type, `replaceClassGenerics` returns null and the fallback returns the field type unchanged (the existing `// TODO: check if this type is allowed in current scope`, [line 47](https://github.com/skylot/jadx/blob/8c28a8530e0fb2d2caba930b38f0f936f118d490/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeBoundFieldGetAssign.java#L47)).
That leaves an out-of-scope type variable owned by another class; type propagation rejects it (TypeUpdate.checkForUnknownTypeVars), so the compared `0` literal keeps its unknown type and TypeGen renders it as `0`.
The bound is built for any field whose type contains a type variable ([TypeInferenceVisitor.java:271](https://github.com/skylot/jadx/blob/8c28a8530e0fb2d2caba930b38f0f936f118d490/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeInferenceVisitor.java#L271), `containsTypeVariable()`), so `T[]` is affected too.

### Sample

Compile with `javac` and run jadx on the `.class` files:

```java
public class GenBug {
    public static class Box<T extends Number> {
        public T value;
        public T[] arr;
    }
    public static boolean isNull(Box box) {   // raw Box
        return box.value == null;
    }
    public static boolean arrNull(Box box) {
        return box.arr == null;
    }
}
```

Current output on master (8c28a85):

```java
public static boolean isNull(Box box) {
    return box.value == 0;   // expected: == null
}
public static boolean arrNull(Box box) {
    return box.arr == 0;     // expected: == null
}
```

### Jadx version

master `8c28a85` (current HEAD)

obus-globus:

### Description

Reading a generic field through a raw instance type and null-checking it emitted `== 0` instead of `== null` (with a `type inference failed` warning), for both a type-variable field (`T`) and arrays of one (`T[]`).

When the field's type variable can't be resolved from the raw instance type, `getResultType` returned it unchanged, leaving an out-of-scope type variable that type propagation rejects, so the compared `0` literal kept its unknown type. Erase the unresolved variable to its leftmost bound (like javac), arrays element-wise, falling back to `Object` when unbounded or when the bound is itself a type variable. This mirrors the extend-type handling in `FixTypesVisitor`, rather than the straight-to-`Object` erasure in `checkRawType`, so a `T extends Number` field stays usable as a `Number`.

Includes an integration test.

Fixes #2886

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeBoundFieldGetAssign.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeBoundFieldGetAssign.java
@@ -1,5 +1,7 @@
 package jadx.core.dex.visitors.typeinference;
 
+import java.util.List;
+
 import jadx.core.dex.info.FieldInfo;
 import jadx.core.dex.instructions.IndexInsnNode;
 import jadx.core.dex.instructions.args.ArgType;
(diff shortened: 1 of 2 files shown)


jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeBoundFieldGetAssign.java
skylot, line 45: Support for missing generic types (raw and wildcard (`<?>`)) should be done inside `TypeUtils.replaceClassGenerics` method.
This will resolve many other issues and not only with field types.
But such task can be hard one, you can try to solve it, I also will try to look into this.

skylot (changes requested):

jadx-core/src/main/java/jadx/core/dex/visitors/typeinference/TypeBoundFieldGetAssign.java
obus-globus: Agreed, `replaceClassGenerics` is the right home for this in general.

I tried it there (erase to the declared bound when a variable can't be resolved from the instance), but it regresses `TestTypeResolver12`: a `null` return there can't tell the class's own in-scope `T` (`WeakReference<T>` read inside `TestCls<T>`, must stay `T`) from an out-of-scope raw-instance variable that should be erased, so erasing both breaks multi-variable inference. The field-get bound has that context; `replaceClassGenerics` doesn't.

Not sure how to continue here, let me know if you need something.

PR #2897, opened 2026-06-23
fix(gui): fix token type error leading to disables right-click context menu rename button

Plasion:
### Problem
RSyntaxTextArea marks all tokens whose lexemes match class names under `java.lang`, `java.io`, and `java.util` as `TokenTypes.FUNCTION`. This happens even when the token is not a corresponding class identifier — such as a method, field, or local variable identifier. This problem not only causes incorrect syntax highlighting but also disables the rename button in the right-click context menu.
### Solution
Refer to the `mergeLongClassNames()` in `JadxTokenMaker`, look up metadata and correct the token types accordingly.
### Figures
result1
result2

### Changed Files
| File | Change |
| --- | --- |
| jadx-gui/src/main/java/jadx/gui/ui/codearea/JadxTokenMaker.java | Correct token types based on metadata. |
| jadx-gui/src/main/java/jadx/gui/ui/codearea/CodeArea.java | Added a modified version of `getJavaClassIfAtPos()` to retrieve metadata other than just class. |
| jadx-core/src/main/java/jadx/core/codegen/ClassGen.java | Fixed an import conflict where `plasion.data.Object` would be redundantly imported after `java.lang.Object` was already imported (see Figure 1). |
| jadx-core/src/main/java/jadx/core/dex/info/ClassInfo.java | Add doc for `getParentClass()` and `getTopParentClass()`. The naming of `getParentClass()` cost me a lot of reading code time. I hope other developers won’t run into the same confusion. |

### DIFF
--- a/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
+++ b/jadx-core/src/main/java/jadx/core/codegen/ClassGen.java
@@ -122,17 +122,18 @@ public class ClassGen {
 	private void addImports(ICodeWriter clsCode) {
 		int importsCount = imports.size();
 		if (importsCount != 0) {
-			List<ClassInfo> sortedImports = new ArrayList<>(imports);
-			sortedImports.sort(Comparator.comparing(ClassInfo::getAliasFullName));
-			sortedImports.forEach(classInfo -> {
-				clsCode.startLine("import ");
-				ClassNode classNode = cls.root().resolveClass(classInfo);
-				if (classNode != null) {
-					clsCode.attachAnnotation(classNode);
-				}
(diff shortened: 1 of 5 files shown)


skylot:
@Plasion
I commit a test to verify imports correctness and other small code changes.
But I am still not sure about code highlighting part, can you share your test case (jar is fine)? It looks nice, and I don't want to recreate it manually.

Plasion:
[Test.zip](https://github.com/user-attachments/files/29655703/Test.zip)
GitHub doesn't support jar files, so I changed the extension to zip.

skylot (approved):

PR #2898, opened 2026-06-24
Fix ImmutableList.lastIndexOf skipping index 0

vasiliy-mikhailov:
## Problem

`ImmutableList.lastIndexOf` scans the backing array backwards but uses `i > 0` as the loop condition:

```java
for (int i = arr.length - 1; i > 0; i--) {
    if (obj.equals(arr[i])) {
        return i;
    }
}
return -1;
```

Because the loop stops before `i == 0`, index 0 is never examined. An element that exists only at index 0 — including any single-element list — returns `-1` instead of `0`.

## Fix

Use `i >= 0` so index 0 is checked.

## Test

Adds `ImmutableListTest` covering `lastIndexOf`, including an element only at index 0 and a single-element list. The index-0 cases fail before the change (`expected: <0> but was: <-1>`) and pass after.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/utils/ImmutableList.java
+++ b/jadx-core/src/main/java/jadx/core/utils/ImmutableList.java
@@ -60,7 +60,7 @@ public final class ImmutableList<E> implements List<E>, RandomAccess {
 
 	@Override
 	public int lastIndexOf(Object o) {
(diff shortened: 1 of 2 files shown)


skylot (approved):
Thanks :+1:

PR #2899, opened 2026-06-24
Handle empty string in Utils.cleanObjectName and cutObject

vasiliy-mikhailov:
## Problem

`Utils.cleanObjectName` and `Utils.cutObject` both start by calling `obj.charAt(0)`:

```java
public static String cleanObjectName(String obj) {
    if (obj.charAt(0) == \x27L\x27) { ... }
}
```

Passing an empty string throws `StringIndexOutOfBoundsException: Index 0 out of bounds for length 0`.

## Fix

Return the empty string unchanged when the input is empty, in both methods.

## Test

Adds `UtilsEmptyStringTest` covering `cleanObjectName("")` and `cutObject("")` (plus normal cases). The empty-string cases fail on `master` (`StringIndexOutOfBoundsException`) and pass with this change.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/utils/Utils.java
+++ b/jadx-core/src/main/java/jadx/core/utils/Utils.java
@@ -41,6 +41,9 @@ public class Utils {
 	}
 
 	public static String cleanObjectName(String obj) {
+		if (obj.isEmpty()) {
+			return obj;
+		}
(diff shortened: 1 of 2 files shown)


skylot:
Empty string is invalid argument in these methods, so exception is fine, but returning empty string is bad and will break something later.

PR #2906, opened 2026-07-04
fix: keep cast that changes operand arithmetic type in arith instructions

obus-globus:

### Description

`(float) a / (float) b` with `int` operands decompiled to `a / b`, an integer division, so the result was truncated: `floatDiv(1, 2)` returned `0.0f` instead of `0.5f`. It reproduces on both the dex and java/class paths, since the cast is dropped in `SimplifyVisitor`, which is shared by all inputs.

`SimplifyVisitor.processCast` drops a primitive cast feeding an `ARITH` insn when `isArithWideUpCast` reports it as unnecessary. That helper only kept casts that grow the operand register count (`castToType.getRegCount() > argType.getRegCount()`). `int` and `float` share a register count, so the `int`->`float` cast was dropped and jadx emitted an integer division, then relied on a widening at the result that does not happen (the division is already done in `int`).

Keep any cast that changes the operand arithmetic type (`int`/`long`/`float`/`double`), because operand types select the arith instruction in the result bytecode. Types narrower than `int` are treated as `int`. The `(long) i << 32` case the old check handled stays covered, since it is also a type change.

Includes an integration test.

### DIFF
--- a/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
+++ b/jadx-core/src/main/java/jadx/core/dex/visitors/SimplifyVisitor.java
@@ -28,6 +28,7 @@ import jadx.core.dex.instructions.args.ArgType;
 import jadx.core.dex.instructions.args.InsnArg;
 import jadx.core.dex.instructions.args.InsnWrapArg;
 import jadx.core.dex.instructions.args.LiteralArg;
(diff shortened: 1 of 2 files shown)


mostafaNazari702:
@obus-globus Why not close it yourself if you clearly see that it is slop code?

PR #2911, opened 2026-07-10
gui: Copy full code snippet when copy all search results

nstion:
### Changes
Modify `CommonSearchDialog#copyAllSearchResults` to also copy matched code snippet from search table, format each output line as:
`[node ref]   code line`

1. Append matched code snippet after node CodeNodeRef when using the Copy button


skylot (approved):
Looks good, thanks

PR #2914, opened 2026-07-11
feat(gui): programmatic open/close project + init plugins with no project open

### LINKED ISSUE #2913: [feature] More automation to JADX
### Describe your idea

I have multiple apps, the same app across multiple versions and i want the AI ( through MCP plugin ) to process them autonomously. The workflow should be: open one APK, analyze it, close it, then open the next, repeating until it's done.

Right now, the "close" and "open" steps require me to manually click through the GUI every time, which unfortunately makes that impossible.

Would it be possible for the plugin manager to run at the application scope instead of the project scope? Currently, any MCP plugin is terminated as soon as a project is closed. That means the plugin goes offline between APKs, making it impossible to automate this kind of multi-project workflow.

mostafaNazari702:
### Description

closes: #2913.

I have been running jadx-gui with an MCP plugin that exposes a small local HTTP endpoint so i can control it from another tool. The workflow is: open an APK, analyze it, close it, then open the next one, without using the GUI. This PR adds support for that.

For the Open/close projects programmatically, MainWindow already has the logic to open and close projects but only through the GUI. What i did is basically add two methods that do the same thing without showing dialogs:

* `closeProject(boolean save, Runnable onFinish)` close the current project, optionally saving it first, then call `onFinish`.
* `openProject(List<Path> paths, Runnable onFinish)` opens the given files as a new project and calls `onFinish` after loading finishes.

Both use the already existing open/close code just without the file chooser and save prompts.

Now i made it possible to initialize plugins before the first project, GUI plugins are currently started only after a project is loaded so a plugin can't do anything before the first APK is opened.

If jadx starts without a project, it now starts plugins against an empty workspace. `JadxWrapper.initPluginsForEmptyWorkspace()` runs the plugin initialization part of `open()` but skips input validation since there are no files yet. Opening a real project later works the same as before.

No changes to existing behaviors done.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/JadxWrapper.java
+++ b/jadx-gui/src/main/java/jadx/gui/JadxWrapper.java
@@ -87,6 +87,31 @@ public class JadxWrapper {
 		}
 	}
 
+	public void initPluginsForEmptyWorkspace() {
+		synchronized (DECOMPILER_UPDATE_SYNC) {
+			if (decompiler != null) {
+				return; // a project (or a previous empty init) is already active
+			}
+			try {
+				JadxArgs jadxArgs = getSettings().toJadxArgs();
+				jadxArgs.setPluginLoader(new JadxExternalPluginsLoader());
+				jadxArgs.setFilesGetter(JadxFilesGetter.INSTANCE);
+				JadxAppCommon.applyEnvVars(jadxArgs);
+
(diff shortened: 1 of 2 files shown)


skylot:
This PR looks more like a workaround, and since plugins also should be changed it is better to implement proper approach. Check my comment for details: https://github.com/skylot/jadx/issues/2913#issuecomment-4948389343

PR #2915, opened 2026-07-11
feat(gui): use ResourcePanel with file-type specific tabs.

jpstotz:
Resources can now be viewed in one ResourcePanel which shows one or more tabs based on the file-type. That way we can reuse existing resource views like code/text and hex views.

E.g. for images and fonts now also the hex view tab is shown.

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/codearea/BinaryContentPanel.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/codearea/BinaryContentPanel.java
@@ -4,8 +4,6 @@ import java.awt.BorderLayout;
 import java.awt.Component;
 import java.nio.charset.StandardCharsets;
 
-import javax.swing.JSplitPane;
-import javax.swing.JTabbedPane;
 import javax.swing.SwingUtilities;
 import javax.swing.border.EmptyBorder;
 
@@ -13,12 +11,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import jadx.api.ResourcesLoader;
-import jadx.api.resources.ResourceContentType;
 import jadx.gui.jobs.BackgroundExecutor;
-import jadx.gui.jobs.IBackgroundTask;
-import jadx.gui.jobs.TaskWithExtraOnFinish;
-import jadx.gui.settings.JadxSettings;
-import jadx.gui.settings.LineNumbersMode;
 import jadx.gui.treemodel.JNode;
 import jadx.gui.treemodel.JResource;
 import jadx.gui.ui.hexviewer.HexPreviewPanel;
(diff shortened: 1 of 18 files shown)


jadx-gui/src/main/java/jadx/gui/ui/codearea/BinaryContentPanel.java
```diff
@@ -27,26 +20,17 @@
...
-		}
 		hexPreviewPanel = new HexPreviewPanel(getSettings());
 		hexPreviewPanel.getInspector().setVisible(false);
+		add(hexPreviewPanel, BorderLayout.CENTER);
 
-		areaTabbedPane = buildTabbedPane();
-		add(areaTabbedPane);
-
-		SwingUtilities.invokeLater(this::loadSelectedPanel);
+		SwingUtilities.invokeLater(this::loadHexView);
```
skylot: Hex panel loaded even if it is not active, better to avoid this.
Also wrong thread is used: to enable thread guard warnings add `JADX_GUI_DEBUG=true` to env variables.

I think all tabs should be loaded on tab activation and not load content on constructor call.
This might require to add common/interface method like `load` and call it in tab panel `addChangeListener`.
Such approach will be the same as in `ClassCodeContentPanel` for code tabs (maybe extract common code later :rofl:)

skylot (changes requested):

jpstotz:
@skylot I updated the BinaryCodePanel to use lazy loading. In addition the data retrieval now used a new class `LazyLoadingBinaryData` which loads the data only on demand block by block (block size 512KB). 

That means when BinaryContentPanel initially loads data only the first 512KB are loaded. If the user scrolls to different position in the file the file data is loaded block-by-block from InputStream and saved internally. This means Jadx will no longer try to load the full resource to a byte array. That improves viewing of very large binary resource files and also avoids potential problems with RAM/heap size if the users just opens such a large resource.

BTW: There was a problem with the new IOException handler. The used translation message `io_error_dialog.close` was missing which leads to an endless loop.

jadx-gui/src/main/java/jadx/gui/ui/codearea/BinaryContentPanel.java
jpstotz: Thanks for your review. I wasn't thiuking that far when developing this PR.

The suggestion is good. I added a ILazyLoad interface. It's `loadData()` method is called when the tab is shown for the first time.

I also tested with JADX_GUI_DEBUG=true. However thread guard warning in BinaryContentPanel that fired in the version you reviewed: I am not sure if that call to `uiThreadGuard()` was correctly placed. 

Anyway the loading system of BinaryCodePanel has been completely changed so that part is now outdated. I mark this as resolved.

jadx-gui/src/main/java/jadx/gui/ui/codearea/BinaryContentPanel.java
```diff
@@ -4,160 +4,92 @@
...
 
 	private void loadHexView() {
 		if (hexPreviewPanel.isDataLoaded()) {
 			return;
 		}
-		UiUtils.notUiThreadGuard();
-		byte[] bytes = getNodeBytes();
-		UiUtils.uiRunAndWait(() -> hexPreviewPanel.setData(bytes));
+		LOG.debug("Loading Hex View of {}", node.getName());
+		UiUtils.uiRunAndWait(() -> hexPreviewPanel.setData(getNodeData()));
```
skylot, line 47: Hm, now data loading executed in UI thread, it is better to avoid this, but since data mostly lazy loaded it is fine for now.

skylot (approved):
Looks good now. Thank you :+1:

skylot:
> BTW: There was a problem with the new IOException handler. The used translation message io_error_dialog.close was missing which leads to an endless loop.

Oh, that's bad, it is worth to prepare a new release, but I will wait several days maybe we get more issues to fix :cry:
I will prepare a fix for endless loop and a test to prevent such issue in the future.

jpstotz:
@skylot Reconsidering the endless loop with the IOException dialog, it doesn't make sense that this was caused by the dialog itself. At that time I have problems with the new LazyLoadingBinaryData which threw IOException because the used InputStream was closed by Jadx. Most likely the loop was caused by frequent calls to LazyLoadingBinaryData causing each time an IOException.

PR #2917, opened 2026-07-13
Add file associations for macos

oriori1703:
This lets you open associated files by double clicking them on MacOS.

Right now this is only for mac, but could potentially be used for windows (or future linux packages).
For this to work in windows we need to also ship an msi installer, because jpackage associations are only registered in install time.

Also I arbitrarily chose `application/x-jadx-project` for the mime-type of project files, so feel free to change it to whatever you want.
This doesn't matter for the purpose of opening it in MacOS. Only the extension matters. 

Follow up to #2916


jadx-gui/build.gradle.kts
```diff
@@ -193,13 +193,18 @@ runtime {
 	jpackage {
 		if (DefaultNativePlatform.getCurrentOperatingSystem().isMacOsX) {
 			imageName = "jadx-gui"
+			val fileAssociations =
+				fileTree("$projectDir/src/main/resources/jpackage-file-associations") { include("*.properties") }
```
skylot: Do we really need these files inside resources directory? They will be inside result jar file for all platforms.
Can we move these files (and `.icns`) info dir like `jadx-gui/dist/macos`?

oriori1703: Done :)

BTW, shouldn't the windows `.ico` file be moved to a similar directory?

skylot (approved):

jadx-gui/build.gradle.kts
skylot: Thanks :+1:

Yeah, `.ico` not needed in resources dir and should be moved.

Issue #2924, opened 2026-07-27
[gui] The list of sub-items in the resource file directory seems to be arranged in a stacked manner.


BBBB1:
### Issue details

 
<img width="531" height="272" alt="Image" src="https://github.com/user-attachments/assets/cabc164f-2777-499c-89d8-3c3655c3d9d7" />

### Jadx version

1.5.6

### Java version

25.0.3

### OS

- [x] Windows
- [ ] Linux
- [ ] macOS

jpstotz:
I assume you have opened the issue because the text of the element can hardly be read? It looks like multiple characters are printed over another. May be some of the folder names uses right-to-left notation?

Because of the combination of folders that only have one sub-folder into one tree node is a feature not a bug. 

Would it be possible to attach the APK file that is shown in your screenshot to this issue? Just add a `.zip` file extension so GitHub doesn't rejects it.

BBBB1:
> 我猜您提交这个问题是因为元素中的文本难以辨认？看起来好像多个字符重叠在一起了。或许某些文件夹名称使用的是从右到左的书写方式？
> 
> 将只有一个子文件夹的文件夹合并到一个树节点中是一种特性，而不是一个错误。
> 
> 能否将截图中显示的 APK 文件附加到此问题中？只需添加`.zip`文件扩展名，以免 GitHub 拒绝上传即可。

[debug.zip](https://github.com/user-attachments/files/30423484/debug.zip)
No problem. Since GitHub has a file size limitation, I removed some unnecessary resources from the APK to reduce its size before uploading.

The removed resources are not related to this issue, and the problem can still be reproduced with this modified APK.

jpstotz:
Ok happens here, too. But the actual behavior is very funny:

After loading I see the same picture as you have posted. 

Selecting the entry makes it fully readable: 

<img width="437" height="48" alt="Image" src="https://github.com/user-attachments/assets/cf8b5e3e-a840-4e2c-8971-889664eab66f" />

Deselecting it often changes the state back to this:

<img width="383" height="44" alt="Image" src="https://github.com/user-attachments/assets/da49d472-ad66-41d7-b74e-757b9493de7d" />

Looks to me like a rendering problem of the used tree component, but the component is  simple JTree. I don't know if both problems are related but your posted screen-shot also shows some empty space before the entry. This is also not normal. 

[...]

PR #2866, opened 2026-05-07
collapse single-child directory chains in resource tree

RuffaloLavoisier:
:exclamation: Please review the [guidelines for contributing](https://github.com/skylot/jadx/blob/master/CONTRIBUTING.md#Pull-Request-Process)

### Description
Please describe your pull request.
Reference issue it fixes.

image

fix #2729

PR #2925, opened 2026-07-27
Fix collapsed single-child directory chains display problems

jpstotz:
The code for collapsing nodes which was introduced in PR #2866 had some problems causing various display glitches (see #2924).

This PR fixes those problems. It simplifies the collapsing code, removing one recursion.
  
In addition it also fixes the problem that the the collapsed path was only updated in the `shortName` of that node, but not in it's `name`. 

In my tests all collapsed entries worked as expected, I extended the sample from #2924 by two-level collapsing, and even collapsed dir chain containing again a collapsed dir chain.

image

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/treemodel/JResource.java
+++ b/jadx-gui/src/main/java/jadx/gui/treemodel/JResource.java
@@ -2,11 +2,12 @@ package jadx.gui.treemodel;
 
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


skylot (approved):
Looks good, thanks :+1:

But I am still not sure about a glitch root cause.
It might be that a node updated without proper UI repaint, and your change return a new node which can trigger such update :thinking:

jpstotz:
@skylot the new node is not the solution. Before I updated the name field I simply updated the first chain node node and the UI problem was already fixed.

The original code updated the last node of the chain. But at the end of the now removed recursion the update part was executed multiple time, that must have triggered somehow the problem.

skylot:
So, multiple updates is a root cause. Still hard to understand. Anyway, thanks for help :slightly_smiling_face:

PR #2933, opened 2026-08-05
feat(UI): Native ELF library symbol view, Resource tree filter, Strings search window, Inheritance graph updates

tim21-f0:
This PR adds multiple features form a private fork:
- Native ELF library symbol view when clicking on native libraries in the resource tree
- Resource tree filter which can filter the expanded tree by package name and class name
- Strings search window accessible from the "View" drop down allows users to view, filter and navigate to defined strings
- Inheritance graph updates enable viewing supertypes and subtypes of the selected class


jpstotz:
@tim21-f0 Thanks for this PR: Personally I would prefer individual PRs for each feature. That way review and finalization (e.g. bug fixing) of each feature would be easier.

In my short tests the resource tree filter seem to blocking the UI, so while the filter process is running Jadx is completely blocked. 

With the new panel concept for resources I wouldn't squeeze two unrelated features into one UI like the hex viewer and the ELF function list. We already have plugins like [jadx-native-libraries-plugin](https://github.com/andyjsmith/jadx-native-libraries-plugin) which allow accessing ELF functions. This plugin also provides more features like jump to the Java native code method declarations of JNI methods. 

Using the strings search window I mostly end up in a totally frozen Jadx-Gui. Even waiting so minutes did not unfreeze the UI. The only resolution I found was killing the Jadx-Gui process.

tim21-f0:
@jpstotz thanks for the feedback.

I will split the features up into separate PRs.

We were not aware of the native libraries plugin but I suspect it fulfils our needs so I'm happy to leave out the symbol view change.

I will spend some time investigating the improvements for the other UI features you mentioned.

PR #2935, opened 2026-08-11
feat(UI): Make inheritance graphs display both super and subclass relations

tim21-f0:
This change displays both super-classes and subclasses of the selected class in inheritance diagrams whereas previously only super-classes were displayed.

In detail:

The change extends the `ClspGraph` to track both aggregate supertypes and implementations, and direct parents and sub-classes. The inheritance graph now walks this cache of `.parent` relationships rather than accessing `.superType` and `.interfaces` directly.

This relies on the implied assumption that ClspGraph is up to date and `.superType` and `.interfaces`  cannot change after it is created. Testing suggests that this is true.

Edges are no longer labelled with `implements` or `extends`. Edges now go the other way - the source is the super-class and the destination is the subclass. Interfaces have dashed boarders and classes have full boarders. The current class is highlighted with a blue border.

The new system loses information about whether a parent is implemented or extended so for classes that cannot be resolved to a `ClassNode` or `ClspClass` we can no longer tell if they are interfaces or classes.


tim21-f0:
Apologies - should be up to date with master now.

tim21-f0:
Force pushed again to remove any changes in the commit history to the ADB.java file - sorry for the confusion.  Originally the source branch was not up to date with master so the ADB.java file erroneously showed up as being changed.

skylot (approved):

skylot:
Hm, looks like new caches in `ClspGraph` will be used only by jadx-gui and only if inheritance graph is build.
In jadx-cli and jadx-core used as a library these maps will not be used and only pollute memory.
So, I moved these caches into separate class which calculate them on demand.
My change also reverts `ClspGraph` to original state, because added comment are pointless in most cases (just repeat method name).

PR #2937, opened 2026-08-13
fix(gui): use save mode for package directory export

HdShare:
### Description

fix package directory export by using save mode for `EXPORT_NODE_FOLDER`.
use `Files.createDirectories` to support nested package directories.


skylot (approved):
Nice! Thank you :+1:

PR #2939, opened 2026-08-24
ci: upgrade actions/setup-java to v6

brunoborges:

### Description

Upgrade all actions/setup-java references in the active build and release workflows from v5 to v6. This moves the action runtime to Node.js 24 while preserving the existing Java distributions and versions.

Validation: `git diff --check` and a scan confirming no pre-v6 actions/setup-java references remain in active workflows.


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #2942, opened 2026-08-26
feat(ui): add strings list view pane

tim21-f0:
This change adds a new "Strings List" pane which can be found under `View` -> `Strings List`.

On opening the new pane, a background task is initiated to search for all smali strings.

The pane has the following:
- A table of the found strings, the left column contains the string and the right column contains the references in the code where it is found.
- On double clicking one of the references the user is navigated to the code location in the main UI's code area.
- Results can be sorted lexicographically.
- Results can be filtered by string (literal and regex), package and minimum string length.


0rShemesh:
You could use my plugin:

https://github.com/0rShemesh/jadx-magic-strings

tim21-f0:
@0rShemesh This plugin does look to handle all features in this current PR and so I will propose to close this one.  The only thing that would be nice to have would be to be able to navigate to the defined strings in the Decompilation Window when double clicking the strings in the 'All Strings' tab (like with the classes and methods in the other tabs).

PR #2943, opened 2026-08-26
fix(gui): avoid project reload when saving settings on start page

HdShare:
### Description

Avoid calling `MainWindow.reopen()` when saving settings on the start page with no input files, preventing the incorrect “Can't open empty project” dialog.

Settings and shortcuts are still applied, and menu accelerators are refreshed immediately. Reload and input-file verification behavior for opened projects remains unchanged.


skylot (approved):
Thanks :+1:

Although, it might be better to add some kind of flag if project is not yet actually opened and handle such case in `reopen` method directly, will try to implement that later.

PR #2948, opened 2026-09-05
fix(gui): load next binary block at block boundary

### LINKED ISSUE #2947: [gui] IndexOutOfBoundsException at 512 KiB boundary in Hex View
### Issue details

Opening a resource larger than 512 KiB in the Hex View can cause an IndexOutOfBoundsException when navigating to offset 0x80000.

LazyLoadingBinaryData uses 512 KiB blocks. The constructor loads the first complete block, so readPos becomes 524288 while blocks.size() is still 1.

When getByte(524288) is called, ensurePositionLoaded(524288) does not load the next block because readPos < position is false. getByte then calculates:

    blockNum = 524288 / 524288 = 1

and accesses blocks.get(1), which throws IndexOutOfBoundsException.

Steps to reproduce:

1. Open a ZIP/APK resource larger than 512 KiB in jadx-gui.
2. Open the resource in Hex View.
3. Navigate to offset 0x80000.

Actual result:

IndexOutOfBoundsException is thrown while reading the byte at the block boundary.

Expected result:

The byte at offset 0x80000 and subsequent bytes should be readable normally.

A minimal reproduction is:

    byte[] source = new byte[1024 * 512 + 2];
    LazyLoadingBinaryData data = new LazyLoadingBinaryData(
        new ByteArrayInputStream(source),
        source.length);

    data.getByte(1024 * 512);

### Jadx version

latest unstable version

### Java version

17.0.19

### OS

- [ ] Windows
- [ ] Linux
- [x] macOS

cuishuang:
:exclamation: Please review the contribution guidelines:
https://github.com/skylot/jadx/blob/master/CONTRIBUTING.md#pull-request-process

### Description

Fix an IndexOutOfBoundsException in LazyLoadingBinaryData when getByte() is called exactly at a 512 KiB block boundary.

The previous implementation only ensured that the stream had been read up to the requested position. For a position equal to readPos, the next block was not loaded even though getByte() subsequently accessed that block.

Fixes https://github.com/skylot/jadx/issues/2947

### DIFF
--- a/jadx-gui/src/main/java/jadx/gui/ui/hexviewer/LazyLoadingBinaryData.java
+++ b/jadx-gui/src/main/java/jadx/gui/ui/hexviewer/LazyLoadingBinaryData.java
@@ -97,8 +97,14 @@ public class LazyLoadingBinaryData implements BinaryData {
 
 	@Override
 	public byte getByte(long position) {
-		if (!ensurePositionLoaded(position)) {
-			throw new RuntimeException("Unreachable position: " + position);
+		if (position < 0) {
+			throw new IndexOutOfBoundsException("Position out of bounds: " + position);
+		}
+		if (!ensurePositionLoaded(position + 1)) {
+			throw new IndexOutOfBoundsException("Unreachable position: " + position);
+		}
+		if (position >= size) {
+			throw new IndexOutOfBoundsException("Position out of bounds: " + position);
 		}
 		int blockNum = (int) (position / blockSize);
 		int blockOffset = (int) (position % blockSize);
--- a/jadx-gui/src/test/java/jadx/gui/ui/hexviewer/LazyLoadingBinaryDataTest.java
+++ b/jadx-gui/src/test/java/jadx/gui/ui/hexviewer/LazyLoadingBinaryDataTest.java
@@ -1,0 +1,55 @@
+package jadx.gui.ui.hexviewer;
+
+import java.io.ByteArrayInputStream;
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.util.Arrays;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class LazyLoadingBinaryDataTest {
+
+	private static final int BLOCK_SIZE = 1024 * 512;
+
+	@Test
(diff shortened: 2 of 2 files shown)


jpstotz:
Thanks for identifying this problem. However your fix also has some problems:

It only works if the used stream knows it's size and provides it via `getAvailable()`. 
If the stream does not know it's size then your size check will fail even if the stream could provide that data:

This code sample simulates such a behavior:

```
byte[] source = new byte[1024 * 512 + 2];
LazyLoadingBinaryData data = new LazyLoadingBinaryData(
		new FilterInputStream(new ByteArrayInputStream(source)) {
			@Override
			public int available() throws IOException {
				return -1; // simulate unknown size
			}
		},
		0);
data.getByte((long) 1024 * 512);
```

data.getByte will fail with `IndexOutOfBoundsException: Position out of bounds: 524288` because the known size is only 524288 (size of the first block).

jpstotz:
If I am not mistaken switching the two checks in `getByte` should be sufficient:

First call `ensurePositionLoaded(position + 1)`- at the block border this will force loading of the next block and thus updating the size. 

`if (!ensurePositionLoaded(position + 1)) {throw new IndexOutOfBoundsException("Unreachable position: " + position);}`

Second we can check if the position is within range:

`if (position < 0 || position >= size) { throw new IndexOutOfBoundsException("Position out of bounds: " + position);}`

That way it should work - do you agree?

cuishuang:
> If I am not mistaken switching the two checks in `getByte` should be sufficient:
> 
> First call `ensurePositionLoaded(position + 1)`- at the block border this will force loading of the next block and thus updating the size.
> 
> `if (!ensurePositionLoaded(position + 1)) {throw new IndexOutOfBoundsException("Unreachable position: " + position);}`
> 
> Second we can check if the position is within range:
> 
> `if (position < 0 || position >= size) { throw new IndexOutOfBoundsException("Position out of bounds: " + position);}`
> 
> That way it should work - do you agree?

Yes, I agree. The size check must happen after ensurePositionLoaded(), because size may only represent the amount known so far when the stream does not provide its size.

I updated getByte() to reject negative positions first, then call ensurePositionLoaded(position + 1), and only afterward check the position against the updated size.

I also added a regression test covering a stream with unknown size (available() == -1) and the byte at the 512 KiB block boundary.

Please review it again.

jpstotz (approved):

skylot (approved):

PR #2945, opened 2026-09-02
Feat/app scoped gui plugins

mostafaNazari702:
:exclamation: Please review the [guidelines for contributing](https://github.com/skylot/jadx/blob/master/CONTRIBUTING.md#Pull-Request-Process). 

Done.

### Description

Initiate GUI plugins on jadx-gui start

Fixes #2913

Right now GUI plugins only start once a project is loaded and they get unloaded again as soon as you close it so a plugin cant do anything on the start page or survive switching between APKs. Added a `JadxGuiPlugin` interface plugins can opt into: they get `initGui(JadxGuiContext)` once when the window opens and `unload()` when jadx-gui exits and they are skipped by the decompiler's plugin manager so they never get a second project-scoped lifecycle.

The GUI plugins context also moves from `JadxWrapper` to `MainWindow´ and now keeps application and project registrations apart so closing a project only clears the project ones. Existing plugins are not affected and they keep using `getGuiContext()` exactly as before.

Issue #2575, opened 2025-07-18
jadx.core.utils.exceptions.JadxRuntimeException: Failed to clear directory C:\Users\nzaar9\AppData\Local\Temp\jadx-temp-9028579072555110246\jadx-instance-16828034844905192271


nzaar9:
Please describe what you did before the error occurred.
**IMPORTANT!** If the error occurs with a specific APK file please attach or provide link to apk file!
 * Jadx version: 1.5.1
* Java version: 23.0.1
* Java VM: Oracle Corporation OpenJDK 64-Bit Server VM
* Platform: Windows 11 (10.0 amd64)
* Max heap size: 34368 MB

```
jadx.core.utils.exceptions.JadxRuntimeException: Failed to clear directory C:\Users\XXXX\AppData\Local\Temp\jadx-temp-9028579072555110246\jadx-instance-16828034844905192271
	at jadx.core.utils.files.FileUtils.clearDir(FileUtils.java:199)
	at jadx.core.utils.files.FileUtils.clearTempRootDir(FileUtils.java:178)
	at jadx.api.JadxDecompiler.close(JadxDecompiler.java:184)
	at jadx.gui.utils.plugins.CollectPlugins.build(CollectPlugins.java:56)
	at jadx.gui.settings.ui.plugins.PluginSettings.build(PluginSettings.java:56)
	at jadx.gui.settings.ui.JadxSettingsWindow.initUI(JadxSettingsWindow.java:140)
	at jadx.gui.settings.ui.JadxSettingsWindow.<init>(JadxSettingsWindow.java:102)
	at jadx.gui.ui.MainWindow.openSettings(MainWindow.java:1493)
	at jadx.gui.utils.ui.ActionHandler.lambda$new$0(ActionHandler.java:20)
	at jadx.gui.utils.ui.ActionHandler.actionPerformed(ActionHandler.java:79)
	at jadx.gui.ui.action.JadxGuiAction.actionPerformed(JadxGuiAction.java:100)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(Unknown Source)
	at java.desktop/javax.swing.AbstractButton.doClick(Unknown Source)
[...]

PR #2960, opened 2026-09-23
feat: load global GUI plugins without decompiler, add first IMainWindow method

mostafaNazari702:

### Description

Follow-up to #2945 (issue #2913): resolves the two TODOs left there.

**1. Global plugins no longer need a decompiler instance**

Global GUI plugins were loaded through a throwaway decompiler only because the plugins manager required one.
Now the plugins manager can be created without a decompiler. Loading, filtering (disabled, version, duplicate ids, `provides`) and registration work the same, a decompiler is only needed when plugins are initialized for a project.

To make this possible the plugin data kept by the manager is separated from the context a plugin receives on init.
The public `JadxPluginContext` API is unchanged, project plugins still get the real project decompiler, passes, events, options, etc.

I checked this at runtime, on the start page with one global plugin loaded there was 1 live decompiler instance before this change and 0 now.

**2. First method in `IMainWindow`**

Added `open(files)`. It exposes the existing "open files" action (same as drag-and-drop or recent projects) so a global plugin can open the next APK which was the original use case in #2913.
It uses only JDK types no jadx-gui internals. I kept it to one method on purpose and can add more once we agree on what the API should cover.

**Tests**
- new tests for the plugins manager with and without a decompiler
- global plugin tests rewritten without reflection: lifecycle across several projects, reload, options, custom settings, disabled plugins
- API test checking `IMainWindow` does not expose jadx-gui internal types
- manual check with the test plugins runner: the test plugin reopens the project through `IMainWindow.open`

**Questions**
- "Unbind": is "a decompiler is only needed to init plugins" what you had in mind or should it be removed from the manager completely?
- Is `open` a good first method? What should come next (close/save project, project info, app-level events)?
- The internal plugin data class no longer implements `JadxPluginContext` (plugins get a separate context object). Should it be renamed?


mostafaNazari702:
**Things i noticed while working on this (not changed in this PR)**
- Disabling or uninstalling a global plugin only takes effect after restart: it keeps running and is still added to new projects.
- A global plugin whose global init failed is still added to every project and gets project init/unload.
- A project plugin with the same id as a global plugin silently replaces it in that project (no duplicate id error).
- Global and project scope share one options object so a project's own option values stay visible to the global plugin after the project is closed.
- While global plugins are still initializing, the settings page already lists them but without their options.
- On Windows, a running global plugin keeps its jar locked so Uninstall removes it from the list but silently leaves the jar on disk. Other plugin jars are also kept open by the global loader until GC.
- Global plugin init runs on the same background thread as project open/close so a slow global init delays opening a project and closing the window. The same thing is what guarantees global plugins init before the first project.
- Calling `open` from a plugin menu action logs "UI thread interrupted": `open` cancels background jobs including the menu action's own task. The project still reloads fine.
- Older behavior: opening plugin settings with no project open unloads the temporary plugins twice and deletes the shared temp directory (possibly related to #2575).
- The Plugins menu is sometimes updated from background threads instead of the UI thread.
- With Windows default encoding, `ResNameUtilsTest` and `ModifiedUTF8DecoderTest` fail because of non-ASCII literals. This is unrelated and the same on master.

Should any of these become separate issues? I can open them.

mostafaNazari702:
@skylot I would be thankful if you review this PR ASAP. I have also some other ideas of my own that depend on this PR. Thanks and good luck while reviewing!

